Issue #7, opened 2019-02-03
write some unit tests


andyzhangx:
Complete following unit tests:
 - `getFStype` func: https://github.com/andyzhangx/azuredisk-csi-driver/blob/d220d45916e2460c689020c06fbde275f63005b5/pkg/azuredisk/nodeserver.go#L280
 - `GetCloudProvider`: https://github.com/andyzhangx/azuredisk-csi-driver/blob/d220d45916e2460c689020c06fbde275f63005b5/pkg/azuredisk/azure.go#L29
- `getCachingMode`: https://github.com/andyzhangx/azuredisk-csi-driver/blob/d220d45916e2460c689020c06fbde275f63005b5/pkg/azuredisk/azure.go#L29

could refer to unit test: 
https://github.com/andyzhangx/azuredisk-csi-driver/blob/d220d45916e2460c689020c06fbde275f63005b5/pkg/azuredisk/azuredisk_test.go#L25-L54

PR #11, opened 2019-02-03
add two unit tests

lizebang:
I only add two unit tests. #7 

I have two questions here when writing tests.

1. `getFStype` func:
https://github.com/andyzhangx/azuredisk-csi-driver/blob/d220d45916e2460c689020c06fbde275f63005b5/pkg/azuredisk/nodeserver.go#L280
I pass in `map[string]string{"fstype": ""}`, and it returns `""`. It's this right? I think it will return `defaultFsType`, but it's not.

2. `GetCloudProvider` func:
https://github.com/andyzhangx/azuredisk-csi-driver/blob/ec2e0055dc374c4701cc61bc941239d8931a63df/pkg/azuredisk/azure.go#L29
I used the `deploy/example/azure.json` file for testing. It clearly failed. Could you add a valid `AZURE_CREDENTIAL_FILE` environment variable to travis-ci environment variable. There may be other ways to complete the test, but I didn't get it.


andyzhangx:
> 1. I pass in `map[string]string{"fstype": ""}`, and it returns `""`. It's this right? I think it will return `defaultFsType`, but it's not.

That's expected since you explicitly specify `fstype` as "".

> 2\. I used the `deploy/example/azure.json` file for testing. It clearly failed. Could you add a valid `AZURE_CREDENTIAL_FILE` environment variable to travis-ci environment variable. There may be other ways to complete the test, but I didn't get it.

You could use `test/integration/azure.json` and replace with following env vars:
```
Setting environment variables from repository settings
$ export aadClientId=[secure]
$ export aadClientSecret=[secure]
$ export tenantId=[secure]
$ export subscriptionId=[secure]
$ export resourceGroup=[secure]
$ export nodeid=[secure]
```

example could be found here:
https://github.com/andyzhangx/azuredisk-csi-driver/blob/bbe21fa4284f4b448c561cc6cc15e6f4897f282f/test/integration/run-test.sh#L23-L27

While I doubt is it really need to use real azure credentials, use fake cred may still pass in the unit test.

pkg/azuredisk/nodeserver_test.go
```diff
@@ -0,0 +1,57 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
andyzhangx: 2019
BTW, could you work out a tool to add this statement to all source code under this project? Thanks.

pkg/azuredisk/controllerserver_test.go
```diff
@@ -0,0 +1,76 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
andyzhangx: 2019

pkg/azuredisk/controllerserver_test.go
```diff
@@ -0,0 +1,76 @@
...
+		},
+		{
+			map[string]string{"cachingmode": "ReadWrite"},
+			compute.CachingTypes("ReadWrite"),
+			nil,
+		},
+		{
+			map[string]string{"cachingmode": "WriteOnly"},
+			compute.CachingTypes(""),
+			fmt.Errorf("azureDisk - %s is not supported cachingmode. Supported values are %s", "WriteOnly", supportedCachingModes.List()),
```
andyzhangx: I think check whether there is error is enough, otherwise if the original error changes, unit test will always change.

pkg/azuredisk/controllerserver_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
+)
+
+func TestGetCachingMode(t *testing.T) {
+	tests := []struct {
+		options   map[string]string
+		expected1 compute.CachingTypes
+		expected2 error
```
andyzhangx: expectedError bool

pkg/azuredisk/controllerserver_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
+)
+
+func TestGetCachingMode(t *testing.T) {
+	tests := []struct {
+		options   map[string]string
+		expected1 compute.CachingTypes
```
andyzhangx: expectedCachingMode

pkg/azuredisk/nodeserver_test.go
```diff
@@ -0,0 +1,57 @@
...
+package azuredisk
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestGetFStype(t *testing.T) {
+	tests := []struct {
+		options  map[string]string
```
andyzhangx: add case like `fstype`: `EXT4`

andyzhangx:
Please follow instructions at https://github.com/kubernetes/kubernetes/wiki/CLA-FAQ to sign the CLA.

pkg/azuredisk/nodeserver_test.go
lizebang: Thanks, I will fix it.

pkg/azuredisk/controllerserver_test.go
lizebang: I will fix it.

pkg/azuredisk/controllerserver_test.go
lizebang: You're right, I will modify it.

lizebang:
@andyzhangx 
GetCloudProvider called azure.NewCloud, and azure.NewCloud has been tested at [azure_test.go](https://github.com/andyzhangx/azuredisk-csi-driver/blob/master/vendor/k8s.io/kubernetes/pkg/cloudprovider/providers/azure/azure_test.go). So, I don't know if it is necessary to test here.
Where should I put the test files if necessary? `test/azuredisk/run-azure-test.sh` and `test/azuredisk/azure_test.go`?

andyzhangx:
@lizebang The `GetCloudProvider` test is already covered in integration test, let's skip it now.
Have you singed the CLA? If not, please follow instructions at https://github.com/kubernetes/kubernetes/wiki/CLA-FAQ to sign the CLA.

lizebang:
I already signed it.

PR #80, opened 2019-05-15
remove invalid OWNERS_ALIASES file

spiffxp:
I'm doing some org/repo-wide OWNERS file parsing, and this is an
invalid OWNERS_ALIASES file.  It doesn't actually contain any aliases so
doesn't seem like there's much point in keeping it around.  Our 
`verify-owners` plugin should catch these in the future.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,5 +1,0 @@
-aliases:
-- andyzhangx
-- brendandburns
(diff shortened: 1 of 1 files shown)


spiffxp:
ref: encountered while doing analysis for https://github.com/kubernetes/community/issues/1673

PR #95, opened 2019-06-28
fix: make azure disk URI as case insensitive

andyzhangx:
**What type of PR is this?**
/kind bug

This PR makes the azure disk URI as case insensitive, otherwise there could be error like following if customer ignore the disk URI case:
```
Error expanding volume "del-me/test-pvc" of plugin kubernetes.io/azure-disk : invalid disk URI: /subscriptions/xxx/resourcegroups/xxx/providers/Microsoft.Compute/disks/Test_Resize_1/”
```

backport PR: https://github.com/kubernetes/kubernetes/pull/79020

Fixes #

**Release note**:
```
fix: make azure disk URI as case insensitive
```


andyzhangx:
cc @ZeroMagic for code review drill

Issue #46, opened 2019-02-22
fix sanity test failures


andyzhangx:
**What happened**:
Follow this guide: https://github.com/kubernetes-sigs/azuredisk-csi-driver/tree/master/test/sanity

Thanks for @Masquerade0097 's contribution, we have a lot more work to do to fix those sanity test failures:

 - below are the sanity test failure:
```
Summarizing 9 Failures:

[Fail] Node Service [It] should work 
/home/priyanshu/work/src/github.com/csi-driver/azuredisk-csi-driver/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:411

[Fail] Controller Service CreateVolume [It] should not fail when requesting to create a volume with already existing name and same capacity. 
/home/priyanshu/work/src/github.com/csi-driver/azuredisk-csi-driver/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:371

[Fail] Controller Service CreateVolume [It] should fail when requesting to create a volume with already existing name and different capacity. 
/home/priyanshu/work/src/github.com/csi-driver/azuredisk-csi-driver/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:442

[Fail] Controller Service DeleteVolume [It] should succeed when an invalid volume id is used 
/home/priyanshu/work/src/github.com/csi-driver/azuredisk-csi-driver/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:543

[Fail] Controller Service ValidateVolumeCapabilities [It] should fail when the requested volume does not exist 
/home/priyanshu/work/src/github.com/csi-driver/azuredisk-csi-driver/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:705

[Fail] Controller Service ControllerPublishVolume [It] should return appropriate values (no optional values added) 
/home/priyanshu/work/src/github.com/csi-driver/azuredisk-csi-driver/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:826

[Fail] Controller Service ControllerPublishVolume [It] should fail when the volume does not exist 
/home/priyanshu/work/src/github.com/csi-driver/azuredisk-csi-driver/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:883

[Fail] Controller Service ControllerPublishVolume [It] should fail when the node does not exist 
[...]

PR #107, opened 2019-07-08
fix: check the diskName

ZeroMagic:
**What type of PR is this?**
/kind bug
/kind failing-test

Check the diskName before creating azure disk.

Fixes #46 

**Release note**:
```
fix: set disk name as volume name
```

### DIFF
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -23,15 +23,18 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-03-01/compute"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/pborman/uuid"
 
 	csicommon "github.com/kubernetes-sigs/azuredisk-csi-driver/pkg/csi-common"
 	kwait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
 	"k8s.io/kubernetes/pkg/util/mount"
+	"k8s.io/kubernetes/pkg/volume/util"
 	"k8s.io/legacy-cloud-providers/azure"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/azuredisk/azuredisk.go
```diff
@@ -315,3 +322,34 @@ func (d *Driver) CreateManagedDisk(ctx context.Context, options *ManagedDiskOpti
...
+
+// The name of the managed disk that is being created.
+// he name can't be changed after the disk is created.
+// Supported characters for the name are a-z, A-Z, 0-9 and _.
+// The maximum name length is 80 characters.
+//
+// See https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate#uri-parameters
+func getValidDiskName(volumeName string) string {
+	diskName := strings.Replace(volumeName, "-", "_", -1)
+	if len(diskName) > diskNameMaxLength {
```
feiskyer: add a warning log here since the name is truncated.

@andyzhangx any possible of same names after truncation?

ZeroMagic: Done. 

In the next step, we will call azure cloud's api to create volume. In the api, the name will be checked if it exists.

andyzhangx: looks like the disk name could contain `-` , while according to doc, it said: 
```
 Supported characters for the name are a-z, A-Z, 0-9 and _. The maximum name length is 80 characters.
```
Let me ask azure disk team about this doc issue.

andyzhangx: @ZeroMagic Let's use the naming rule azure portal:
- _ . are all allowed chars

pkg/azuredisk/azuredisk.go
```diff
@@ -315,3 +322,36 @@ func (d *Driver) CreateManagedDisk(ctx context.Context, options *ManagedDiskOpti
...
+	}
+
+	return diskName
+}
+
+func checkDiskName(diskName string) bool {
+	length := len(diskName)
+
+	for i, v := range diskName {
+		if !(unicode.IsLetter(v) || unicode.IsDigit(v) || v == '_' || v == '.' || v == '-') ||
```
andyzhangx: `- _ .` are all allowed chars

andyzhangx: fix this

pkg/azuredisk/azuredisk_test.go
```diff
@@ -200,3 +200,109 @@ func TestIsValidDiskURI(t *testing.T) {
...
+		{
+			diskName: "1_",
+			expected: true,
+		},
+		{
+			diskName: "_1",
+			expected: false,
+		},
+		{
+			diskName: "1.",
```
andyzhangx: add some positive cases, e.g. "1.2", "1_2", "1-3"

pkg/azuredisk/azuredisk.go
ZeroMagic: done

pkg/azuredisk/azuredisk_test.go
ZeroMagic: done

PR #108, opened 2019-07-08
fix: set driver name

ZeroMagic:
Driver name should be a certain value. 
With the fix, "Identity Service GetPluginInfo [It] should return appropriate information" will be fixed.

Signed-off-by: ZeroMagic <anthonyliu@zju.edu.cn>

**What type of PR is this?**
>
> /kind bug
> /kind failing-test

Fixes #46

**Release note**:
```
fix: set driver name
```


ZeroMagic:
done.

PR #116, opened 2019-07-10
test: init klog flags

ZeroMagic:
**What type of PR is this?**
/kind flake

- init klog flags

Because the version of klog is v0.3.3, we should init flags. So that we can get the logs from terminal.

Fixes #

**Release note**:
```
none
```


ZeroMagic:
ok. I will do it later.

PR #119, opened 2019-07-11
fix: return appropriate error code when snapshot exists

ZeroMagic:
**What type of PR is this?**
/kind bug
/kind failing-test

When creating the snapshot, we should return appropriate error code "AlreadyExists" if snapshot exists.

Fixes #46 

**Release note**:
```
fix: return appropriate error code when snapshot exists
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -464,7 +464,11 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 	klog.V(2).Infof("begin to create snapshot(%s) under rg(%s)", snapshotName, d.cloud.ResourceGroup)
 	future, err := d.cloud.SnapshotsClient.CreateOrUpdate(ctx, d.cloud.ResourceGroup, snapshotName, snapshot)
 	if err != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("create snapshot error: %v", err))
+		if strings.Contains(err.Error(), "existing disk") {
+			return nil, status.Error(codes.AlreadyExists, fmt.Sprintf("request snapshot(%s) under rg(%s) already exists, but the SourceVolumeId is different, error details: %v", snapshotName, d.cloud.ResourceGroup, err))
+		} else {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("create snapshot error: %v", err))
+		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/azuredisk/controllerserver.go
```diff
@@ -461,7 +461,11 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 	klog.V(2).Infof("begin to create snapshot(%s) under rg(%s)", snapshotName, d.cloud.ResourceGroup)
 	future, err := d.cloud.SnapshotsClient.CreateOrUpdate(ctx, d.cloud.ResourceGroup, snapshotName, snapshot)
 	if err != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("create snapshot error: %v", err))
+		if strings.Contains(err.Error(), "existing") {
```
andyzhangx: what's the exact error? only matching `existing` word may be not enough

ZeroMagic: [CreateSnapshot] error(compute.SnapshotsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="BadRequest" Message="Changing property 'sourceResourceId' is not allowed for existing disk 'CreateSnapshot-snapshot-2'.")

andyzhangx: use `existing disk` word?

pkg/azuredisk/controllerserver.go
```diff
@@ -464,7 +464,11 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 	klog.V(2).Infof("begin to create snapshot(%s) under rg(%s)", snapshotName, d.cloud.ResourceGroup)
 	future, err := d.cloud.SnapshotsClient.CreateOrUpdate(ctx, d.cloud.ResourceGroup, snapshotName, snapshot)
 	if err != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("create snapshot error: %v", err))
+		if strings.Contains(err.Error(), "existing disk") {
+			return nil, status.Error(codes.AlreadyExists, fmt.Sprintf("the request snapshot already exists, but the SourceVolumeId is different"))
```
andyzhangx: `request snapshot(%s) under rg(%s) already exists, but the SourceVolumeId is different, error details: %v`
we should always show the error details and provide more details if possible.

PR #120, opened 2019-07-11
fix: validate snapshot name

ZeroMagic:
**What type of PR is this?**
/kind bug
/kind failing-test

- fix: validate snapshot name

With this PR, "should not fail when creating snapshot with maximum-length name" in sanity test will pass.

Fixes #46 

**Release note**:
```
fix: validate snapshot name
```

### DIFF
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -53,6 +53,8 @@ const (
 	// see https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate#uri-parameters
 	diskNameMinLength = 1
 	diskNameMaxLength = 80
+	// maxLength = 80 - (4 for ".vhd") = 76
+	diskNameGenerateMaxLength = 76
 
 	//default disk size is 1 GiB
 	defaultDiskSize = 1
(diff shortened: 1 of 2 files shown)


pkg/azuredisk/azuredisk.go
```diff
@@ -344,30 +345,43 @@ func (d *Driver) CreateManagedDisk(ctx context.Context, options *ManagedDiskOpti
...
+		klog.Warningf("since the maximum name length is %d, so it is truncated as (%q)", diskOrSnapshotNameMaxLength, validName)
 	}
-	if !checkDiskName(diskName) || len(diskName) < diskNameMinLength {
-		// maxLength = 80 - (4 for ".vhd") = 75
-		// todo: get cluster name
-		diskName = util.GenerateVolumeName("pvc-disk", uuid.NewUUID().String(), 76)
-		klog.Warningf("the requested volume name (%q) is invalid, so it is regenerated as (%q)", volumeName, diskName)
+	if !checkDiskOrSnapshotName(validName) || len(validName) < diskOrSnapshotNameMinLength {
+		switch volumeType {
+		case "disk":
```
andyzhangx: I think append a volume name prefix is better instead of a self-defined `volumeType`

pkg/azuredisk/azuredisk.go
```diff
@@ -51,8 +51,9 @@ const (
 	defaultDiskMBpsReadWrite = 100
 
 	// see https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate#uri-parameters
-	diskNameMinLength = 1
-	diskNameMaxLength = 80
+	// see https://docs.microsoft.com/en-us/rest/api/compute/snapshots/createorupdate#uri-parameters
+	diskOrSnapshotNameMinLength = 1
```
andyzhangx: keep `diskNameMinLength` and add comment and link that snapshot naming rule is the same

pkg/azuredisk/azuredisk.go
```diff
@@ -344,30 +345,43 @@ func (d *Driver) CreateManagedDisk(ctx context.Context, options *ManagedDiskOpti
...
-	diskName := volumeName
-	if len(diskName) > diskNameMaxLength {
-		diskName = diskName[0:diskNameMaxLength]
-		klog.Warningf("since the maximum name length is 80, so it is truncated as (%q)", diskName)
+//
+//
+// Snapshot name must begin with a letter or number, end with a letter, number or underscore,
+// and may contain only letters, numbers, underscores, periods, or hyphens.
+// See https://docs.microsoft.com/en-us/rest/api/compute/snapshots/createorupdate#uri-parameters
+func getValidDiskOrSnapshotName(volumeName, volumeType string) string {
```
andyzhangx: keep `getValidDiskName` and add comment and link that snapshot naming rule is the same

PR #121, opened 2019-07-11
fix: return empty when the specify snapshot id does not exist

ZeroMagic:
**What type of PR is this?**
/kind bug
/kind failing-test

When listing the snapshot, we should return empty when the specify snapshot id does not exist.

Fixes #46 

**Release note**:
```
fix: return empty when the specify snapshot id does not exist
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -518,6 +518,9 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
 	if len(req.GetSnapshotId()) != 0 {
 		snapshot, err := d.getSnapshotByID(ctx, req.GetSnapshotId(), req.SourceVolumeId)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/azuredisk/controllerserver.go
```diff
@@ -515,6 +515,9 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
 	if len(req.GetSnapshotId()) != 0 {
 		snapshot, err := d.getSnapshotByID(ctx, req.GetSnapshotId(), req.SourceVolumeId)
 		if err != nil {
+			if strings.Contains(err.Error(), "none-exist-id") {
```
andyzhangx: how do you know this const string `none-exist-id`, is it a const from somewhere?

ZeroMagic: ok. It is from log. Deeply, it is the message from azure server. And I should use "ResourceNotFound" as identifier.

PR #122, opened 2019-07-11
fix: set the sourceVolumeID when it is empty in request

ZeroMagic:
**What type of PR is this?**
/kind bug
/kind failing-test

If sourceVolumeID is empty in request, we should set it.
With this PR, two failure sanity test will be fixed.

- ListSnapshots [Controller Server] [It] should return appropriate values (no optional values added)
- ListSnapshots [Controller Server] [It] should return snapshots that match the specify snapshot id

Fixes #46 

**Release note**:
```
fix: set the sourceVolumeID when it is empty in request
```


pkg/azuredisk/controllerserver.go
```diff
@@ -650,6 +650,14 @@ func generateCSISnapshot(sourceVolumeID string, snapshot *compute.Snapshot) (*cs
 		return nil, fmt.Errorf("diskSizeGB of snapshot property is nil")
 	}
 
+	if sourceVolumeID == "" {
+		if snapshot.SnapshotProperties != nil &&
```
andyzhangx: `snapshot.SnapshotProperties != nil` is already checked by L639.

PR #123, opened 2019-07-12
fix: return empty when the specify source volume id does not exist

ZeroMagic:
**What type of PR is this?**
/kind bug
/kind failing-test

If the SourceVolumeId in request is not empty and does not match the listed snapshot's SourceVolumeId, we should return empty.

With this PR, we can fix `ListSnapshots [Controller Server] [It] should return empty when the specify source volume id is not exist`

Fixes #46

**Release note**:
```
fix: return empty when the specify source volume id does not exist
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -543,14 +543,23 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
 	if err != nil {
 		return nil, status.Error(codes.Internal, fmt.Sprintf("Unknown list snapshot error: %v", err))
 	}
+	match := false
 	entries := []*csi.ListSnapshotsResponse_Entry{}
 	for _, snapshot := range snapshotListPage.Values() {
+		if req.SourceVolumeId == getSourceVolumeId(&snapshot) {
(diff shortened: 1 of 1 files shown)


pkg/azuredisk/controllerserver.go
```diff
@@ -543,14 +543,23 @@ func (d *Driver) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsReques
...
+		if req.SourceVolumeId == getSourceVolumeId(&snapshot) {
+			match = true
+		}
 		csiSnapshot, err := generateCSISnapshot(req.SourceVolumeId, &snapshot)
 		if err != nil {
 			return nil, fmt.Errorf("failed to generate snapshot entry: %v", err)
 		}
 		entries = append(entries, &csi.ListSnapshotsResponse_Entry{Snapshot: csiSnapshot})
 	}
+	// If the SourceVolumeId in request is not empty and does not match the listed snapshot's SourceVolumeId, we should return empty.
```
andyzhangx: comments better be 
`return empty if SourceVolumeId is not empty and does not match the listed snapshot's SourceVolumeId`
avoid word `we` in comments and make comments short and simple

PR #171, opened 2019-10-23
chore: update sanity tests

### LINKED ISSUE #38: [enhancement] Support for Volume Source (snapshot restore)

This is a new feature in CSI 1.0, need to figure out whether azure disk support cloning first. It should be related to snapshot restore functionality:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/6cd39e09efff106c56c69b3ed242a523e37b044d/pkg/azuredisk/controllerserver.go#L215-L220

```
type VolumeContentSource_SnapshotSource struct {
	// Contains identity information for the existing source snapshot.
	// This field is REQUIRED. Plugin is REQUIRED to support creating
	// volume from snapshot if it supports the capability
	// CREATE_DELETE_SNAPSHOT.
	SnapshotId           string   `protobuf:"bytes,1,opt,name=snapshot_id,json=snapshotId,proto3" json:"snapshot_id,omitempty"`
	XXX_NoUnkeyedLiteral struct{} `json:"-"`
	XXX_unrecognized     []byte   `json:"-"`
	XXX_sizecache        int32    `json:"-"`
}
```

**Additional context**

ZeroMagic:
**What type of PR is this?**
/kind bug
/kind flake

- Update the version of csi-test

- Remove unused dependency "github.com/kubernetes-csi/csi-test"

- According to the `sourceResourceId` format of https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate#create-a-managed-disk-by-copying-a-snapshot. , use the correct sourceResourceId to create volume.

Fixes #38 

**Release note**:
```
update sanity tests
```


ZeroMagic:
@andyzhangx help to retest

ZeroMagic:
Since when I ran `./hack/verify-all.sh`, a binary file `golangci-lint` was generated. So I add `/bin/*` in .gitignore.

ZeroMagic:
All tests have passed. We can continue to review.

pkg/azuredisk/azuredisk.go
```diff
@@ -398,3 +395,18 @@ func getSourceVolumeId(snapshot *compute.Snapshot) string {
...
 }
+
+func getValidCreationData(subscriptionID, resourceGroup, sourceResourceID string) compute.CreationData {
+	if sourceResourceID == "" {
+		return compute.CreationData{
+			CreateOption: compute.Empty,
+		}
+	}
+	if match := diskSnapshotPathRE.FindString(sourceResourceID); match == "" {
+		sourceResourceID = "/subscriptions/" + subscriptionID + "/resourceGroups/" + resourceGroup + "/providers/Microsoft.Compute/snapshots/" + sourceResourceID
```
andyzhangx: I think you may also add basic verification on `sourceResourceID`, e.g. it is only a snapshot name, and if not, return error.
And also may use a diskSnapshotPath const, and use fmt.SPrintf:
```
diskSnapshotPath = `"/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/snapshots/%s" sourceResourceID`
```

pkg/azuredisk/azuredisk.go
```diff
@@ -398,3 +395,18 @@ func getSourceVolumeId(snapshot *compute.Snapshot) string {
 	}
 	return ""
 }
+
+func getValidCreationData(subscriptionID, resourceGroup, sourceResourceID string) compute.CreationData {
```
andyzhangx: may return error if it could not get a valid snapshot URL.

pkg/azuredisk/azuredisk.go
```diff
@@ -321,6 +315,9 @@ func (d *Driver) CreateManagedDisk(ctx context.Context, options *ManagedDiskOpti
 
 	_, err = d.cloud.DisksClient.CreateOrUpdate(ctx, options.ResourceGroup, options.DiskName, model)
 	if err != nil {
+		if strings.Contains(err.Error(), "NotFound") {
```
andyzhangx: provide the error info in the comment

ZeroMagic:
```
compute.DisksClient#CreateOrUpdate: Failure sending request: StatusCode=404 -- Original Error: Code="NotFound" Message="Resource non-existing-snapshot-id is not found."
```

pkg/azuredisk/azuredisk.go
ZeroMagic: done.

pkg/azuredisk/azuredisk.go
ZeroMagic: done.
Here if the provided sourceResourceID is wrong, it will also be jointed. So we just need to judge it according to the amount of '/'. Have a look at some new mistake samples in the unit test.

pkg/azuredisk/azuredisk.go
```diff
@@ -321,6 +321,9 @@ func (d *Driver) CreateManagedDisk(ctx context.Context, options *ManagedDiskOpti
...
+			return "", status.Errorf(codes.NotFound, "the specify snapshot(%s) is not found under rg(%s)", options.SourceResourceID, options.ResourceGroup)
```
andyzhangx: pls also have the error msg in returned error

ZeroMagic: done.

PR #172, opened 2019-10-26
fix golint errors

bysph:
Signed-off-by: sph <shepenghui1@huawei.com>


andyzhangx:
@bysph thanks for your contribution.
I have one question: how did you detect and fix those golint errors? Could you also update the golint verification under `https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/hack/verify-golint.sh` in this PR? Thanks.

PR #186, opened 2019-11-10
feat: support volume resize

### LINKED ISSUE #55: support volume resize in k8s v1.14

CSI support is in alpha in 1.14 release and requires enabling `ExpandCSIVolumes` feature gate.
https://github.com/kubernetes/website/pull/12928
https://github.com/kubernetes/enhancements/issues/556#issuecomment-468760297

https://kubernetes-csi.github.io/docs/external-resizer.html

**Additional context**

ZeroMagic:
**What type of PR is this?**
/kind feature

Since azuredisk could be resized when mounted, we just need to implement `func ControllerExpandVolume()`. The method `func NodeExpandVolume()` does not need to be implemented.

Fixes #55 

**Release note**:
```
support volume resize
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -32,6 +32,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/klog"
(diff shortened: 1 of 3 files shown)


chewong:
Looks like we are experiencing the quota issue in eastus2 again. We might have to move CSI jobs to westus2.

pkg/azuredisk/controllerserver.go
```diff
@@ -435,7 +436,61 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+	}
+
+	capacityBytes := req.GetCapacityRange().GetRequiredBytes()
+	if capacityBytes == 0 {
+		return nil, status.Error(codes.InvalidArgument, "volume capacity range missing in request")
+	}
+	requestSize := *resource.NewQuantity(capacityBytes, resource.BinarySI)
+
+	diskURI := req.GetVolumeId()
+	if err := isValidDiskURI(diskURI); err != nil {
```
andyzhangx: no need to check here since `isManagedDisk` already did the check

pkg/azuredisk/controllerserver.go
```diff
@@ -435,7 +436,61 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+	disk, err := d.cloud.DisksClient.Get(ctx, resourceGroup, diskName)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "could not get the disk(%s) under rg(%s) with error(%v)", diskName, resourceGroup, err)
+	}
+	if disk.DiskProperties.DiskSizeGB == nil {
+		return nil, status.Errorf(codes.Internal, "could not get size of the disk(%s)", diskName)
+	}
+	oldSize := *resource.NewQuantity(int64(*disk.DiskProperties.DiskSizeGB), resource.BinarySI)
+
+	klog.V(2).Infof("expanding azure disk(%s)", diskURI)
```
andyzhangx: `begin to expand azure disk(%s) with new size(%v)`

pkg/azuredisk/controllerserver.go
ZeroMagic:
```go
func isValidDiskURI(diskURI string) error {
	if isManagedDisk(diskURI) {
		if strings.Index(diskURI, "/subscriptions/") != 0 {
			return fmt.Errorf("Inavlid DiskURI: %v, correct format: %v", diskURI, diskURISupportedManaged)
		}
	} else {
```
```go
func isManagedDisk(diskURI string) bool {
	if len(diskURI) > 4 && strings.ToLower(diskURI[:4]) == "http" {
		return false
	}
	return true
}
```
`isManagedDisk` does not check the format. So here we need `isValidDiskURI` to check it.
Then if the URI which is not a ManagedDisk URI can also pass `isValidDiskURI`, we should use `isManagedDisk` to check whether it is a ManagedDisk URI.

pkg/azuredisk/controllerserver.go

pkg/azuredisk/controllerserver.go
andyzhangx: then pls also add the check in `isManagedDisk` func, current `isManagedDisk` func is too simple, we should only need one function(`isManagedDisk`) to check the managed disk format

chewong:
> /retest

Quota issue will be resolved by https://github.com/kubernetes/test-infra/pull/15270/files

pkg/azuredisk/controllerserver.go
ZeroMagic: I think we do not need to change the logic. The duties of the two functions are divided clear now. 
`isManagedDisk ` is just used to distinguish `managedDisk` from `Blob`. It should not return error info.  
`isValidDiskURI ` is responsible for checking the format of both kinds.

So here I just changed the order of checking. Now it seems to be more reasonable.
Firstly, judge the URI if it is a managed URI. Secondly, check its validation.

andyzhangx:
@ZeroMagic have you tried this feature? I think it should also add `external-resizer` in deployment, pls aslo add a doc and e2e test for this feautre.
Added a new issue to track this: https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/201

ZeroMagic:
Oh, it really needs the external-resizer. I will add them.

PR #196, opened 2019-11-25
feat: support volume cloning

### LINKED ISSUE #153: Add docs for Volume Cloning feature

Volume Cloning feature is alpha in v1.15: https://kubernetes-csi.github.io/docs/volume-cloning.html
In driver v0.3.0, we already added volume cloning feature: https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/0deb537f201a376179831171de80f9a10ec2c6cb/pkg/azuredisk/azuredisk.go#L266-L268
need to check whether it works and if it works, need docs

**Additional context**

ZeroMagic:
**What type of PR is this?**
/kind feature

Support `volume cloning` feature. 
Now we can create a managed disk from an existing managed disk.

- [x] add volume cloning functionality
- [x] check whether there are sanity tests to cover this volume cloning functionality
- [x] add e2e tests for this volume cloning functionality
- [x] add a doc about volume cloning
- [x] update helm charts

Fixes #153 #197 

The related sanity tests have passed.

**Release note**:
```
support volume cloning
```

### DIFF
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -66,6 +66,9 @@ const (
 
 	// see https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate#create-a-managed-disk-by-copying-a-snapshot.
 	diskSnapshotPath = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/snapshots/%s"
+
+	// see https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate#create-a-managed-disk-from-an-existing-managed-disk-in-the-same-or-different-subscription.
+	managedDiskPath = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/disks/%s"
 )
 
 var (
@@ -121,6 +124,7 @@ func (d *Driver) Run(endpoint string) {
 			csi.ControllerServiceCapability_RPC_PUBLISH_UNPUBLISH_VOLUME,
 			csi.ControllerServiceCapability_RPC_CREATE_DELETE_SNAPSHOT,
 			csi.ControllerServiceCapability_RPC_LIST_SNAPSHOTS,
+			csi.ControllerServiceCapability_RPC_CLONE_VOLUME,
 			csi.ControllerServiceCapability_RPC_EXPAND_VOLUME,
 		})
 	d.AddVolumeCapabilityAccessModes([]csi.VolumeCapability_AccessMode_Mode{csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER})
@@ -237,6 +241,8 @@ type ManagedDiskOptions struct {
 	DiskMBpsReadWrite string
 	// if SourceResourceID is not empty, then it's a disk copy operation(for snapshot)
 	SourceResourceID string
+	// The type of source
+	SourceType string
 }
 
 //CreateManagedDisk : create managed disk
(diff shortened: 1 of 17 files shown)


pkg/azuredisk/azuredisk.go
```diff
@@ -401,18 +407,36 @@ func getSourceVolumeId(snapshot *compute.Snapshot) string {
...
 	if sourceResourceID == "" {
 		return compute.CreationData{
 			CreateOption: compute.Empty,
 		}, nil
 	}
-	if match := diskSnapshotPathRE.FindString(sourceResourceID); match == "" {
-		sourceResourceID = fmt.Sprintf(diskSnapshotPath, subscriptionID, resourceGroup, sourceResourceID)
+
+	switch sourceType {
+	case sourceSnapshot:
```
andyzhangx: make sourceType as case insensitive?

ZeroMagic: Since sourceType is set by ourselves, there are just two values for it. We do not need to care about case sensitive

ZeroMagic:
There are two sanity tests to cover it.

```
------------------------------
Controller Service [Controller Server] CreateVolume
  should create volume from an existing source volume
  /root/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:782
STEP: reusing connection to CSI driver at unix:///tmp/csi.sock
STEP: creating mount and staging directories
STEP: creating a volume
STEP: creating a volume from source volume
STEP: cleaning up deleting the volume created from source volume
STEP: cleaning up deleting the source volume

• [SLOW TEST:16.370 seconds]
Controller Service [Controller Server]
/root/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
  CreateVolume
  /root/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:380
    should create volume from an existing source volume
    /root/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:782
------------------------------
Controller Service [Controller Server] CreateVolume
  should fail when the volume source volume is not found
  /root/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:817
STEP: reusing connection to CSI driver at unix:///tmp/csi.sock
STEP: creating mount and staging directories
STEP: creating a volume from source snapshot

• [SLOW TEST:5.945 seconds]
Controller Service [Controller Server]
/root/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
  CreateVolume
  /root/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:380
    should fail when the volume source volume is not found
    /root/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:817
------------------------------
```

ZeroMagic:
Since the usage of helm 3 is very different from helm 2, I update the README.md.

In addition, if we want to use volume cloning, the version of external-provisioner is at least v1.4.0. After updating the templates, we can specify the images' tag in values.yaml without having to modify the templates every time

test/e2e/suite_test.go
```diff
@@ -117,6 +117,15 @@ var _ = AfterSuite(func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(strings.HasSuffix(projectRoot, "azuredisk-csi-driver")).To(Equal(true))
 
+		log.Println("===================azuredisk log===================")
```
andyzhangx: it's better than print logs if test failed

ZeroMagic: Here I found no function to judge whether the tests failed or not. So no matter what happens, I will print the log.

ZeroMagic:
All tests have passed. We can continue to review.

PR #227, opened 2019-12-25
ignore case for managedDiskPathRE and diskSnapshotPathRE

levimm:
**What type of PR is this?**
/kind bug

Fix the volume not found bug following [example](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/deploy/example/e2e_usage.md#option2-azuredisk-static-provisioninguse-an-existing-azure-disk).
The disk URI regex validation should be case insensitive such as `subscriptions`, `resourcegroups`, `Microsoft.Compute`...

**Release note**:
```
ignore case for managedDiskPathRE and diskSnapshotPathRE
```

### DIFF
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -72,9 +72,9 @@ const (
 )
 
 var (
-	managedDiskPathRE       = regexp.MustCompile(`.*/subscriptions/(?:.*)/resourceGroups/(?:.*)/providers/Microsoft.Compute/disks/(.+)`)
+	managedDiskPathRE       = regexp.MustCompile(`(?i).*/subscriptions/(?:.*)/resourceGroups/(?:.*)/providers/Microsoft.Compute/disks/(.+)`)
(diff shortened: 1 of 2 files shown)


pkg/azuredisk/azuredisk.go
```diff
@@ -72,9 +72,9 @@ const (
 )
 
 var (
-	managedDiskPathRE       = regexp.MustCompile(`.*/subscriptions/(?:.*)/resourceGroups/(?:.*)/providers/Microsoft.Compute/disks/(.+)`)
+	managedDiskPathRE       = regexp.MustCompile(`(?i).*/subscriptions/(?:.*)/resourceGroups/(?:.*)/providers/Microsoft.Compute/disks/(.+)`)
```
andyzhangx: add `(?i)` means ignore case? and could you add some test cases into `azuredisk_test.go` in this PR?

levimm: ut added

PR #235, opened 2019-12-29
feat: BYOK(SSE+CMK) support

andyzhangx:
**What type of PR is this?**
/kind feature

feat: BYOK(SSE+CMK) support
Actually BYOK is already supported in in-tree azure disk driver, this PR just switch to use `CreateManagedDisk` in azure cloud provider, and then BYOK would be also supported in CSI driver.

Fixes #

**Release note**:
```
feat: BYOK(SSE+CMK) support
```

### DIFF
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -21,25 +21,21 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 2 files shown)


andyzhangx:
/hold
sanity test failure due to return error is not well defined by create disk function, need to address that issue on azure cloud provider first.

PR #247, opened 2020-01-06
Introduce CSI proxy calls

### LINKED ISSUE #45: CSI driver support on Windows
**Is your feature request related to a problem? Please describe.**

Currently CSI driver is linux based, cannot run on Windows. The main difficulty is design difference between Linux and Windows, all devices are files while it's not true for Windows. 
Here is a issue link in moby talking about this issue: https://github.com/moby/moby/issues/38479

We could make the attacher and azure disk csi driver running as a standalone process on Windows, that could work. (could refer to kube-proxy and kubelet implementation in aks-engine: https://github.com/Azure/aks-engine/blob/master/parts/k8s/kuberneteswindowssetup.ps1#L129-L130)
For the provision part, it could still run on Linux(on master node?). That would make it a little easier to implement.

**Additional context**

https://github.com/kubernetes/enhancements/blob/master/keps/sig-windows/20190714-windows-csi-support.md

[node-driver-registrar support on Windows](https://github.com/kubernetes-csi/node-driver-registrar/pull/47)

kkmsft:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test

/kind feature

> /kind flake

Changes to add calls into CSI proxy for privileged operations in windows.

Fixes #45

TODO: The format and mount options to the volume will be added in a future PR..

**Release note**:
```
Windows support -integrate with CSI proxy (https://github.com/kubernetes-csi/csi-proxy).
```


pkg/azuredisk/azuredisk.code-workspace
```diff
@@ -0,0 +1,7 @@
+{
```
andyzhangx: remove this?

kkmsft:
> I think you may also run `dep ensure -v -update` to vendoring new packages from `csi-proxy`

Thanks, Andy ! Once the PRs in the csi-proxy repo merges, will update the dependancies.

pkg/azuredisk/azure_common_windows.go
```diff
@@ -19,100 +19,69 @@ limitations under the License.
...
 }
 
-// search Windows disk number by LUN
-func findDiskByLun(lun int, iohandler ioHandler, exec mount.Exec) (string, error) {
-	cmd := `Get-Disk | select number, location | ConvertTo-Json`
-	output, err := exec.Run("powershell", "/c", cmd)
-	if err != nil {
-		klog.Errorf("Get-Disk failed in findDiskByLun, error: %v, output: %q", err, string(output))
-		return "", err
+// TODO: Return error from scsiHostRescan. Modify for Linux as well.
```
andyzhangx: so even `scsiHostRescan` return error, we should still continue find disk lun, right?

pkg/mounter/safe_mounter_windows.go
```diff
@@ -46,8 +92,36 @@ func (mounter *CSIProxyMounter) IsMountPointMatch(mp mount.MountPoint, dir strin
 	return mp.Path == dir
 }
 
-func (mounter *CSIProxyMounter) IsLikelyNotMountPoint(file string) (bool, error) {
-	return false, fmt.Errorf("IsLikelyNotMountPoint not implemented for CSIProxyMounter")
+func (mounter *CSIProxyMounter) IsPathExists(path string) (bool, error) {
+	isExistsRequest := &fsv1alpha1.PathExistsRequest{
+		Path: path,
+	}
+	isExistsResponse, err := mounter.FsClient.PathExists(context.Background(), isExistsRequest)
```
andyzhangx: why not write like this:
```
return mounter.FsClient.PathExists(context.Background(), &fsv1alpha1.PathExistsRequest{Path: path})
```

And maybe don't need this func?

pkg/azuredisk/azure_common_windows.go
kkmsft: Yep. Actually the TODO was unnecessary here.

pkg/mounter/safe_mounter_windows.go
kkmsft: the return needs to be separated out. So could not move it to single line and remove the call, but have been able to reduce the code.

PR #250, opened 2020-01-09
chore: remove nodeid restriction

andyzhangx:
**What type of PR is this?**
/kind cleanup

remove nodeid restriction

Fixes #

**Release note**:
```
none
```

### DIFF
--- a/pkg/azurediskplugin/main.go
+++ b/pkg/azurediskplugin/main.go
@@ -54,8 +54,8 @@ func main() {
 	}
 
 	if *nodeID == "" {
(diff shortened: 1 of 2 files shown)


pkg/azurediskplugin/main.go
```diff
@@ -54,8 +54,7 @@ func main() {
 	}
 
 	if *nodeID == "" {
-		klog.Error("--nodeid is a required parameter")
-		os.Exit(1)
+		klog.Warning("nodeid is empty")
```
feiskyer: Add a comment of when nodeID is allowed to be empty?

PR #255, opened 2020-01-15
feat: add inline volume support

### LINKED ISSUE #128: inline volume support

https://kubernetes-csi.github.io/docs/ephemeral-local-volumes.html
https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md

Need Generic Ephemeral Inline Volumes support which is only alpha state

**Additional context**

ZeroMagic:
**What type of PR is this?**
/kind documentation
/kind feature

- [x] add inline volume support

- [x] add the inline feature doc

- [ ] add the e2e test

Since there isn't any related case in sanity tests, we also need an e2e test to ensure the feature.

Fixes #128 

**Release note**:
```
add inline volume support
```

### DIFF
--- a/deploy/example/inline/nginx-pod-azuredisk-inline.yaml
+++ b/deploy/example/inline/nginx-pod-azuredisk-inline.yaml
@@ -1,0 +1,26 @@
+kind: Pod
+apiVersion: v1
+metadata:
+  name: nginx-azuredisk-inline
+spec:
+  nodeSelector:
+    beta.kubernetes.io/os: linux
+  containers:
+  - image: nginx
+    name: nginx-azuredisk
+    command:
+    - "/bin/sh"
+    - "-c"
+    - while true; do echo $(date) >> /mnt/azuredisk/outfile; sleep 1; done
+    volumeMounts:
+    - name: azuredisk02
+      mountPath: "/mnt/azuredisk"
+  volumes:
+  - name: azuredisk02
+    csi:
+      driver: disk.csi.azure.com
+      volumeAttributes:
+        capacity: 10Gi # default is 1Gi
+        storageAccountType: Standard_LRS # default is Standard_LRS
+      readOnly: false  # default is false
+      fsType: ext4 # default is ext4
--- a/pkg/azuredisk/nodeserver.go
+++ b/pkg/azuredisk/nodeserver.go
@@ -44,6 +44,10 @@ const (
 	defaultLinuxFsType      = "ext4"
 	defaultWindowsFsType    = "ntfs"
 	defaultAzureVolumeLimit = 16
+
+	ephemeralVolumeIdTemplate = "ephemeral-%s"
+	// see https://github.com/kubernetes/api/blob/master/storage/v1beta1/types.go#L294
+	ephemeralVolumeKey = "csi.storage.k8s.io/ephemeral"
 )
 
 // store vm size list in current region
@@ -186,6 +190,11 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 		return nil, status.Error(codes.InvalidArgument, "Volume ID missing in request")
 	}
 
+	ephemeral := req.GetVolumeContext()[ephemeralVolumeKey] == "true"
+	if ephemeral {
+		return d.nodePublishEphemeralVolume(ctx, req)
+	}
+
 	source := req.GetStagingTargetPath()
 	if len(source) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "Staging target not provided")
@@ -247,12 +256,24 @@ func (d *Driver) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublish
 	targetPath := req.GetTargetPath()
 	volumeID := req.GetVolumeId()
 
+	ephemeral := false
+	// In the persistent scenario, the volumeID is like /subscriptions/xxx/resourceGroups/xxx/providers/Microsoft.Compute/disks/xxx.
+	// In the inline scenario, according to the volumeHandle(https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_mounter.go#L450-L454),
+	// the volumeID is like "csi-xxxxxx".
+	if match := managedDiskPathRE.FindString(volumeID); match == "" && strings.Contains(volumeID, "csi-") {
+		ephemeral = true
+	}
+
 	klog.V(2).Infof("NodeUnpublishVolume: unmounting volume %s on %s", volumeID, targetPath)
 	err := d.mounter.Unmount(req.GetTargetPath())
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	klog.V(2).Infof("NodeUnpublishVolume: unmount volume %s on %s successfully", volumeID, targetPath)
+
+	if ephemeral {
+		return d.nodeUnpublishEphemeralVolume(ctx, req)
+	}
 
 	return &csi.NodeUnpublishVolumeResponse{}, nil
 }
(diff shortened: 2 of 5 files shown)


pkg/azuredisk/nodeserver.go
```diff
@@ -272,13 +278,22 @@ func (d *Driver) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublish
 	targetPath := req.GetTargetPath()
 	volumeID := req.GetVolumeId()
 
+	ephemeral := false
+	if match := managedDiskPathRE.FindString(volumeID); match == "" {
```
ZeroMagic: In the persistent scenario, the volumeID is like `/subscriptions/xxx/resourceGroups/xxx/providers/Microsoft.Compute/disks/xxx`. But in the inline scenario, it is like `csi-fcfeb6b3905ea15047fd0a6a6c724049eae9e022ba0bd02c4af00feac400ac7e`

andyzhangx: I think you still need to format check on `volumeID` since it could be invalid format, and there is the format doc link of inline volume?

pkg/azuredisk/nodeserver.go
```diff
@@ -506,3 +521,117 @@ func (d *Driver) getBlockSizeBytes(devicePath string) (int64, error) {
...
+	// create volume
+	diskURI, err := d.cloud.CreateManagedDisk(volumeOptions)
+	if err != nil {
+		if strings.Contains(err.Error(), "NotFound") {
+			return nil, status.Error(codes.NotFound, err.Error())
+		}
+		return nil, err
+	}
+	defer func() {
+		if err != nil {
```
andyzhangx: if there is error, it should stop and report error, not delete disk

pkg/azuredisk/nodeserver.go
```diff
@@ -187,6 +188,11 @@ func (d *Driver) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 		return nil, status.Error(codes.InvalidArgument, "Volume ID missing in request")
 	}
 
+	ephemeral := req.GetVolumeContext()["csi.storage.k8s.io/ephemeral"] == "true"
```
andyzhangx: where is this defined? will it change in the future? looks  like it would be broken easily.

ZeroMagic: See the last sentence of the paragraph. 
https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md#support-for-inline-csi-volumes

The parameter is handled in Kubernetes. https://github.com/kubernetes/api/blob/master/storage/v1beta1/types.go#L279-L304

pkg/azuredisk/nodeserver.go
ZeroMagic: Here is the generating process.
https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_mounter.go#L450-L454

The format totally complies with the azuredisk naming convention.

pkg/azuredisk/nodeserver.go
andyzhangx: pls add `csi.storage.k8s.io/ephemeral` as a const and add links in the comments

pkg/azuredisk/nodeserver.go
andyzhangx: at least check prefix `csi-`, and also add link in the comment, thanks

pkg/azuredisk/nodeserver.go
andyzhangx: and I see there is already delete disk operation in `nodeUnpublishEphemeralVolume` func

ZeroMagic: @pohly I am implementing the inline feature. Here we have a question. If the disk is not be deleted when some errors interrupt the `NodePublishVolume`, will the disk leak? Or will it be handled by k8s?

pohly: This was discussed recently on Slack: https://kubernetes.slack.com/archives/C8EJ01Z46/p1578650595018200

The gist of that that discussion is that while Kubernetes may call `NodePublishVolume` again depending the error code that it returns, a CSI driver cannot count on that and therefore should try to clean up as much as possible after a failure and before returning. For example, if the pod gets deleted after a failed `NodePublishVolume`, then `NodeUnpublishVolume` will not be called.

ZeroMagic: It is a detailed discussion. Thanks pretty for your guide!😊@pohly

pkg/azuredisk/nodeserver.go
```diff
@@ -506,3 +521,117 @@ func (d *Driver) getBlockSizeBytes(devicePath string) (int64, error) {
...
+		return nil, err
+	}
+	klog.V(2).Infof("nodePublishEphemeralVolume successfully")
+
+	return &csi.NodePublishVolumeResponse{}, nil
+}
+
+func (d *Driver) nodeUnpublishEphemeralVolume(ctx context.Context, req *csi.NodeUnpublishVolumeRequest) (*csi.NodeUnpublishVolumeResponse, error) {
+	volumeId := req.GetVolumeId()
+	diskName := fmt.Sprintf("ephemeral-%s", volumeId)
```
andyzhangx: pls use a const string for this,e.g.
```
	// see https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate#create-a-managed-disk-by-copying-a-snapshot.
	diskSnapshotPath = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/snapshots/%s"
```

pkg/azuredisk/nodeserver.go
```diff
@@ -506,3 +521,117 @@ func (d *Driver) getBlockSizeBytes(devicePath string) (int64, error) {
...
+		size = defaultDiskSize
+	}
+
+	storageAccountType := req.GetVolumeContext()["storageAccountType"]
+	skuName, err := normalizeStorageAccountType(storageAccountType)
+	if err != nil {
+		return nil, err
+	}
+
+	ephemeralVolumeId := fmt.Sprintf("ephemeral-%s", volumeId)
```
andyzhangx: pls use a const string for this,e.g.
```
	// see https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate#create-a-managed-disk-by-copying-a-snapshot.
	diskSnapshotPath = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/snapshots/%s"
```

pkg/azuredisk/nodeserver.go
```diff
@@ -506,3 +521,117 @@ func (d *Driver) getBlockSizeBytes(devicePath string) (int64, error) {
...
+	klog.V(2).Infof("nodePublishEphemeralVolume successfully")
+
+	return &csi.NodePublishVolumeResponse{}, nil
+}
+
+func (d *Driver) nodeUnpublishEphemeralVolume(ctx context.Context, req *csi.NodeUnpublishVolumeRequest) (*csi.NodeUnpublishVolumeResponse, error) {
+	volumeId := req.GetVolumeId()
+	diskName := fmt.Sprintf("ephemeral-%s", volumeId)
+
+	resourceGroup := d.cloud.ResourceGroup
```
andyzhangx: why use a new var here, just use `d.cloud.ResourceGroup`

pkg/azuredisk/nodeserver.go
```diff
@@ -506,3 +521,117 @@ func (d *Driver) getBlockSizeBytes(devicePath string) (int64, error) {
...
+
+func (d *Driver) nodePublishEphemeralVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
+	volumeId := req.GetVolumeId()
+
+	capacity := req.GetVolumeContext()["capacity"]
+	size, err := strconv.Atoi(strings.TrimSuffix(capacity, "Gi"))
+	if err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("Unable to parse capacity %v", err))
+	}
+	if size == 0 {
```
andyzhangx: if it's 0, then return error

pkg/azuredisk/nodeserver.go
ZeroMagic: done.

pkg/azuredisk/nodeserver.go
ZeroMagic: done

pkg/azuredisk/nodeserver.go
ZeroMagic: done

pkg/azuredisk/nodeserver.go
ZeroMagic: done

pkg/azuredisk/nodeserver.go
ZeroMagic: done

pkg/azuredisk/nodeserver.go
```diff
@@ -489,3 +510,111 @@ func (d *Driver) ensureBlockTargetFile(target string) error {
...
+		Parameters:         parameters,
+	}
+	createVolumeResp, err := d.CreateVolume(ctx, createVolumeReq)
+	if err != nil {
+		return nil, err
+	}
+	diskURI := createVolumeResp.Volume.VolumeId
+	defer func() {
+		if err != nil {
+			if err1 := d.cloud.DeleteManagedDisk(diskURI); err1 != nil {
```
andyzhangx: could you use `DeleteVolume` here?

pkg/azuredisk/nodeserver.go
```diff
@@ -489,3 +510,111 @@ func (d *Driver) ensureBlockTargetFile(target string) error {
...
+		return nil, status.Errorf(codes.Internal, "Unable to parse capacity %v", err)
+	}
+	if size <= 0 {
+		return nil, status.Errorf(codes.Internal, "size(%v) is illegal, please provide the positive size value", size)
+	}
+
+	parameters := make(map[string]string)
+	parameters["skuname"] = req.GetVolumeContext()["storageAccountType"]
+
+	klog.V(2).Infof("nodePublishEphemeralVolume: creating the managed disk(%s)", ephemeralVolumeId)
```
andyzhangx: `klog.V(2).Infof("nodePublishEphemeralVolume: creating volume(%s)", ephemeralVolumeId`

pkg/azuredisk/nodeserver.go
```diff
@@ -489,3 +510,111 @@ func (d *Driver) ensureBlockTargetFile(target string) error {
...
+	size, err := strconv.Atoi(strings.TrimSuffix(capacity, "Gi"))
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Unable to parse capacity %v", err)
+	}
+	if size <= 0 {
+		return nil, status.Errorf(codes.Internal, "size(%v) is illegal, please provide the positive size value", size)
+	}
+
+	parameters := make(map[string]string)
+	parameters["skuname"] = req.GetVolumeContext()["storageAccountType"]
```
andyzhangx: could you just pass all VolumeContext as parameter since there could be other parameters defined in VolumeContext and it's case insensitive: 
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/e1c2613898996dbaf8223c866c791a14335598c2/pkg/azuredisk/controllerserver.go#L105-L113

pkg/azuredisk/nodeserver.go
```diff
@@ -489,3 +510,111 @@ func (d *Driver) ensureBlockTargetFile(target string) error {
...
+		PublishContext:    controllerPublishResp.PublishContext,
+		StagingTargetPath: req.GetTargetPath(),
+		VolumeCapability:  req.GetVolumeCapability(),
+		VolumeContext:     req.GetVolumeContext(),
+	}
+	_, err = d.NodeStageVolume(ctx, nodeStageVolumeReq)
+	if err != nil {
+		return nil, err
+	}
+	klog.V(2).Infof("nodePublishEphemeralVolume successfully")
```
andyzhangx: always provide volume id in logging

pkg/azuredisk/nodeserver.go
```diff
@@ -489,3 +510,111 @@ func (d *Driver) ensureBlockTargetFile(target string) error {
...
+
+	return &csi.NodePublishVolumeResponse{}, nil
+}
+
+func (d *Driver) nodeUnpublishEphemeralVolume(ctx context.Context, req *csi.NodeUnpublishVolumeRequest) (*csi.NodeUnpublishVolumeResponse, error) {
+	volumeId := req.GetVolumeId()
+	diskName := fmt.Sprintf(ephemeralVolumeIdTemplate, volumeId)
+
+	_, diskURI, err := d.cloud.GetDisk(d.cloud.ResourceGroup, diskName)
```
andyzhangx: there should be already GetDisk logic in `ControllerUnpublishVolume`, so it's not necessary now.

pkg/azuredisk/nodeserver.go
```diff
@@ -489,3 +510,111 @@ func (d *Driver) ensureBlockTargetFile(target string) error {
...
+	_, err = d.ControllerUnpublishVolume(ctx, controllerUnpublishVolumeReq)
+	if err != nil {
+		return nil, err
+	}
+
+	err = d.cloud.DeleteManagedDisk(diskURI)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "could not delete azuredisk(%s): %v", diskURI, err)
+	}
+	klog.V(2).Infof("nodeUnPublishEphemeralVolume successfully")
```
andyzhangx: always provide volume id and diskURI in logging

ZeroMagic: done

pkg/azuredisk/nodeserver.go
ZeroMagic: done

pkg/azuredisk/nodeserver.go
ZeroMagic: done

pkg/azuredisk/nodeserver.go
ZeroMagic: done

pkg/azuredisk/nodeserver.go
ZeroMagic: done

pkg/azuredisk/nodeserver.go
ZeroMagic: `diskURI ` is still needed in the other part of the function.

pkg/azuredisk/nodeserver.go
```diff
@@ -489,3 +510,111 @@ func (d *Driver) ensureBlockTargetFile(target string) error {
...
+	controllerUnpublishVolumeReq := &csi.ControllerUnpublishVolumeRequest{
+		VolumeId: diskURI,
+		NodeId:   d.NodeID,
+	}
+	_, err = d.ControllerUnpublishVolume(ctx, controllerUnpublishVolumeReq)
+	if err != nil {
+		return nil, err
+	}
+
+	err = d.cloud.DeleteManagedDisk(diskURI)
```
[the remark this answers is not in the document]
ZeroMagic: done. It is better to use `DeleteVolume` func.

ZeroMagic:
Maybe we can support both situations. If the user specifies an existing diskURI, we do not invoke `CreateVolume` an `DeleteVolume`. Otherwise, we will create an ephemeral disk which lifecycle is the same as the pod.

ZeroMagic:
The static provisioning scenario could not be supported here. 
It is ok to execute `NodePublishVolume` with the existing diskURI. But during node unpublishing, `NodeUnpublishVolume` could not be aware of the diskURI. 

There are just two fields in `NodeUnpublishVolumeReq`, `VolumeId` which is created by the Kubernetes and passed to the driver by the kubelet, and `VolumePath`. 

If we want to execute `NodeUnpublishVolume` for inline volume, the following two steps are needed.
1. use `VolumePath` to unmount inline volume.
2. use `diskURI` to detach the inline volume. But we just know the `VolumeId`.

deploy/example/inline/nginx-pod-azuredisk-inline.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - while true; do echo $(date) >> /mnt/azuredisk/outfile; sleep 1; done
+    volumeMounts:
+    - name: azuredisk02
+      mountPath: "/mnt/azuredisk"
+  volumes:
+  - name: azuredisk02
+    csi:
+      driver: disk.csi.azure.com
+      volumeAttributes:
+        capacity: 10Gi # default is 1Gi
```
andyzhangx: only add two fields(`diskURI`, `fsType`) in `volumeAttributes`, and then we could create a new volumeid (exactly same as `diskURI`), this volumeid would be used in both node publish and node unpublish?

BTW, is it possible add a `volumeHandle` field like following, then we could have exact same logic as pv/pvc static provisioning:
```
  csi:
    driver: disk.csi.azure.com
    readOnly: false
    volumeHandle: /subscriptions/{sub-id}/resourcegroups/{group-name}/providers/microsoft.compute/disks/{disk-id}
    volumeAttributes:
      fsType: ext4
```

PR #265, opened 2020-01-23
chore: set replica of controller as 2

andyzhangx:
**What type of PR is this?**
/kind feature

set replica of controller as 2

Fixes #

**Release note**:
```
none
```


charts/latest/azuredisk-csi-driver/values.yaml
```diff
@@ -39,4 +39,4 @@ rbac:
   create: true
 
 controller:
-  replicas: 3
+  replicas: 2
```
ZeroMagic: Currently, we just added the follow rule in `azuredisk-external-provisioner-role`.
```
- apiGroups: ["coordination.k8s.io"]
  resources: ["leases"]
  verbs: ["get", "watch", "list", "delete", "update", "create"]
```
`azuredisk-external-attacher-role`, `azuredisk-external-snapshotter-role` and `azuredisk-external-resizer-role` also need the rule. We can add them in this pr. 
Then we can judge the replica num 2 whether it is suitable.

ZeroMagic: This rule is used for leader election

andyzhangx: good catch, fixed.

andyzhangx:
/hold
also need to update rbac setting of chart config

PR #285, opened 2020-02-14
fix: find lun with timeout issue

andyzhangx:
**What type of PR is this?**
/kind bug

PR(https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/281) does not fix find lun issue in a complete PR, when there is timeout, should report error other than just return.

cc @kosta709

Fixes #

**Release note**:
```
fix: find lun with timeout issue
```


pkg/azuredisk/nodeserver.go
```diff
@@ -132,9 +132,10 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 		fstype = volContextFSType
 	}
 
-	source, lun, err := d.findDiskAndLun(devicePath)
+	// devicePath is actually a lun num
```
ZeroMagic: Since the device is always a lun num, it seems that we can remove the return value `lun` in func `findDiskAndLun`.

andyzhangx: the in-tree disk driver still uses this function and need lun num, I think we could keep this function now, otherwise we need maintain two different functions.

andyzhangx: BTW, I will make same fix into k/k when this PR merged.

andyzhangx:
/hold
Let me make the code logic more clear

pkg/azuredisk/nodeserver.go
```diff
@@ -538,10 +539,9 @@ func (d *Driver) findDiskAndLun(devicePath string) (string, int32, error) {
 		// wait until timeout
 		return false, nil
 	})
-	if err != nil {
-		return "", -1, err
+	if newDevicePath == "" {
```
andyzhangx: the logic is more clear now  @ZeroMagic

ZeroMagic:
```go
if newDevicePath, err = findDiskByLun(int(lun), io, d.mounter.Exec); err != nil {
			return false, fmt.Errorf("azureDisk - findDiskByLun(%v) failed with error(%s)", lun, err)
		}
```
If the error existed, the newDevicePath is still "", and the error will be changed here.

andyzhangx: good catch, fixed by:
```
	if err == nil && newDevicePath == "" {
		err = fmt.Errorf("azureDisk - findDiskByLun(%v) failed within timeout", lun)
	}
```

andyzhangx:
/hold cancel
@ZeroMagic PTAL, thanks

PR #293, opened 2020-02-15
fix yaml format

andyzhangx:
**What type of PR is this?**
 /kind documentation

fix yamllint format

Fixes #

**Release note**:
```
none
```


deploy/example/resizing/storageclass-azuredisk-csi-resize.yaml
```diff
@@ -1,12 +0,0 @@
-apiVersion: storage.k8s.io/v1
-kind: StorageClass
-metadata:
-  name: disk.csi.azure.com
-provisioner: disk.csi.azure.com
-allowVolumeExpansion: true
```
ZeroMagic: If `allowVolumeExpansion ` does not set to true, volume expansion feature will not work.

andyzhangx: I have set this parameter as true in `deploy/example/storageclass-azuredisk-csi.yaml`, don't need to create a new yaml for this.

PR #298, opened 2020-02-20
test: add in-tree-migration.json

andyzhangx:
**What type of PR is this?**
/kind test

This PR add in-tree-migration.json for migration from in-tree driver to CSI driver. Later, we also need to add following config, while following config depends on https://github.com/kubernetes/kubernetes/pull/88014 which is merged in `v1.18.0-beta.0`, `v1.17.4`, so I would make following config change when we switched to use `v1.18.0`. In current stage, let's use this config.
```
"kubernetesConfig": {
    "kubeletConfig": {
        "--feature-gates": "CSIMigration=true,CSIMigrationAzureDisk=true"
    }
```

Fixes #

/assign @chewong 

**Release note**:
```
none
```

### DIFF
--- a/test/e2e/manifest/in-tree-migration.json
+++ b/test/e2e/manifest/in-tree-migration.json
@@ -1,0 +1,64 @@
+{
+    "apiVersion": "vlabs",
+    "location": "westus2",
+    "properties": {
+        "orchestratorProfile": {
+            "orchestratorType": "Kubernetes",
+            "orchestratorRelease": "1.17",
+            "kubernetesConfig": {
+                "networkPolicy": "none",
+                "cloudProviderRateLimitQPS": 6,
+                "cloudProviderRateLimitBucket": 20,
+                "controllerManagerConfig": {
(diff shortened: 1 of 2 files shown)


chewong:
Are we running e2e test on this cluster configuration without building and installing the driver?

andyzhangx:
> Are we running e2e test on this cluster configuration without building and installing the driver?

@chewong you are right, migration tests should run in cluster with our csi driver installed, with in-tree driver storage class. Let me change the PR a little.

andyzhangx:
@chewong I have verified that with following aks-engine cluster definition, basic test on in-tree azure disk driver to csi driver works well on `1.18.0-beta.1`, `1.17.4` should also works well.

```
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes",
      "orchestratorVersion": "1.18.0-beta.1",
      "kubernetesConfig": {
         "controllerManagerConfig": {
           "--feature-gates": "CSIInlineVolume=true,LocalStorageCapacityIsolation=true,ServiceNodeExclusion=true,CSIMigration=true,CSIMigrationAzureDisk=true"
         },
        "kubeletConfig": {
          "--feature-gates": "CSIMigration=true,CSIMigrationAzureDisk=true"
        }
      }
    },
```

chewong:
I will start setting the presubmit job.

Issue #283, opened 2020-02-13
Containerize sanity test and deploy it to an Azure cluster


chewong:
**Is your feature request related to a problem?/Why is this needed**

    
we hit a test failure when running StageVolume test since that StageVolume is invoked by a container on google cloud.

**Describe the solution you'd like in detail**

Containerize the sanity test and run it on an Azure cluster.

**Describe alternatives you've considered**

**Additional context**

/assign

chewong:
@andyzhangx I tried to containerize the sanity test and ran it as a pod on an Azure cluster but `should fail when volume does not exist on the specified path` is still failing.
```
Node Service NodeGetVolumeStats
  should fail when volume does not exist on the specified path
  /go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:458
STEP: reusing connection to CSI driver at unix:///tmp/csi.sock
STEP: creating mount and staging directories
I0213 21:16:17.888545       8 utils.go:112] GRPC call: /csi.v1.Identity/GetPluginCapabilities
I0213 21:16:17.888555       8 utils.go:113] GRPC request:
I0213 21:16:17.888559       8 identityserver.go:59] Using default capabilities
I0213 21:16:17.888562       8 utils.go:119] GRPC response: capabilities:<service:<type:CONTROLLER_SERVICE > > capabilities:<service:<type:VOLUME_ACCESSIBILITY_CONSTRAINTS > >
I0213 21:16:17.888791       8 utils.go:112] GRPC call: /csi.v1.Controller/ControllerGetCapabilities
I0213 21:16:17.888800       8 utils.go:113] GRPC request:
I0213 21:16:17.888803       8 controllerserver.go:448] Using default ControllerGetCapabilities
I0213 21:16:17.888809       8 utils.go:119] GRPC response: capabilities:<rpc:<type:CREATE_DELETE_VOLUME > > capabilities:<rpc:<type:PUBLISH_UNPUBLISH_VOLUME > > capabilities:<rpc:<type:CREATE_DELETE_SNAPSHOT > > capabilities:<rpc:<type:LIST_SNAPSHOTS > > capabilities:<rpc:<type:CLONE_VOLUME > > capabilities:<rpc:<type:EXPAND_VOLUME > >
I0213 21:16:17.889067       8 utils.go:112] GRPC call: /csi.v1.Node/NodeGetCapabilities
I0213 21:16:17.889084       8 utils.go:113] GRPC request:
[...]

PR #304, opened 2020-02-28
fix: correct the behavior of deleting the target direcotory

ZeroMagic:
**What type of PR is this?**
/kind bug

In CSI Spec, it requires that the SP MUST delete the file or directory it created at this path.
Ref: https://github.com/container-storage-interface/spec/blob/master/spec.md#nodeunpublishvolume

Fixes #

The PR will solve the problem in the issue #283.

**Release note**:
```

```


andyzhangx:
@ZeroMagic could you also port this PR to azure file & blobfuse driver? thanks.

ZeroMagic:
> could you also port this PR to azure file & blobfuse driver? thanks.

ok, I will

PR #327, opened 2020-03-25
doc: shared disk doc

andyzhangx:
**What type of PR is this?**
 /kind documentation

shared disk doc

Fixes #

cc @VybavaRamadoss

**Release note**:
```
none
```

### DIFF
--- a/deploy/example/sharedisk/README.md
+++ b/deploy/example/sharedisk/README.md
@@ -1,0 +1,81 @@
+# Shared disk(Multi-node ReadWrite)
+
+## Feature Status: Alpha
+
+[Azure shared disks](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/disks-shared) (preview) is a new feature for Azure managed disks that enables attaching an Azure managed disk to multiple virtual machines (VMs) simultaneously. Attaching a managed disk to multiple VMs allows you to either deploy new or migrate existing clustered applications to Azure.
+
+### Note
+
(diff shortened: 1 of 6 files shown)


deploy/example/sharedisk/README.md
```diff
@@ -0,0 +1,77 @@
...
+      volumes:
+        - name: azuredisk
+          persistentVolumeClaim:
+            claimName: pvc-azuredisk
+```
+
+3. Check block device in pod
+
+```console
+# kubectl exec -it deployment-sharedisk-7454978bc6-xh7jp bash
```
kkmsft: Since this is shared disk doc, it would be great to have an example of the how the contents are written from replica and then can be ready from other replica. ie. on  replica designated as write replica - dd if=<some real content> and then on read replica dd if=/dev/sdx and see the content.

deploy/example/sharedisk/README.md
```diff
@@ -0,0 +1,77 @@
...
+        app: nginx
+      name: deployment-azuredisk
+    spec:
+      containers:
+        - name: deployment-azuredisk
+          image: nginx
+          volumeDevices:
+            - name: azuredisk
+              devicePath: /dev/sdx
+          volumeMounts:
```
kkmsft: Is this required for this deployment. Perhaps we can remove this to avoid any confusion for users.

andyzhangx: good catch, fixed.

deploy/example/sharedisk/README.md
andyzhangx: added a note, from driver level, we cannot grantee single write, multi read

andyzhangx:
We are running out of testing quota again, merge this doc change first to reduce quota.

PR #354, opened 2020-04-09
Allow disabling controller/node service via command line flags

rfranzke:
**What type of PR is this?**
/kind feature

This PR adds two new command line flags `--run-controller-service` and `--run-node-service` (both default to `true`) which allow to particularly disable individual services if not required.

This enables the use-case of running the CSI controllers (csi-provisioner, csi-attacher, etc., + the driver controller) separately outside of the cluster they are serving. The disk plugin (node service part of CSI driver) still runs inside the cluster.

**Release note**:
```
It is now possible to disable the controller service by setting `--run-controller-service=false`. Similarly, it is possible to disable the node service by setting `--run-node-service=false`. The latter enables running the controller server of the Azure Disk driver separately/outside of the cluster it is serving.
```

### DIFF
--- a/docs/driver-services.md
+++ b/docs/driver-services.md
@@ -1,0 +1,12 @@
+### `disk.csi.azure.com` driver services
+ 
+Traditionally, you run the CSI controllers with the Azure Disk driver in the same Kubernetes cluster.
+Though, there may be cases where you will only want to run a subset of the available driver services (for example, one scenario is running the controllers outside of the cluster they are serving (while the Azure Disk driver still runs inside the served cluster), but there might be others scenarios).
+The CSI driver consists out of these services:
+
+* The **controller** service starts the GRPC server that serves `CreateVolume`, `DeleteVolume`, etc. It is depending on the Azure cloud config + credentials and talks with the Azure API.
+* The **identity** service is responsible to provide identity services like capability information of the CSI plugin.
+* The **node** service implements the various operations for volumes that are run locally from the node, for example `NodePublishVolume`, `NodeStageVolume`, etc. It does not do operations like `CreateVolume` or `ControllerPublish`. Also, it runs directly on the Azure VM instances. It is depending on the Azure cloud config but does not necessarily need to have the credentials.
+
+The CSI driver has two command line flags, `--run-controller-service` and `--run-node-service` which both default to `true`.
+You can disable the individual services by setting the respective flags to `false`.
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -92,25 +92,25 @@ func NewDriver(nodeID string) *Driver {
 }
 
 // Run driver initialization
-func (d *Driver) Run(endpoint, kubeconfig string) {
+func (d *Driver) Run(endpoint, kubeconfig string, runControllerService, runNodeService bool) {
 	versionMeta, err := GetVersionYAML()
 	if err != nil {
 		klog.Fatalf("%v", err)
 	}
+
 	klog.Infof("\nDRIVER INFORMATION:\n-------------------\n%s\n\nStreaming logs below:", versionMeta)
-	cloud, err := GetCloudProvider(kubeconfig)
-	if err != nil || cloud.TenantID == "" || cloud.SubscriptionID == "" {
-		klog.Fatalf("failed to get Azure Cloud Provider, error: %v", err)
-	}
-	d.cloud = cloud
 
-	d.mounter, err = mounter.NewSafeMounter()
-	if err != nil {
-		klog.Fatalf("Failed to get safe mounter. Error: %v", err)
-	}
+	var (
+		s = csicommon.NewNonBlockingGRPCServer()
+		// If enabled, Driver d acts as IdentityServer, ControllerServer and NodeServer
+		cs csi.ControllerServer
+		ns csi.NodeServer
+	)
 
-	d.AddControllerServiceCapabilities(
-		[]csi.ControllerServiceCapability_RPC_Type{
+	if runControllerService {
+		cs = d
+
+		d.AddControllerServiceCapabilities([]csi.ControllerServiceCapability_RPC_Type{
 			csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME,
 			csi.ControllerServiceCapability_RPC_PUBLISH_UNPUBLISH_VOLUME,
 			csi.ControllerServiceCapability_RPC_CREATE_DELETE_SNAPSHOT,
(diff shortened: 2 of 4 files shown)


docs/driver-services.md
```diff
@@ -0,0 +1,12 @@
+### `disk.csi.azure.com` driver services
+ 
+Traditionally, you run the CSI controllers with the Azure Disk driver in the same Kubernetes cluster.
+Though, there may be cases where you will only want to run a subset of the available driver services (for example, one scenario is running the controllers outside of the cluster they are serving (while the Azure Disk driver still runs inside the served cluster), but there might be others scenarios).
```
andyzhangx, line 4: hi @rfranzke thanks for the contribution.
Your scenario is running controller out side of cluster, then what about just modify following script and run `csi-azuredisk-controller` out side of cluster?
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/deploy/csi-azuredisk-controller.yaml
Then you don't need to modify the csi driver code.

rfranzke: Thanks @andyzhangx, sure, if that's possible, even better. What exactly do you suggest to modify?

andyzhangx: so your have two k8s clusters? just run this script in outside cluster:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/deploy/csi-azuredisk-controller.yaml

You may take care about the identity, it's now an `azure.json` file
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/7948c4ed3d43ba906382e5e710d4927a01b56974/deploy/csi-azuredisk-controller.yaml#L151

here is the tips to set that identity file: https://github.com/kubernetes-sigs/azuredisk-csi-driver#prerequisite

rfranzke: Yes, maybe a quick description about the setup: Cluster `A` runs the control plane (KCM, CCM, CSI controllers (provisioner, attacher, etc.)) for the cluster `B`. In cluster `B` only the node part of the CSI shall be run (driver w/ node service, driver-registrar, liveness-probe). Hence, I will basically end up with two instances of the Azure Disk CSI plugin running responsible for cluster `B`: The controller service shall run outside in cluster `A`, and the node service shall run inside in cluster `B`.

I tried what you suggested, but then the CSI driver in `A` also starts the unused node service, and the CSI driver in `B` also starts the unused controller service (here, it's probably even more dangerous, because now two controller services are running and might conflict with each other (?)).

WDYT?

andyzhangx: @rfranzke actually that's the design goal of CSI driver, one csi driver plugin could be used in both controller and node service, it's easy for deployment. There won't be conflict with that since in controller deployment, there should be `csi-provisioner` sidecar container, and in node deployment, there should be `csi-node-driver-registrar` sidecar container.

rfranzke: That's all fine and I don't want to change anything in those regards. The PR is only about allowing better configuration options to help cleanly separate the services for scenarios like the one described. The default behaviour is still like it is today, nobody has to use these flags / the are completely optional.

Anyways, maybe I misconfigured something. Let me try what you suggested once again, I'll come back soon with the results.

Maybe some more background on why I filed this PR: When working with AWS EBS plugin and GCE PD plugin it simply wasn't possible to follow the approach that you described. Both CSI driver implementations were making assumptions that they only run inside the cluster they are serving. In their controller services they assumed that they can talk to their cloud's metadata services. This does not work well with above described scenarios, because cluster `A` might not have access to it, and it even run in different cloud accounts/projects/subscriptions. Similarly, as you couldn't disable the individual driver services, the CSI instances running inside the cluster `B` tried to start the controller service (without having a cloud config) and crashed. Only the node service is needed in `B`. https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/438/ and https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/449. Maybe it works out-of-the-box with Azure, but I still think it might make sense to optionally allow disabling the services if they are needed/used anyways.

rfranzke: @andyzhangx I tried it again with v0.6.0 and it indeed already works out-of-the-box for Azure Disk CSI. Probably I misconfigured something yesterday. So, we don't need to code changes for now. I still think that allowing this optional separation makes sense and also helps that future changes don't break this scenario. Though, if you don't think it's a valueable improvement right now then I'll close this PR for now.

rfranzke: Hm, just now the CSI driver running inside the cluster got a `nil` pointer exception:

```
I0409 09:34:34.984567       1 utils.go:113] GRPC request: volume_id:"/subscriptions/0abd6c55-1c19-4eab-b070-913dec9f2ab0/resourceGroups/my-resource-group/providers/Microsoft.Compute/disks/pv-1317b820-7092-4da1-a83e-d5312e351d45" volume_path:"/var/lib/kubelet/pods/ef50f85b-8b81-4c5a-880e-be9132baf7dc/volumes/kubernetes.io~csi/pv-1317b820-7092-4da1-a83e-d5312e351d45/mount"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x126ea8f]

goroutine 80 [running]:
sigs.k8s.io/azuredisk-csi-driver/pkg/azuredisk.(*Driver).checkDiskExists(0xc00011d4d0, 0x17f74c0, 0xc0002d9860, 0xc000192000, 0xba, 0x469702, 0x1)
	/root/go/src/sigs.k8s.io/azuredisk-csi-driver/pkg/azuredisk/azuredisk.go:178 +0xcf
sigs.k8s.io/azuredisk-csi-driver/pkg/azuredisk.(*Driver).NodeGetVolumeStats(0xc00011d4d0, 0x17f74c0, 0xc0002d9860, 0xc000336320, 0xc00011d4d0, 0x22e44a2, 0xc00011aa80)
	/root/go/src/sigs.k8s.io/azuredisk-csi-driver/pkg/azuredisk/nodeserver.go:331 +0x25e
sigs.k8s.io/azuredisk-csi-driver/vendor/github.com/container-storage-interface/spec/lib/go/csi._Node_NodeGetVolumeStats_Handler.func1(0x17f74c0, 0xc0002d9860, 0x151ba00, 0xc000336320, 0x1, 0x1, 0x17d, 0x18)
	/root/go/src/sigs.k8s.io/azuredisk-csi-driver/vendor/github.com/container-storage-interface/spec/lib/go/csi/csi.pb.go:5630 +0x86
sigs.k8s.io/azuredisk-csi-driver/pkg/csi-common.logGRPC(0x17f74c0, 0xc0002d9860, 0x151ba00, 0xc000336320, 0xc0002f23c0, 0xc0002f23e0, 0xc00001fb30, 0x48b578, 0x14d8960, 0xc0002d9860)
	/root/go/src/sigs.k8s.io/azuredisk-csi-driver/pkg/csi-common/utils.go:115 +0x1ed
sigs.k8s.io/azuredisk-csi-driver/vendor/github.com/container-storage-interface/spec/lib/go/csi._Node_NodeGetVolumeStats_Handler(0x159c8a0, 0xc00011d4d0, 0x17f74c0, 0xc0002d9860, 0xc0000b6ea0, 0x167d9b8, 0x17f74c0, 0xc0002d9860, 0xc0000c89a0, 0x159)
	/root/go/src/sigs.k8s.io/azuredisk-csi-driver/vendor/github.com/container-storage-interface/spec/lib/go/csi/csi.pb.go:5632 +0x14b
sigs.k8s.io/azuredisk-csi-driver/vendor/google.golang.org/grpc.(*Server).processUnaryRPC(0xc000310000, 0x1813b80, 0xc000310900, 0xc0000fa000, 0xc0001418f0, 0x2365b00, 0x0, 0x0, 0x0)
	/root/go/src/sigs.k8s.io/azuredisk-csi-driver/vendor/google.golang.org/grpc/server.go:1024 +0x4f4
sigs.k8s.io/azuredisk-csi-driver/vendor/google.golang.org/grpc.(*Server).handleStream(0xc000310000, 0x1813b80, 0xc000310900, 0xc0000fa000, 0x0)
	/root/go/src/sigs.k8s.io/azuredisk-csi-driver/vendor/google.golang.org/grpc/server.go:1313 +0xd97
sigs.k8s.io/azuredisk-csi-driver/vendor/google.golang.org/grpc.(*Server).serveStreams.func1.1(0xc0000a2900, 0xc000310000, 0x1813b80, 0xc000310900, 0xc0000fa000)
	/root/go/src/sigs.k8s.io/azuredisk-csi-driver/vendor/google.golang.org/grpc/server.go:722 +0xbb
created by sigs.k8s.io/azuredisk-csi-driver/vendor/google.golang.org/grpc.(*Server).serveStreams.func1
	/root/go/src/sigs.k8s.io/azuredisk-csi-driver/vendor/google.golang.org/grpc/server.go:720 +0xa1
```

Do you think it's related to my setup?

andyzhangx: thanks for the test. About the cloud's metadata services, in controller, we already disabled it: https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/7948c4ed3d43ba906382e5e710d4927a01b56974/pkg/azuredisk/controllerserver.go#L405-L407, so that's reason why Azure CSI driver works?

andyzhangx: I am ok to close this PR now since it already works with your testing, thanks.

andyzhangx: @rfranzke the error could be due to disksClient is nil:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/22f446b2a93dcbb383b11e088bf015ed0c6b37d9/pkg/azuredisk/azuredisk.go#L178-L179
How did you set the credentials? read-from-secret only works on master branch now.

PR #368, opened 2020-04-22
Remove deprecated csi-provisioner args

ialidzhikov:
**What type of PR is this?**
/kind cleanup

`--provisioner` and `--connection-timeout` flags are deprecated and have no effect since v1.1.0 of csi-provisioner. Ref https://github.com/kubernetes-csi/external-provisioner/blob/v1.6.0/CHANGELOG-1.1.md

Fixes #

**Release note**:
```
NONE
```


ialidzhikov:
> Thanks for your PR. Could you also port the change to helm charts? They are under charts directory.

Nice catch.

PR #379, opened 2020-04-26
Fix golangci-lint findings

### LINKED ISSUE #372: fix golint issues

Also could enable `golangci-lint run --no-config --enable=golint --disable=typecheck --deadline=10m` for `hack/verify-golint.sh` script
```
# golangci-lint run --no-config --enable=golint --disable=typecheck --deadline=10m
test/utils/credentials/credentials.go:67:6: type name will be used as credentials.CredentialsConfig by other packages, and that stutters; consider calling this Config (golint)
type CredentialsConfig struct {
     ^
test/utils/credentials/credentials.go:73:6: type name will be used as credentials.CredentialsFromProw by other packages, and that stutters; consider calling this FromProw (golint)
type CredentialsFromProw struct {
     ^
test/utils/azure/azure_helpers.go:35:6: type name will be used as azure.AzureClient by other packages, and that stutters; consider calling this Client (golint)
type AzureClient struct {
     ^
pkg/azuredisk/azure_common_linux.go:75:2: don't use underscores in Go names; var `scsi_path` should be `scsiPath` (golint)
        scsi_path := "/sys/class/scsi_host/"
        ^
pkg/azuredisk/azure_common_linux.go:101:2: don't use underscores in Go names; var `sys_path` should be `sysPath` (golint)
        sys_path := "/sys/bus/scsi/devices"
        ^
pkg/azuredisk/azuredisk.go:253:6: func getSourceVolumeId should be getSourceVolumeID (golint)
func getSourceVolumeId(snapshot *compute.Snapshot) string {
     ^
pkg/azuredisk/azuredisk.go:290:10: `if` block ends with a `return` statement, so drop this `else` and outdent its block (golint)
                } else {
                       ^
pkg/azuredisk/azuredisk_test.go:391:2: var `sourceResourceSnapshotId` should be `sourceResourceSnapshotID` (golint)
        sourceResourceSnapshotId := "/subscriptions/xxx/resourceGroups/xxx/providers/Microsoft.Compute/snapshots/xxx"
        ^
pkg/azuredisk/azuredisk_test.go:392:2: var `sourceResourceVolumeId` should be `sourceResourceVolumeID` (golint)
        sourceResourceVolumeId := "/subscriptions/xxx/resourceGroups/xxx/providers/Microsoft.Compute/disks/xxx"
        ^
pkg/azuredisk/controllerserver.go:610:10: `if` block ends with a `return` statement, so drop this `else` and outdent its block (golint)
                } else {
                       ^
pkg/azuredisk/nodeserver.go:52:9: `if` block ends with a `return` statement, so drop this `else` and outdent its block (golint)
        } else {
               ^
```

**Additional context**

ialidzhikov:
**What type of PR is this?**
/kind cleanup

Fixes #372

**Release note**:
```
NONE
```

### DIFF
--- a/hack/verify-golint.sh
+++ b/hack/verify-golint.sh
@@ -21,6 +21,8 @@ if [[ -z "$(command -v golangci-lint)" ]]; then
   curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.24.0
 fi
 
-golangci-lint run --deadline=10m
+readonly PKG_ROOT="$(git rev-parse --show-toplevel)"
 
-echo "Congratulations! Lint check completed for all Go source files."
+golangci-lint run --config ${PKG_ROOT}/.golangci.yml
+
+echo "Success! Lint check completed for all Go source files."
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,8 @@
+run:
+  deadline: 10m
+
+linters:
+  disable:
+  - typecheck
+  enable:
+  - golint
--- a/pkg/azuredisk/azure_common_linux.go
+++ b/pkg/azuredisk/azure_common_linux.go
@@ -72,17 +72,17 @@ func getDiskLinkByDevName(io ioHandler, devLinkPath, devName string) (string, er
 }
 
 func scsiHostRescan(io ioHandler, m *mount.SafeFormatAndMount) {
-	scsi_path := "/sys/class/scsi_host/"
-	if dirs, err := io.ReadDir(scsi_path); err == nil {
+	scsiPath := "/sys/class/scsi_host/"
+	if dirs, err := io.ReadDir(scsiPath); err == nil {
 		for _, f := range dirs {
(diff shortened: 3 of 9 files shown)


hack/verify-golint.sh
```diff
@@ -21,6 +21,8 @@ if [[ -z "$(command -v golangci-lint)" ]]; then
   curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.24.0
 fi
 
-golangci-lint run --deadline=10m
+readonly PKG_ROOT="$(git rev-parse --show-toplevel)"
 
-echo "Congratulations! Lint check completed for all Go source files."
+golangci-lint run --config ${PKG_ROOT}/.golangci.yml
```
andyzhangx, line 26: why not use this standalone command:
```
golangci-lint run --no-config --enable=golint --disable=typecheck --deadline=10m
```

ialidzhikov: I think that the `.golangci.yaml` is more convenient and readable as in future we may want to enable/disable other linters and the command line args would grow. WDYT?

PR #468, opened 2020-08-01
[WIP]fix: DNS latency in controller

### LINKED ISSUE #441: Create Disk/snapshot/Resize Disk costs 10s while it should only costs 2s
**What happened**:
Actually we already respects Retry-After in [WaitForCompletionRef](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/b3b48d380066d8adc094ae6aa79c84aa6e19e437/vendor/github.com/Azure/go-autorest/autorest/azure/async.go#L171-L201), but not sure why it’s still taking so much time(10s vs 2s):
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/b3b48d380066d8adc094ae6aa79c84aa6e19e437/vendor/github.com/Azure/go-autorest/autorest/azure/async.go#L171-L201

The way  [WaitForCompletionRef](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/b3b48d380066d8adc094ae6aa79c84aa6e19e437/vendor/github.com/Azure/go-autorest/autorest/azure/async.go#L171-L201)  is implemented, it seems like there will be one polling call before the Retry-After field is considered. Is it possible that the initial polling error is due to this behavior where client is not respecting the Retry-After hint?

also related to bug: https://github.com/Azure/go-autorest/issues/538

**How to reproduce it**:

**Environment**:
- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

andyzhangx:
**What type of PR is this?**
/kind bug

fix: DNS latency in controller

Create Disk/snapshot/Resize Disk costs 10s now, with this PR, it could complete within ~5s(sometimes 2s)

Fixes #441

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

##dnsConfig options

We can update the dnsConfig in pod’s spec to avoid the race condition. For example, we can enable the option single-request-reopen:
```
dnsConfig:  
  options:  
    - name: single-request-reopen
```
The description about the option is:
```
single-request-reopen (since glibc 2.9)
    Sets RES_SNGLKUPREOP in _res.options.  The resolver
    uses the same socket for the A and AAAA requests.  Some
    hardware mistakenly sends back only one reply.  When
    that happens the client system will sit and wait for
    the second reply.  Turning this option on changes this
    behavior so that if two requests from the same port are
    not handled correctly it will close the socket and open
    a new one before sending the second request.
```
Similarly, we can enable the option use-vc to enforce TCP when resolving DNS:
```
use-vc (since glibc 2.14)
    Sets RES_USEVC in _res.options.  This option forces the
    use of TCP for DNS resolutions.
```
Notice that these two options have dependencies on glibc, but glibc is unavailable in some images, such as alpine. We have to use other images if we’d like to go with dnsConfig options.

If we put everything together, the pod spec may look like:
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: app
spec:
  template:
    spec:
      dnsConfig:
        options:
         - name: single-request-reopen
         - name: ndots
           value: "5"
         - name: use-vc 
      containers:
```

**Release note**:
```
fix: DNS latency in controller
```

### DIFF
--- a/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
+++ b/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
@@ -24,6 +24,12 @@ spec:
           operator: "Equal"
           value: "true"
           effect: "NoSchedule"
+      dnsConfig:
+        options:
+          - name: single-request-reopen
+          - name: ndots
+            value: "5"
+          - name: use-vc
       containers:
         - name: csi-provisioner
           image: "{{ .Values.image.csiProvisioner.repository }}:{{ .Values.image.csiProvisioner.tag }}"
--- a/deploy/csi-azuredisk-controller.yaml
+++ b/deploy/csi-azuredisk-controller.yaml
@@ -24,6 +24,12 @@ spec:
           operator: "Equal"
           value: "true"
           effect: "NoSchedule"
+      dnsConfig:
+        options:
+          - name: single-request-reopen
+          - name: ndots
+            value: "5"
+          - name: use-vc
       containers:
         - name: csi-provisioner
           image: mcr.microsoft.com/oss/kubernetes-csi/csi-provisioner:v1.5.0


andyzhangx:
not related to dns issue

PR #476, opened 2020-08-05
ci: add options to skip installing / uninstalling azuredisk-csi-driver during e2e test

chewong:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature
> /kind flake

Currently working on incorporating e2e tests in an internal pipeline and it'd be nice to have an option to run e2e tests on a cluster that already has azuredisk-csi-driver install.

/assign @andyzhangx  

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


Makefile
```diff
@@ -70,6 +70,7 @@ e2e-test:
 
 .PHONY: e2e-bootstrap
 e2e-bootstrap: install-helm
+ifndef E2E_SKIP_INSTALL
```
andyzhangx, line 73: so in this PR, when run `make e2e-bootstrap`, it only installs helm, why not user don't run `make e2e-bootstrap`

andyzhangx: maybe we don't need this `E2E_SKIP_INSTALL` flag, and only run `make e2e-test` could solve the issue?

chewong: It's part of the e2e test suite - https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/test/e2e/suite_test.go#L88-L93. We also shouldn't install helm. I can modify the PR.

chewong:
/hold
I am exploring different options of running e2e test internally.

PR #488, opened 2020-08-18
test: add disk statefulset e2e-test

### LINKED ISSUE #238: add statefulset e2e test cases
**Is your feature request related to a problem? Please describe.**

Add following e2e cases which is more similar to real e2e scenarios:
 - run a statefulset which has one pod with disk mount
 - delete the only pod and new pod will be scheduled on the same node
 - wait for pod status as `Running` 
 - if it's not in `Running` state for about 10min, then test fail, otherwise succeed

Config examples:
 - deployment example:
https://github.com/andyzhangx/demo/blob/master/linux/azuredisk/attach-stress-test/deployment/deployment-azuredisk1.yaml

 - statefulset example:
https://github.com/andyzhangx/demo/blob/master/linux/azuredisk/attach-stress-test/statefulset/statefulset-azuredisk1.yaml

**Additional context**

Sakuralbj:
**What type of PR is this?**
/kind test

Fixes #238 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/test/e2e/testsuites/specs.go
+++ b/test/e2e/testsuites/specs.go
@@ -162,6 +162,28 @@ func (pod *PodDetails) SetupDeployment(client clientset.Interface, namespace *v1
 	return tDeployment, cleanupFuncs
 }
 
+func (pod *PodDetails) SetupStatefulset(client clientset.Interface, namespace *v1.Namespace, csiDriver driver.DynamicPVTestDriver) (*TestStatefulset, []func()) {
+	cleanupFuncs := make([]func(), 0)
+	volume := pod.Volumes[0]
+	ginkgo.By("setting up the StorageClass")
+	storageClass := csiDriver.GetDynamicProvisionStorageClass(driver.GetParameters(), volume.MountOptions, volume.ReclaimPolicy, volume.VolumeBindingMode, volume.AllowedTopologyValues, namespace.Name)
+	tsc := NewTestStorageClass(client, namespace, storageClass)
+	createdStorageClass := tsc.Create()
+	cleanupFuncs = append(cleanupFuncs, tsc.Cleanup)
+	ginkgo.By("setting up the PVC")
+	tpvc := NewTestPersistentVolumeClaim(client, namespace, volume.ClaimSize, volume.VolumeMode, &createdStorageClass)
+	storageClassName := ""
+	if tpvc.storageClass != nil {
+		storageClassName = tpvc.storageClass.Name
+	}
+	tpvc.requestedPersistentVolumeClaim = generateStatefulSetPVC(tpvc.namespace.Name, storageClassName, tpvc.claimSize, tpvc.volumeMode, tpvc.dataSource)
+	ginkgo.By("setting up the statefulset")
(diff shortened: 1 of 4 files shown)


test/e2e/testsuites/specs.go
```diff
@@ -162,6 +162,29 @@ func (pod *PodDetails) SetupDeployment(client clientset.Interface, namespace *v1
...
+	createdStorageClass := tsc.Create()
+	cleanupFuncs = append(cleanupFuncs, tsc.Cleanup)
+	ginkgo.By("setting up the PVC")
+	tpvc := NewTestPersistentVolumeClaim(client, namespace, volume.ClaimSize, volume.VolumeMode, &createdStorageClass)
+	storageClassName := ""
+	if tpvc.storageClass != nil {
+		storageClassName = tpvc.storageClass.Name
+	}
+	tpvc.requestedPersistentVolumeClaim = generateStatefulSetPVC(tpvc.namespace.Name, storageClassName, tpvc.claimSize, tpvc.volumeMode, tpvc.dataSource)
+	//cleanupFuncs = append(cleanupFuncs, tpvc.Cleanup)
```
andyzhangx, line 180: remove this?

test/e2e/testsuites/testsuites.go
```diff
@@ -504,6 +528,149 @@ func (t *TestDeployment) Logs() ([]byte, error) {
...
+func (t *TestStatefulset) DeletePodAndWait() {
+	e2elog.Logf("Deleting pod %q in namespace %q", t.podName, t.namespace.Name)
+	err := t.client.CoreV1().Pods(t.namespace.Name).Delete(context.TODO(), t.podName, metav1.DeleteOptions{})
+	if err != nil {
+		if !apierrs.IsNotFound(err) {
+			framework.ExpectNoError(fmt.Errorf("pod %q Delete API error: %v", t.podName, err))
+		}
+		return
+	}
+	time.Sleep(40 * time.Second)
```
andyzhangx, line 639: why sleep here?

test/e2e/testsuites/specs.go
Sakuralbj: Ok.I remove it.

test/e2e/testsuites/testsuites.go
Sakuralbj: After pod is deleted, waitForPodready will  ensure the new pod is ready. If don't sleep here,waitForPodready will be passed before old pod is deleted.

andyzhangx: ok, pls add a comment before `time.Sleep`.

andyzhangx: so statefulset is different from deployment, in deployment, it's not necessary to wait

Sakuralbj: Actually,in deployment also use wait,but it use funcition like this.
```
err = e2epod.WaitForPodNoLongerRunningInNamespace(t.client, t.podName, t.namespace.Name)
```
But in statefulset situation, If I use this function,the deleted pod will never be NoLongerRunning State as the new pod has the same name and namespace with old one. So after the new pod created,the test will timeout here. And I choose to use sleep to wait.

andyzhangx (approved):
/lgtm
/approve

PR #603, opened 2020-11-25
fix: use --mountpoint instead of --target for findmnt

songjiaxun:
**What type of PR is this?**
/kind bug

As this [ticket](https://bugzilla.redhat.com/show_bug.cgi?id=1405238) described, for some old version `findmnt`, if the volume path does not exist, using flag `--target` will not give you any error (exit code is 0), which is not what we want.

In Azure Stack test, `findmnt` with the flag `--target` always gives exit code 0 no matter what path it is given. However, the flag `--mountpoint` works fine.

As the `findmnt` [manual](https://man7.org/linux/man-pages/man8/findmnt.8.html) says, "It's recommended to use the option --mountpoint when checks of path elements are unwanted and path is a strictly specified mountpoint."

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/nodeserver.go
+++ b/pkg/azuredisk/nodeserver.go
@@ -380,7 +380,7 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
 	volSizeBytes := int64(capacityBytes)
 	requestGiB := volumehelper.RoundUpGiB(volSizeBytes)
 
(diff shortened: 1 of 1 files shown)


andyzhangx (approved):
/lgtm
/approve

andyzhangx:
could you also port this change to azure file driver? thanks.

PR #606, opened 2020-11-26
fix: add Azure Stack specific logic

songjiaxun:
**What type of PR is this?**
/kind bug
/kind test

Add function `IsAzureStackCloud` to decide if the driver is running on Azure Stack Cloud.
If it is on Azure Stack Cloud, then the shared disk is not supported, only Standard_LRS and Premium_LRS are supported, the incremental snapshot is not supported. Corresponding documentation will be in the next PR.
Apply the same logic on the e2e tests.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -161,6 +161,12 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		}
 	}
 
+	if IsAzureStackCloud(d.cloud) {
+		if maxShares > 1 {
+			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Invalid maxShares value: %d as Azure Stack does not support shared disk.", maxShares))
+		}
+	}
+
 	if maxShares < 2 {
 		for _, c := range volCaps {
 			mode := c.GetAccessMode().Mode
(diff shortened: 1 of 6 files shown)


test/e2e/driver/azuredisk_driver.go
```diff
@@ -123,6 +123,6 @@ func (d *azureDiskDriver) GetPersistentVolume(volumeID string, fsType string, si
 
 func GetParameters() map[string]string {
 	return map[string]string{
-		"skuName": "StandardSSD_LRS",
+		"skuName": "Standard_LRS",
```
andyzhangx, line 126: we still want to test other `skuName` for public cloud. Should add a check here whether it's AzureStack.

test/e2e/driver/azuredisk_driver.go
```diff
@@ -77,6 +77,12 @@ func (d *azureDiskDriver) GetDynamicProvisionStorageClass(parameters map[string]
 		}
 	}
 
+	if strings.EqualFold(os.Getenv("AZURE_CLOUD_NAME"), "AZURESTACKCLOUD") {
```
andyzhangx, line 80: is it still necessary since in `controllerserver.go`, you already default the skuName to `Standard_LRS`

songjiaxun: yes, it is still necessary because in the migration test the in-tree provisioner does not have similar logic. Without this change, unacceptable sku still got to be sent to in-tree provisioner.

pkg/azuredisk/controllerserver.go
```diff
@@ -186,6 +192,14 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, err
 	}
 
+	if IsAzureStackCloud(d.cloud) {
+		if skuName != compute.StandardLRS && skuName != compute.PremiumLRS {
+			klog.V(2).Infof("Use sku %s instead as Aure Stack does not support %s.", compute.StandardLRS, skuName)
```
andyzhangx, line 197: Azure

andyzhangx (approved):
/lgtm
/approve
/retest

Issue #591, opened 2020-11-18
Raw block + volume expansion does not work


Jiawei0227:
**What happened**:

I am running kubernetes e2e tests specifically the volume expansion test https://github.com/kubernetes/kubernetes/blob/160c33a6a17b6c49fee365d7cbe2acc7f1fe2f8e/test/e2e/storage/testsuites/volume_expand.go#L169

And the combination of raw block and offline volume expansion does not work. The volume mount fail with following error in the pod events:
```
MapVolume.MarkVolumeAsMounted failed while expanding volume for volume "pvc-b060b08d-71e1-48cf-97f4-590ed86e7363" : MountVolume.NodeExpandVolume failed : Expander.NodeExpand failed to expand the volume : rpc error: code = Internal desc = Could not get size of block volume at path devtmpfs[/sdc]: error when getting size of block volume at path devtmpfs[/sdc]: output: , err: exit status 1
```

**What you expected to happen**:
Expansion succeed.

**How to reproduce it**:
Create a raw block PVC and do volume expansion on it.

**Anything else we need to know?**:

**Environment**:
- CSI Driver version: 0.9.0
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

andyzhangx:
thanks, do you know is there any special handling to support `Raw block + volume expansion`,  actually filesystem+volume expansion already works now.

Jiawei0227:
I am not really sure... I think there might something need to be done in the NodePublishVolume where you check if File System Expansion is required? Again I dont have much knowledge on this. 
/cc @msau42 to help understand this better.

msau42:
For raw block PV, usually the block device is resized in the controller call, and there's nothing to do in the Node call because there's no filesystem. For example, in PD driver, this is a noop:
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/f4608fbaa86182ffde2df3279bf61d10067ab454/pkg/gce-pd-csi-driver/node.go#L467

Jiawei0227:
Keep exploring at this, I found the code in PDCSI is not being executed: 
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/f4608fbaa86182ffde2df3279bf61d10067ab454/pkg/gce-pd-csi-driver/node.go#L467

[...]

PR #607, opened 2020-12-01
fix: skip raw block node volume expansion

Jiawei0227:
**What type of PR is this?**
/kind bug

This PR tries to skip node expansion when volume type is block. However, this does not guarantee to work well because volumeCapability is not yet being passed from kubelet until v1.19.

https://github.com/kubernetes/kubernetes/commit/7d6959ce2ce49fa5b37b3aa28d20fdeb01c02223

The root cause for this is the way it get devicePath. I am not familiar with this domain so I dont know how to get the devicepath given volumepath when we cant use the mount information.

[Update] I am using  `isBlock, err := hostutil.NewHostUtil().PathIsDevice(volumePath)` to check if this is a block device and it seems to work.

Testing:
Running e2e testsuite offline volume expansion and it worked for block volume.

Fix https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/591

**Release note**:
```
none
```

/cc @msau42 @jingxu97 
/assign @andyzhangx

### DIFF
--- a/pkg/azuredisk/nodeserver.go
+++ b/pkg/azuredisk/nodeserver.go
@@ -380,19 +380,55 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
 	volSizeBytes := int64(capacityBytes)
 	requestGiB := volumehelper.RoundUpGiB(volSizeBytes)
 
-	args := []string{"-o", "source", "--noheadings", "--mountpoint", req.GetVolumePath()}
-	output, err := d.mounter.Exec.Command("findmnt", args...).Output()
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Could not determine device path: %v", err)
+	volumePath := req.GetVolumePath()
+	if len(volumePath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
 	}
 
-	devicePath := strings.TrimSpace(string(output))
-	if len(devicePath) == 0 {
-		return nil, status.Errorf(codes.Internal, "Could not get valid device for mount path: %q", req.GetVolumePath())
+	isBlock, err := hostutil.NewHostUtil().PathIsDevice(volumePath)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to determine device path for volumePath [%v]: %v", volumePath, err)
+	}
+	if isBlock {
+		// Noop for Block NodeExpandVolume
+		klog.V(4).Infof("NodeExpandVolume succeeded on %v to %s, path check is block so this is a no-op", volumeID, volumePath)
+		return &csi.NodeExpandVolumeResponse{}, nil
+	}
+
+	// Hopefully devicePath was passed from req
+	// In kubernetes 1.19 above, it will be passed by kubelet.
+	devicePath := req.GetStagingTargetPath()
+	if devicePath == "" {
+		args := []string{"-o", "source", "--noheadings", "--mountpoint", volumePath}
+		output, err := d.mounter.Exec.Command("findmnt", args...).Output()
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "Could not determine device path: %v", err)
+		}
+
+		devicePath = strings.TrimSpace(string(output))
+		if len(devicePath) == 0 {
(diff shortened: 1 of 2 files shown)


pkg/azuredisk/nodeserver.go
```diff
@@ -380,19 +380,43 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
...
+	if len(volumePath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
 	}
 
-	devicePath := strings.TrimSpace(string(output))
-	if len(devicePath) == 0 {
-		return nil, status.Errorf(codes.Internal, "Could not get valid device for mount path: %q", req.GetVolumePath())
+	// Hopefully devicePath was passed from req
+	devicePath := req.GetStagingTargetPath()
+	if devicePath == "" {
```
Jiawei0227, line 390: Here it should use some other logic to get the device path for volumes that are raw block for v1.19 before.

Jiawei0227: Update: I am using isBlock to check if this is Block. After testing it seems to work

pkg/azuredisk/nodeserver.go
```diff
@@ -380,19 +380,54 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
...
 	requestGiB := volumehelper.RoundUpGiB(volSizeBytes)
 
-	args := []string{"-o", "source", "--noheadings", "--mountpoint", req.GetVolumePath()}
-	output, err := d.mounter.Exec.Command("findmnt", args...).Output()
+	volumePath := req.GetVolumePath()
+	if len(volumePath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
+	}
+
+	// Hopefully devicePath was passed from req
```
msau42, line 388: Add a note that this Kubernetes only started passing staging path in 1.19

pkg/azuredisk/nodeserver.go
```diff
@@ -380,19 +380,54 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
...
-		return nil, status.Errorf(codes.Internal, "Could not get valid device for mount path: %q", req.GetVolumePath())
+	volumeCapability := req.GetVolumeCapability()
+	if volumeCapability != nil {
+		// VolumeCapability is optional, if specified, validate it
+		caps := []*csi.VolumeCapability{volumeCapability}
+		if !isValidVolumeCapabilities(caps) {
+			return nil, status.Error(codes.InvalidArgument, "VolumeCapability is invalid.")
+		}
+
+		if blk := volumeCapability.GetBlock(); blk != nil {
```
msau42, line 421: Do we need both "PathIsDevice" and volume capability checks?

pkg/azuredisk/nodeserver.go
```diff
@@ -380,19 +380,54 @@ func (d *Driver) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolume
...
+	volumePath := req.GetVolumePath()
+	if len(volumePath) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "volume path must be provided")
+	}
+
+	// Hopefully devicePath was passed from req
+	devicePath := req.GetStagingTargetPath()
+	if devicePath == "" {
+		args := []string{"-o", "source", "--noheadings", "--mountpoint", volumePath}
+		output, err := d.mounter.Exec.Command("findmnt", args...).Output()
```
msau42, line 392: If the PV.volumemode is block, does this "findmnt" command work?

pkg/azuredisk/nodeserver.go
Jiawei0227: Yeah I am also wondering about this. That's why I had this log below. So if a path is a not a block but the capability says it is, then we should be alerted here. I can remove it if you feel like this is unnecessary

pkg/azuredisk/nodeserver.go
Jiawei0227: Yes, it will not fail. But the output will not make sense. I think most of the time it just return "" empty. Maybe I should move the isBlock part up

andyzhangx (approved):
/lgtm
/approve
thanks for the fix, really appreciate that!

Issue #617, opened 2020-12-16
Add Unit Tests for ListVolumes API


nearora-msft:
**Is your feature request related to a problem?/Why is this needed**
Implement Unit Tests for ListVolumes API

**Describe the solution you'd like in detail**
The TestListVolumes UT in the controllerserver_test.go file is skipped for now as it is dependent on some changes in the legacy-cloud-provider.(https://github.com/kubernetes/kubernetes/issues/97341)   Make the code changes for the UT once the dependency are available.

**Describe alternatives you've considered**

**Additional context**
Related to issue : https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/378

PR #618, opened 2020-12-16
feat: Implement ListVolumes

### LINKED ISSUE #378: Implement ListVolumes

When a Node gets deleted, and Pod is rescheduled, Kubernetes thinks that the the volume is still attached to the original node and fails to attach to the new node with Multi-Attach error.

The latest csi-attacher v2.1 adds support for the ListVolumes capability: https://github.com/kubernetes-csi/external-attacher/blob/v2.1.1/CHANGELOG-2.1.md

https://github.com/container-storage-interface/spec/blob/v1.3.0/spec.md#listvolumes
https://docs.microsoft.com/en-us/rest/api/compute/disks/listbyresourcegroup
https://github.com/kubernetes/kubernetes/blame/master/staging/src/k8s.io/legacy-cloud-providers/azure/clients/snapshotclient/azure_snapshotclient.go

It addresses problems when volumes may get detached out of band, and need to be reattached to a node. It could potentially help the other way around too, but that hasn't been tested.

depends on PR(https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/377)

**Additional context**

nearora-msft:
**What type of PR is this?**
/kind feature

Fixes #378 #617 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
feat: Implement ListVolumes
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -538,7 +538,68 @@ func (d *Driver) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (
 
 // ListVolumes return all available volumes
 func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (*csi.ListVolumesResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "")
+	var err error
+	start := 0
+	if req.StartingToken != "" {
+		start, err = strconv.Atoi(req.StartingToken)
+		if err != nil {
+			return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%s) parsing with error: %v", req.StartingToken, err)
+		}
+		if start < 0 {
+			return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%d) can not be negative", start)
+		}
+	}
+
+	disks, derr := d.cloud.DisksClient.ListByResourceGroup(ctx, d.cloud.ResourceGroup)
+	if derr != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("ListVolumes on rg(%s) failed with error: %v", d.cloud.ResourceGroup, derr.Error()))
+	}
+
+	if start != 0 && start >= len(disks) {
+		return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%d) on rg(%s) is greater than total number of volumes", start, d.cloud.ResourceGroup)
+	}
+
+	maxEntries := len(disks) - start
+	if req.MaxEntries > 0 && int(req.MaxEntries) < maxEntries {
+		maxEntries = int(req.MaxEntries)
+	}
+	pageEnd := start + maxEntries
+
+	vmSet := d.cloud.VMSet
+	entries := []*csi.ListVolumesResponse_Entry{}
+	for i := start; i < pageEnd; i++ {
+		d := disks[i]
+		nodeList := []string{}
+
+		if d.ManagedBy != nil {
+			attachedNode, err := vmSet.GetNodeNameByProviderID(*d.ManagedBy)
+			if err != nil {
+				return nil, err
+			}
+			nodeList = append(nodeList, string(attachedNode))
+		}
+
+		entries = append(entries, &csi.ListVolumesResponse_Entry{
+			Volume: &csi.Volume{
+				VolumeId: *d.ID,
+			},
+			Status: &csi.ListVolumesResponse_VolumeStatus{
+				PublishedNodeIds: nodeList,
+			},
+		})
+	}
+
+	nextTokenString := ""
+	if pageEnd < len(disks) {
+		nextTokenString = strconv.Itoa(pageEnd)
+	}
+
+	listVolumesResp := &csi.ListVolumesResponse{
+		Entries:   entries,
+		NextToken: nextTokenString,
+	}
+
+	return listVolumesResp, nil
 }
 
 // ControllerExpandVolume controller expand volume
--- a/pkg/azuredisk/controllerserver_test.go
+++ b/pkg/azuredisk/controllerserver_test.go
@@ -1481,12 +1481,133 @@ func TestGetCapacity(t *testing.T) {
 }
 
 func TestListVolumes(t *testing.T) {
-	d, _ := NewFakeDriver(t)
-	req := csi.ListVolumesRequest{}
-	resp, err := d.ListVolumes(context.Background(), &req)
-	assert.Nil(t, resp)
-	if !reflect.DeepEqual(err, status.Error(codes.Unimplemented, "")) {
-		t.Errorf("Unexpected error: %v", err)
+	testCases := []struct {
+		name     string
+		testFunc func(t *testing.T)
+	}{
+		{
+			name: "Valid list without max_entries or starting_token",
+			testFunc: func(t *testing.T) {
+				req := csi.ListVolumesRequest{}
+				d, _ := NewFakeDriver(t)
+				fakeVolumeID := "test"
+				disk := compute.Disk{ID: &fakeVolumeID}
+				disks := []compute.Disk{}
+				disks = append(disks, disk)
+				ctrl := gomock.NewController(t)
+				defer ctrl.Finish()
+				d.cloud.DisksClient.(*mockdiskclient.MockInterface).EXPECT().ListByResourceGroup(gomock.Any(), gomock.Any()).Return(disks, nil).AnyTimes()
+				expectedErr := error(nil)
+				listVolumesResponse, err := d.ListVolumes(context.TODO(), &req)
+				if !reflect.DeepEqual(err, expectedErr) {
+					t.Errorf("actualErr: (%v), expectedErr: (%v)", err, expectedErr)
+				}
+				if listVolumesResponse.NextToken != "" {
+					t.Errorf("actualNextToken: (%v), expectedNextToken: (%v)", listVolumesResponse.NextToken, "")
+				}
+			},
+		},
+		{
+			name: "Valid list with max_entries",
+			testFunc: func(t *testing.T) {
+				req := csi.ListVolumesRequest{
+					MaxEntries: 1,
+				}
+				d, _ := NewFakeDriver(t)
+				fakeVolumeID := "test"
+				disk1, disk2 := compute.Disk{ID: &fakeVolumeID}, compute.Disk{ID: &fakeVolumeID}
+				disks := []compute.Disk{}
+				disks = append(disks, disk1, disk2)
+				ctrl := gomock.NewController(t)
+				defer ctrl.Finish()
+				d.cloud.DisksClient.(*mockdiskclient.MockInterface).EXPECT().ListByResourceGroup(gomock.Any(), gomock.Any()).Return(disks, nil).AnyTimes()
+				expectedErr := error(nil)
+				listVolumesResponse, err := d.ListVolumes(context.TODO(), &req)
+				if !reflect.DeepEqual(err, expectedErr) {
+					t.Errorf("actualErr: (%v), expectedErr: (%v)", err, expectedErr)
+				}
+				if len(listVolumesResponse.Entries) != int(req.MaxEntries) {
+					t.Errorf("Actual number of entries: (%v), Expected number of entries: (%v)", len(listVolumesResponse.Entries), req.MaxEntries)
+				}
+				if listVolumesResponse.NextToken != "1" {
+					t.Errorf("actualNextToken: (%v), expectedNextToken: (%v)", listVolumesResponse.NextToken, "1")
+				}
+			},
+		},
+		{
+			name: "Valid list with max_entries and starting_token",
+			testFunc: func(t *testing.T) {
+				req := csi.ListVolumesRequest{
+					StartingToken: "1",
+					MaxEntries:    1,
+				}
+				d, _ := NewFakeDriver(t)
+				fakeVolumeID1, fakeVolumeID12 := "test1", "test2"
+				disk1, disk2 := compute.Disk{ID: &fakeVolumeID1}, compute.Disk{ID: &fakeVolumeID12}
+				disks := []compute.Disk{}
+				disks = append(disks, disk1, disk2)
+				ctrl := gomock.NewController(t)
+				defer ctrl.Finish()
+				d.cloud.DisksClient.(*mockdiskclient.MockInterface).EXPECT().ListByResourceGroup(gomock.Any(), gomock.Any()).Return(disks, nil).AnyTimes()
+				expectedErr := error(nil)
+				listVolumesResponse, err := d.ListVolumes(context.TODO(), &req)
+				if !reflect.DeepEqual(err, expectedErr) {
+					t.Errorf("actualErr: (%v), expectedErr: (%v)", err, expectedErr)
+				}
+				if len(listVolumesResponse.Entries) != int(req.MaxEntries) {
+					t.Errorf("Actual number of entries: (%v), Expected number of entries: (%v)", len(listVolumesResponse.Entries), req.MaxEntries)
+				}
+				if listVolumesResponse.NextToken != "" {
+					t.Errorf("actualNextToken: (%v), expectedNextToken: (%v)", listVolumesResponse.NextToken, "")
+				}
+				if listVolumesResponse.Entries[0].Volume.VolumeId != fakeVolumeID12 {
+					t.Errorf("actualVolumeId: (%v), expectedVolumeId: (%v)", listVolumesResponse.Entries[0].Volume.VolumeId, fakeVolumeID12)
+				}
+			},
+		},
+		{
+			name: "ListVolumes request with starting token but no entries in response",
+			testFunc: func(t *testing.T) {
+				req := csi.ListVolumesRequest{
+					StartingToken: "1",
+				}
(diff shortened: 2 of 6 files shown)


pkg/azuredisk/controllerserver.go
```diff
@@ -524,7 +524,68 @@ func (d *Driver) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (
 
 // ListVolumes return all available volumes
 func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (*csi.ListVolumesResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "")
+
+	klog.V(2).Infof("ListVolumes called with request %v", *req)
```
andyzhangx, line 528: not necessary to print request here, and also remove empty line

pkg/azuredisk/controllerserver.go
```diff
@@ -524,7 +524,68 @@ func (d *Driver) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (
...
+
+	pageEnd := len(disks)
+	if req.MaxEntries > 0 {
+		pageEnd = start + int(req.MaxEntries)
+	}
+
+	entries := []*csi.ListVolumesResponse_Entry{}
+
+	for i := start; i < pageEnd; i++ {
+
```
andyzhangx, line 560: remove empty line here

andyzhangx (commented):
could you also add `list-volumes` in integration test:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/test/integration/run-test.sh

also is there any way to add e2e tests for this feature? thanks.

```console
# csc controller list-volumes -h
NAME
    list-volumes -- invokes the rpc "ListVolumes"

SYNOPSIS
    csc controller list-volumes [flags]

ALIASES
    list-volumes, ls, list, volumes

OPTIONS
        --format
        The Go template format used to emit the results

    -h, --help
        help for list-volumes

        --max-entries
        The maximum number of entries to return

        --paging
        Enables auto-paging

        --starting-token
        The starting token used to retrieve paged data
```

pkg/azuredisk/controllerserver.go
```diff
@@ -524,7 +524,68 @@ func (d *Driver) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (
...
+		return nil, status.Error(codes.Internal, fmt.Sprintf("Unknown list disks error: %v", derr.Error()))
+	}
+
+	if start >= len(disks) {
+		return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%d) is greater than total number of volumes", start)
+	}
+
+	pageEnd := len(disks)
+	if req.MaxEntries > 0 {
+		pageEnd = start + int(req.MaxEntries)
```
edreed, line 554: This would allow `pageEnd` to be `> len(disk)` which would cause a run-time panic on line 561 below.
```suggestion
		maxEntries := len(disks) - start
		if maxEntries > int(req.MaxEntries) {
			maxEntries = int(req.MaxEntries)
		}
		pageEnd = start + maxEntries
```

pkg/azuredisk/controllerserver_test.go
```diff
@@ -1481,13 +1481,7 @@ func TestGetCapacity(t *testing.T) {
 }
 
 func TestListVolumes(t *testing.T) {
-	d, _ := NewFakeDriver(t)
```
andyzhangx, line 1484: pls enable ut after this PR(https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/627) merged.

andyzhangx (commented):
Could you also add integration tests and fix the sanity test failure?

pkg/azuredisk/controllerserver.go
```diff
@@ -524,7 +524,65 @@ func (d *Driver) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (
...
+		if err != nil {
+			return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%s) parsing with error: %v", req.StartingToken, err)
+		}
+		if start < 0 {
+			return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%d) can not be negative", start)
+		}
+	}
+
+	disks, derr := d.cloud.DisksClient.ListByResourceGroup(ctx, d.cloud.ResourceGroup)
+
```
andyzhangx, line 540: no empty line here

pkg/azuredisk/controllerserver.go
```diff
@@ -524,7 +524,65 @@ func (d *Driver) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (
...
+	if derr != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("Unknown list disks error: %v", derr.Error()))
```
andyzhangx, line 542: `ListVolumes on rg(%s) failed with error: %v`, d.cloud.ResourceGroup

pkg/azuredisk/controllerserver.go
```diff
@@ -524,7 +524,65 @@ func (d *Driver) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (
...
+	}
+
+	if start != 0 && start >= len(disks) {
+		return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%d) is greater than total number of volumes", start)
```
andyzhangx, line 546: `ListVolumes on rg(%s) starting token(%d) is greater than total number of volumes(%d)`

pkg/azuredisk/controllerserver.go
```diff
@@ -524,7 +524,65 @@ func (d *Driver) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (
...
+	}
+
+	maxEntries := len(disks) - start
+	if req.MaxEntries > 0 && int(req.MaxEntries) < maxEntries {
+		maxEntries = int(req.MaxEntries)
+	}
+	pageEnd := start + maxEntries
+
+	entries := []*csi.ListVolumesResponse_Entry{}
+
```
andyzhangx, line 556: no new line here

andyzhangx (commented):
in the error logs, pls provide as much info as possible, that would help in troubleshooting

pkg/azuredisk/controllerserver.go
```diff
@@ -524,7 +524,65 @@ func (d *Driver) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (
...
+	for i := start; i < pageEnd; i++ {
+		d := disks[i]
+		nodeList := []string{}
+
+		if d.ManagedBy != nil {
+			nodeList = append(nodeList, *d.ManagedBy)
```
andyzhangx, line 562: this is incorrect, `ManagedBy` string format is actually not nodeName for VMSS instance, let's remove this implementation in this PR(remove ), and only supports `LIST_VOLUMES` in this PR, and put implementation of `LIST_VOLUMES_PUBLISHED_NODES` in another PR.

pkg/azuredisk/controllerserver.go
```diff
@@ -524,7 +524,65 @@ func (d *Driver) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (
...
+		start, err = strconv.Atoi(req.StartingToken)
+		if err != nil {
+			return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%s) parsing with error: %v", req.StartingToken, err)
+		}
+		if start < 0 {
+			return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%d) can not be negative", start)
+		}
+	}
+
+	disks, derr := d.cloud.DisksClient.ListByResourceGroup(ctx, d.cloud.ResourceGroup)
```
andyzhangx, line 539: actually there are other scenarios that disks are in created other resource group, so ListVolumes result is only accurate for default condition. On the other hand, agent nodes are all in `d.cloud.ResourceGroup`, so ListVolumes of all agent nodes would be accurate. I think this could be right solution: list disks of all agent nodes under `d.cloud.ResourceGroup`

andyzhangx: I am thinking use `d.cloud.KubeClient.CoreV1().PersistentVolumes()` to get all PV info and then get all diskURI from pv.csi.volumeid and `disk.ManagedBy`, that would provide complete disk list, could refer to:
 - `d.cloud.KubeClient.CoreV1()` interface
https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/bb2550070c810ffe94c35eb38aad09c6a3136ade/pkg/azurefile/azurefile.go#L556

andyzhangx: this would require much more work, while it could list all disks this cluster is using, otherwise it could cover all disks and may provide incomplete disk list, not sure what would happen if disk list is incomplete.

pkg/azuredisk/controllerserver.go
andyzhangx: use transform like this, it will provide you the nodeName:
```
			attachedNode, err := vmset.GetNodeNameByProviderID(*disk.ManagedBy)
			if err != nil {
				return -1, err
			}
```

nearora-msft: Ok sure, thanks. Shall I still remove this part and take this in a separate PR?

andyzhangx:
about sanity test failure, need to fix the 5 failures before upgrade to csi sanity test v4.0.2: https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/622

pkg/azuredisk/controllerserver.go
andyzhangx: support `LIST_VOLUMES` and `LIST_VOLUMES_PUBLISHED_NODES` is ok, while current disk list could be incomplete since some disks could be in other resource group, do you know what's the impact if `ListVolumes` return incomplete disk list?

nearora-msft:
> about sanity test failure, need to fix the 5 failures before upgrade to csi sanity test v4.0.2: #622

@andyzhangx I think then we can enable the LIST_VOLUMES_PUBLISHED_NODES only after issue #622 is resolved?

andyzhangx:
> > about sanity test failure, need to fix the 5 failures before upgrade to csi sanity test v4.0.2: #622
> 
> @andyzhangx I think then we can enable the LIST_VOLUMES_PUBLISHED_NODES only after issue #622 is resolved?

another way is disable that failed test in sanity tests temporarily

pkg/azuredisk/controllerserver.go
nearora-msft: AFAIK, ListVolumes is used by the sidecar csi-attacher to reconcile VolumeAttachment objects by comparing the state of VolumeAttachment objects with the ListVolumes response. if the list of disks is incomplete, reconciliation of the VolumeAttachment objects for the corresponding missing disks will be hampered.

andyzhangx: then disable the sanity test failure temporarily here:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/44dedf04edda6f5d22a625a0756a98044fbfb002/test/sanity/run-test.sh#L38

and then work on #622

also open an issue that `ListVolume` does not work if disks are created in external rg, I think finally we may use `d.cloud.KubeClient.CoreV1().PersistentVolumes()` to reimplement the ListVolumes logic, that could discover all disks.

pkg/azuredisk/controllerserver_test.go
andyzhangx: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/627 is merged

nearora-msft:
> > > about sanity test failure, need to fix the 5 failures before upgrade to csi sanity test v4.0.2: #622
> > 
> > 
> > @andyzhangx I think then we can enable the LIST_VOLUMES_PUBLISHED_NODES only after issue #622 is resolved?
> 
> another way is disable that failed test in sanity tests temporarily

Disabled the failing sanity test temporarily

pkg/azuredisk/controllerserver_test.go
```diff
@@ -1481,13 +1481,7 @@ func TestGetCapacity(t *testing.T) {
...
-	req := csi.ListVolumesRequest{}
-	resp, err := d.ListVolumes(context.Background(), &req)
-	assert.Nil(t, resp)
-	if !reflect.DeepEqual(err, status.Error(codes.Unimplemented, "")) {
-		t.Errorf("Unexpected error: %v", err)
-	}
+	t.Skip("Skipping this test. Tracking with issue #617")
```
andyzhangx, line 1484: could you enable this ut?

andyzhangx (commented):
pls add csc integration test and squash all commits by `git rebase -i HEAD~7`

test/sanity/run-test.sh
```diff
@@ -35,4 +35,4 @@ _output/azurediskplugin --endpoint "$endpoint" --nodeid "$nodeid" -v=5 &
 
 echo 'Begin to run sanity test...'
 readonly CSI_SANITY_BIN='csi-sanity'
-"$CSI_SANITY_BIN" --ginkgo.v --csi.endpoint="$endpoint" --ginkgo.skip='should work|should fail when volume does not exist on the specified path'
+"$CSI_SANITY_BIN" --ginkgo.v --csi.endpoint="$endpoint" --ginkgo.skip='should work|should fail when volume does not exist on the specified path|should return appropriate capabilities'
```
andyzhangx, line 38: also add an issue to track this disable issue: `should return appropriate capabilities`

pkg/azuredisk/controllerserver.go
nearora-msft: Issue to track the external resource group scenario : https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/632

test/sanity/run-test.sh
nearora-msft: https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/633

nearora-msft:
> csc

Added the integration test. Will squash commits once all the checks complete successfully.

andyzhangx:
```
• Failure [17.089 seconds]
Controller Service [Controller Server]
/home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
  ListVolumes
  /home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:166
    pagination should detect volumes added between pages and accept tokens when the last volume from a page is deleted [It]
skipped 1156 lines unfold_more
[Fail] Controller Service [Controller Server] ListVolumes [It] pagination should detect volumes added between pages and accept tokens when the last volume from a page is deleted 
/home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:336
Ran 63 of 72 Specs in 503.670 seconds
FAIL! -- 62 Passed | 1 Failed | 0 Pending | 9 Skipped
```

nearora-msft:
> ```
> • Failure [17.089 seconds]
> Controller Service [Controller Server]
> /home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
>   ListVolumes
>   /home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:166
>     pagination should detect volumes added between pages and accept tokens when the last volume from a page is deleted [It]
> skipped 1156 lines unfold_more
> [Fail] Controller Service [Controller Server] ListVolumes [It] pagination should detect volumes added between pages and accept tokens when the last volume from a page is deleted 
> /home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:336
> Ran 63 of 72 Specs in 503.670 seconds
> FAIL! -- 62 Passed | 1 Failed | 0 Pending | 9 Skipped
> ```

Disabled this test because of existing issue https://github.com/kubernetes-csi/csi-test/issues/223

andyzhangx:
would you squash all commits first? thanks.

andyzhangx (approved):
/lgtm
/approve

andyzhangx:
looks like it's a flacky test, could you fix it? thanks:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_azuredisk-csi-driver/618/pull-azuredisk-csi-driver-sanity/1341304585670627328

/hold

nearora-msft:
> looks like it's a flacky test, could you fix it? thanks:
> https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_azuredisk-csi-driver/618/pull-azuredisk-csi-driver-sanity/1341304585670627328
> 
> /hold

The test that is failing performs these steps:
1. Call List Volumes
2. Create a volume
3. List Volumes again

It expects that the List Volumes count in step 3 should be List Volumes count in step 1 + 1. But the actual case is that List Volumes count in step 3 = List Volumes count in step1 +2. This is why the test fails.

I have observed that when a vm is created a disk is attached to it by default. While running the tests, we create an rg and a sanity-test-node in that rg . The creation of this node should create a disk and attach is to the newly created node. So when the ListVolumes call is made the 1st time, it usually lists the disk that was created with the creation of sanity-test-node. So, we get 1 entry in the 1st call and 2 entries in the 2nd call.

But in this failed case, the 1st ListVolumes call doesn't give any results(Could be because the default disk attach has not happened yet ) And when the ListVolumes call is made the second time, it returns 2 disks(One that was created by the vm creation and the other that was created by the test)

Not sure how we can fix it at our end(if that's possible)

PR #635, opened 2020-12-21
[WIP]feat: use disk attach/detach batch operation

andyzhangx:
**What type of PR is this?**
/kind feature

feat: use disk attach/detach batch operation
This PR use batch operation for disk attach/detach, that would improve disk attach/detach operation speed quite a lot, e.g. if we attach 20 disks to one vmss instance at one time, with this PR, only two disk attach operations would be triggered, that's 10x faster than original.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
feat: use disk attach/detach batch operation
```


andyzhangx:
already addressed by https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/650

PR #642, opened 2020-12-24
fix: List only snapshots that match the volumeId

### LINKED ISSUE #641: ListSnapshots returns all the snapshots when a specific VolumeId is given
**What happened**:
ListSnapshots returns all the snapshots in an rg even when a specific volumeId is given

ListSnapshots  should return only the snapshots made for that particular volume whose volumeId is given in request

**How to reproduce it**:

This is causing one of the failures mentioned in #622 

**Environment**:
- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

nearora-msft:
**What type of PR is this?**
/kind bug

ListSnapshot implementation currently returns all the snapshots in an rg even when a specific volumeId is given. This fix will only return snapshots related to a particular volume when the volumeId is given.

Fixes #641 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -974,33 +974,26 @@ func getEntriesAndNextToken(req *csi.ListSnapshotsRequest, snapshots []compute.S
 			return nil, status.Errorf(codes.Aborted, "ListSnapshots starting token(%d) can not be negative", start)
 		}
 	}
-	perPage := 0
-	if req.MaxEntries > 0 {
-		perPage = int(req.MaxEntries)
-	}
 
-	match := false
-	entries := []*csi.ListSnapshotsResponse_Entry{}
-	for i, snapshot := range snapshots {
-		if req.SourceVolumeId == getSourceVolumeID(&snapshot) {
-			match = true
-		}
-		csiSnapshot, err := generateCSISnapshot(req.SourceVolumeId, &snapshot)
-		if err != nil {
-			return nil, fmt.Errorf("failed to generate snapshot entry: %v", err)
-		}
-		if i >= start && (perPage == 0 || (perPage > 0 && i < start+perPage)) {
-			entries = append(entries, &csi.ListSnapshotsResponse_Entry{Snapshot: csiSnapshot})
-		}
+	maxEntries := len(snapshots) - start
+	if req.MaxEntries > 0 && int(req.MaxEntries) < maxEntries {
+		maxEntries = int(req.MaxEntries)
 	}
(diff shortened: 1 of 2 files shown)


pkg/azuredisk/controllerserver.go
```diff
@@ -974,33 +974,27 @@ func getEntriesAndNextToken(req *csi.ListSnapshotsRequest, snapshots []compute.S
...
-	for i, snapshot := range snapshots {
-		if req.SourceVolumeId == getSourceVolumeID(&snapshot) {
-			match = true
-		}
-		csiSnapshot, err := generateCSISnapshot(req.SourceVolumeId, &snapshot)
-		if err != nil {
-			return nil, fmt.Errorf("failed to generate snapshot entry: %v", err)
-		}
-		if i >= start && (perPage == 0 || (perPage > 0 && i < start+perPage)) {
+	count := 0
```
andyzhangx, line 983: what about use original logic (I think it's more clear):
```
	for i, snapshot := range snapshots {
		if req.SourceVolumeId == getSourceVolumeID(&snapshot) {
			match = true
		}
		csiSnapshot, err := generateCSISnapshot(req.SourceVolumeId, &snapshot)
		if err != nil {
			return nil, fmt.Errorf("failed to generate snapshot entry: %v", err)
		}
		if i >= start && (perPage == 0 || (perPage > 0 && i < start+perPage)) {
			if match || req.SourceVolumeId == "" {
				entries = append(entries, &csi.ListSnapshotsResponse_Entry{Snapshot: csiSnapshot})
			}
		}
	}
```

that would be a simple fix?

nearora-msft: Hmm this logic will not work properly when only some of the snapshots are linked to the VolumeId given in request. Lets say we have total number of snapshots as 3 and req.maxEntries=1 and start=0. Lets say that snapshot[2] matched the required volumeId. But with this logic we are only considering the scenarios when i<start+perPage. The entry that we are looking for is snapshot[2] but it will never be added because when i=2,  i becomes greater than start+perPage - which is 1

 We should iterate and check all the entries until we find maxEntries(which is 1 in this case) or until we have have exhausted all the snapshots.

andyzhangx: thanks for the explanation. one single comment, move `count=0` to:
```
for count=0; start < len(snapshots) && count < maxEntries; start++ {
```

andyzhangx (approved):
/lgtm
/approve

PR #646, opened 2020-12-28
feat: Add tags that in-tree driver uses

### LINKED ISSUE #579: add tags that in-tree driver uses

Currently only one tag is added by default for CSI driver:
image

Should add same or similar tags as in-tree driver:
image

**Additional context**

kassarl:
**What type of PR is this?**
/kind feature

Adds the tags that the in-tree driver uses a

Fixes #579 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

N/A

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -111,6 +111,9 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		diskName                string
 		diskEncryptionSetID     string
 		customTags              string
+		pvcNameKey              string
+		pvcNamespaceKey         string
+		pvNameKey               string
 		writeAcceleratorEnabled string
 		maxShares               int
 	)
@@ -119,6 +122,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	if parameters == nil {
 		parameters = make(map[string]string)
 	}
+
 	for k, v := range parameters {
 		switch strings.ToLower(k) {
 		case skuNameField:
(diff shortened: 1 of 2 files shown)


pkg/azuredisk/controllerserver.go
```diff
@@ -145,6 +149,12 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			diskEncryptionSetID = v
 		case tagsField:
 			customTags = v
+		case pvNameKeyField:
```
andyzhangx, line 152: these values should be from k8s other than user input, as I could see in below logs, `kubernetes.io-created-for-pv-name` could be `req.GetName()`, I did not see we could get pvc name from `CreateVolume` now, let's add `kubernetes.io-created-for-pv-name`

```
I1225 14:40:04.687393       1 utils.go:84] GRPC call: /csi.v1.Controller/CreateVolume
I1225 14:40:04.687413       1 utils.go:85] GRPC request: {"accessibility_requirements":{"preferred":[{"segments":{"topology.disk.csi.azure.com/zone":""}}],"requisite":[{"segments":{"topology.disk.csi.azure.com/zone":""}}]},"capacity_range":{"required_bytes":10737418240},"name":"pvc-2eccae3e-2f51-43d1-ae03-9bfb87fe5135","parameters":{"skuName":"Standard_LRS"},"volume_capabilities":[{"AccessType":{"Mount":{"fs_type":"ext4","mount_flags":["barrier=1","acl"]}},"access_mode":{"mode":1}}]}
```

kassarl: The idea behind my changes is that if you add the --extra-create-metadata tag in the arguments for csi-provisioner in the csi-azuredisk-controller.yaml, it would pass those values to CreateVolume. Is this what you mean by user input?

andyzhangx: those tags are pv, pvc names, it's dynamic, we cannot input those names in storage class since one storage class could provision hundreds of pv/pvc, I have created a PR to add `kubernetes.io-created-for-pv-name`: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/649, looks like there is no way to get `kubernetes.io-created-for-pvc-name`, `kubernetes.io-created-for-pvc-namespace` in CSI CreateVolume interface.

PR #676, opened 2021-01-12
test: Use the resource group's managedBy property if none provided

### LINKED ISSUE #675: e2e tests fail when using a resource group having a non-nil `managedBy` property
**What happened**:

When using an cluster through Azure Kubernetes Service (either through the portal or CLI), it creates a separate resource group that has the `managedBy` property set to a reference to the resource group hosting the cluster. This "managed" resource group is where dynamic PVs are created by default. If you specific this resource group as the one to use for e2e tests, the tests fail to initialize with the error;

> The managed property of the resource group cannot be changed from its current value...

The e2e tests should work without error.

**How to reproduce it**:

1. Create an AKS cluster.
2. Specify the name of the "MC_*" resource group in the AZURE_RESOURCE_GROUP environment variable when setting up the e2e tests.
3. Run the e2e tests.

**Environment**:
- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

edreed:
If `nil` is passed to the `managedBy` parameter of `azure.Client.EnsureResourceGroup()` method, use resource group's current value when calling `resources.groupsClient.CreateOrUpdate()`.

**What type of PR is this?**

/kind bug

This is convenient for those who use AKS to create and manage test clusters.

Fixes #675

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/test/utils/azure/azure_helpers.go
+++ b/test/utils/azure/azure_helpers.go
@@ -75,6 +75,9 @@ func (az *Client) EnsureResourceGroup(ctx context.Context, name, location string
 	} else {
 		tags = make(map[string]*string)
 	}
+	if managedBy == nil {
+		managedBy = group.ManagedBy
+	}
 	// Tags for correlating resource groups with prow jobs on testgrid
 	tags["buildID"] = stringPointer(os.Getenv("BUILD_ID"))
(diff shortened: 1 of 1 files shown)


andyzhangx (approved):
/ok-to-test
/lgtm
/approve

PR #725, opened 2021-02-09
fix: ListVolumes should be able to return volumes from external rgs as well

### LINKED ISSUE #632: ListVolumes API doesn't list the volumes that are created in an external resource group
**What happened**:
ListVolumes API doesn't list the volumes that are created in a different resource group

ListVolumes API should list all the disks that are attached to the nodes in a resource group

**How to reproduce it**:

With azuredisk csi driver, disks might get created in different resource groups. The current implementation of ListVolumes(https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/618) only supports the default case when the disks are created in the current resource group

**Environment**:
- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sunpa93:
**What type of PR is this?**

/kind bug

Fixes #632 

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -29,9 +30,9 @@ import (
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	cloudprovider "k8s.io/cloud-provider"
 	volerr "k8s.io/cloud-provider/volume/errors"
@@ -72,6 +73,13 @@ const (
 	resizeRequired           = "resizeRequired"
 	sourceDiskSearchMaxDepth = 10
 )
+
+type listVolumeStatus struct {
+	numVisited    int
+	isCompleteRun bool
+	entries       []*csi.ListVolumesResponse_Entry
+	err           error
+}
 
 // CreateVolume provisions an azure disk
 func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
@@ -506,9 +514,9 @@ func (d *Driver) GetCapacity(ctx context.Context, req *csi.GetCapacityRequest) (
 
 // ListVolumes return all available volumes
 func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (*csi.ListVolumesResponse, error) {
-	var err error
 	start := 0
 	if req.StartingToken != "" {
+		var err error
 		start, err = strconv.Atoi(req.StartingToken)
 		if err != nil {
 			return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%s) parsing with error: %v", req.StartingToken, err)
@@ -517,52 +525,83 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
 			return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%d) can not be negative", start)
 		}
 	}
+	if d.cloud.KubeClient != nil && d.cloud.KubeClient.CoreV1() != nil && d.cloud.KubeClient.CoreV1().PersistentVolumes() != nil {
+		klog.V(6).Infof("List Volumes in Cluster:")
+		return d.listVolumesInCluster(ctx, start, int(req.MaxEntries))
+	}
+	klog.V(6).Infof("List Volumes in Node Resource Group:")
+	return d.listVolumesInNodeResourceGroup(ctx, start, int(req.MaxEntries))
+}
 
-	disks, derr := d.cloud.DisksClient.ListByResourceGroup(ctx, d.cloud.ResourceGroup)
-	if derr != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("ListVolumes on rg(%s) failed with error: %v", d.cloud.ResourceGroup, derr.Error()))
+// listVolumesInCluster is a helper function for ListVolumes used for when there is an available kubeclient
+func (d *Driver) listVolumesInCluster(ctx context.Context, start, maxEntries int) (*csi.ListVolumesResponse, error) {
+	pvList, err := d.cloud.KubeClient.CoreV1().PersistentVolumes().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "ListVolumes failed while fetching PersistentVolumes List with error: %v", err.Error())
 	}
 
-	if start != 0 && start >= len(disks) {
-		return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%d) on rg(%s) is greater than total number of volumes", start, d.cloud.ResourceGroup)
-	}
-
-	maxEntries := len(disks) - start
-	if req.MaxEntries > 0 && int(req.MaxEntries) < maxEntries {
-		maxEntries = int(req.MaxEntries)
-	}
-	pageEnd := start + maxEntries
-
-	vmSet := d.cloud.VMSet
-	entries := []*csi.ListVolumesResponse_Entry{}
-	for i := start; i < pageEnd; i++ {
-		d := disks[i]
-		nodeList := []string{}
-
-		// HyperVGeneration property is only setup for os disks. Only the non os disks should be included in the list
-		if d.DiskProperties == nil || d.DiskProperties.HyperVGeneration == "" {
-			if d.ManagedBy != nil {
-				attachedNode, err := vmSet.GetNodeNameByProviderID(*d.ManagedBy)
-				if err != nil {
-					return nil, err
-				}
-				nodeList = append(nodeList, string(attachedNode))
+	// get all resource groups and put them into a sorted slice
+	rgMap := make(map[string]bool)
+	volSet := make(map[string]bool)
+	for _, pv := range pvList.Items {
+		if pv.Spec.CSI != nil && pv.Spec.CSI.Driver == DriverName {
+			diskURI := pv.Spec.CSI.VolumeHandle
+			if err := isValidDiskURI(diskURI); err != nil {
+				klog.Warningf("invalid disk uri (%s) with error(%v)", diskURI, err)
+				continue
 			}
-
-			entries = append(entries, &csi.ListVolumesResponse_Entry{
-				Volume: &csi.Volume{
-					VolumeId: *d.ID,
-				},
-				Status: &csi.ListVolumesResponse_VolumeStatus{
-					PublishedNodeIds: nodeList,
-				},
-			})
+			rg, err := GetResourceGroupFromURI(diskURI)
+			if err != nil {
+				klog.Warningf("failed to get resource group from disk uri (%s) with error(%v)", diskURI, err)
+				continue
+			}
+			volSet[diskURI] = true
+			if _, visited := rgMap[rg]; visited {
+				continue
+			}
+			rgMap[rg] = true
 		}
 	}
 
+	if start != 0 && start > len(volSet) {
+		return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%d) on rg(%s) is greater than total number of volumes", start, d.cloud.ResourceGroup)
+	}
+	resourceGroups := make([]string, len(rgMap))
+	i := 0
+	for rg := range rgMap {
+		resourceGroups[i] = rg
+		i++
+	}
+	sort.Strings(resourceGroups)
+
+	// loop through each resourceGroup to get disk lists
+	entries := []*csi.ListVolumesResponse_Entry{}
+	numVisited := 0
+	isCompleteRun, startFound := true, false
+	for _, resourceGroup := range resourceGroups {
+		if !isCompleteRun || (maxEntries > 0 && len(entries) >= maxEntries) {
+			isCompleteRun = false
+			break
+		}
+		localStart := start - numVisited
+		if startFound {
+			localStart = 0
+		}
+		listStatus := d.listVolumesByResourceGroup(ctx, resourceGroup, entries, localStart, maxEntries-len(entries), &volSet)
+		numVisited += listStatus.numVisited
+		if listStatus.err != nil {
+			if status.Code(listStatus.err) == codes.FailedPrecondition {
+				continue
+			}
+			return nil, listStatus.err
+		}
+		startFound = true
+		entries = listStatus.entries
+		isCompleteRun = isCompleteRun && listStatus.isCompleteRun
+	}
 	nextTokenString := ""
-	if pageEnd < len(disks) {
-		nextTokenString = strconv.Itoa(pageEnd)
+	if !isCompleteRun {
+		nextTokenString = strconv.Itoa(start + numVisited)
 	}
 
 	listVolumesResp := &csi.ListVolumesResponse{
@@ -571,6 +610,85 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
 	}
 
 	return listVolumesResp, nil
+}
+
+// listVolumesInNodeResourceGroup is a helper function for ListVolumes used for when there is no available kubeclient
+func (d *Driver) listVolumesInNodeResourceGroup(ctx context.Context, start, maxEntries int) (*csi.ListVolumesResponse, error) {
+	entries := []*csi.ListVolumesResponse_Entry{}
+	listStatus := d.listVolumesByResourceGroup(ctx, d.cloud.ResourceGroup, entries, start, maxEntries, nil)
+	if listStatus.err != nil {
+		return nil, listStatus.err
+	}
+
+	nextTokenString := ""
+	if !listStatus.isCompleteRun {
+		nextTokenString = strconv.Itoa(listStatus.numVisited)
+	}
+
+	listVolumesResp := &csi.ListVolumesResponse{
+		Entries:   listStatus.entries,
+		NextToken: nextTokenString,
+	}
+
+	return listVolumesResp, nil
+}
+
+// listVolumesByResourceGroup is a helper function that updates the ListVolumeResponse_Entry slice and returns number of total visited volumes, number of volumes that needs to be visited and an error if found
+func (d *Driver) listVolumesByResourceGroup(ctx context.Context, resourceGroup string, entries []*csi.ListVolumesResponse_Entry, start, maxEntries int, volSet *map[string]bool) listVolumeStatus {
+	disks, derr := d.cloud.DisksClient.ListByResourceGroup(ctx, resourceGroup)
+	if derr != nil {
+		return listVolumeStatus{err: status.Errorf(codes.Internal, "ListVolumes on rg(%s) failed with error: %v", resourceGroup, derr.Error())}
+	}
+	if start > 0 && start >= len(disks) {
+		return listVolumeStatus{
+			numVisited: len(disks),
+			err:        status.Errorf(codes.FailedPrecondition, "ListVolumes starting token(%d) on rg(%s) is greater than total number of volumes", start, d.cloud.ResourceGroup),
+		}
+	}
+	if start < 0 {
+		start = 0
+	}
+	i := start
+	isCompleteRun := true
+	// Loop until
+	for ; i < len(disks); i++ {
+		if maxEntries > 0 && len(entries) >= maxEntries {
+			isCompleteRun = false
+			break
+		}
+
+		disk := disks[i]
+		// if given a set of volumes from KubeClient, only continue if the disk can be found in the set
+		if volSet != nil && !(*volSet)[*disk.ID] {
+			continue
+		}
+		// HyperVGeneration property is only setup for os disks. Only the non os disks should be included in the list
+		if disk.DiskProperties == nil || disk.DiskProperties.HyperVGeneration == "" {
+			nodeList := []string{}
+
+			if disk.ManagedBy != nil {
+				attachedNode, err := d.cloud.VMSet.GetNodeNameByProviderID(*disk.ManagedBy)
+				if err != nil {
+					return listVolumeStatus{err: err}
+				}
+				nodeList = append(nodeList, string(attachedNode))
+			}
+
+			entries = append(entries, &csi.ListVolumesResponse_Entry{
+				Volume: &csi.Volume{
+					VolumeId: *disk.ID,
+				},
+				Status: &csi.ListVolumesResponse_VolumeStatus{
+					PublishedNodeIds: nodeList,
+				},
+			})
+		}
+	}
+	return listVolumeStatus{
+		numVisited:    i - start,
+		isCompleteRun: isCompleteRun,
+		entries:       entries,
+	}
 }
 
 // ControllerExpandVolume controller expand volume
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -74,9 +74,10 @@ const (
 	sourceDiskSearchMaxDepth = 10
 )
 
+// listVolumeStatus explains the return status of `listVolumesByResourceGroup`
 type listVolumeStatus struct {
-	numVisited    int
-	isCompleteRun bool
+	numVisited    int  // the number of iterated azure disks
+	isCompleteRun bool // isCompleteRun is flagged true if the function iterated through all azure disks
 	entries       []*csi.ListVolumesResponse_Entry
 	err           error
 }
@@ -529,7 +530,7 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
 		klog.V(6).Infof("List Volumes in Cluster:")
 		return d.listVolumesInCluster(ctx, start, int(req.MaxEntries))
 	}
-	klog.V(6).Infof("List Volumes in Node Resource Group:")
+	klog.V(6).Infof("List Volumes in Node Resource Group: %s", d.cloud.ResourceGroup)
 	return d.listVolumesInNodeResourceGroup(ctx, start, int(req.MaxEntries))
 }
 
@@ -555,6 +556,7 @@ func (d *Driver) listVolumesInCluster(ctx context.Context, start, maxEntries int
 				klog.Warningf("failed to get resource group from disk uri (%s) with error(%v)", diskURI, err)
 				continue
 			}
+			rg, diskURI = strings.ToLower(rg), strings.ToLower(diskURI)
 			volSet[diskURI] = true
 			if _, visited := rgMap[rg]; visited {
 				continue
@@ -563,9 +565,6 @@ func (d *Driver) listVolumesInCluster(ctx context.Context, start, maxEntries int
 		}
 	}
 
-	if start != 0 && start > len(volSet) {
-		return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%d) on rg(%s) is greater than total number of volumes", start, d.cloud.ResourceGroup)
-	}
 	resourceGroups := make([]string, len(rgMap))
 	i := 0
 	for rg := range rgMap {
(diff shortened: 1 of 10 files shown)


andyzhangx (commented):
/ok-to-test

pkg/azuredisk/controllerserver.go
```diff
@@ -518,51 +518,70 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
 		}
 	}
 
-	disks, derr := d.cloud.DisksClient.ListByResourceGroup(ctx, d.cloud.ResourceGroup)
-	if derr != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("ListVolumes on rg(%s) failed with error: %v", d.cloud.ResourceGroup, derr.Error()))
+	if d.cloud.KubeClient == nil || d.cloud.KubeClient.CoreV1() == nil || d.cloud.KubeClient.CoreV1().PersistentVolumes() == nil {
```
andyzhangx, line 521: if kubeClient is not set, should return `DisksClient.ListByResourceGroup` result instead of error

pkg/azuredisk/controllerserver.go
```diff
@@ -518,51 +518,70 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+			}
+			diskName, err := GetDiskName(diskURI)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "could not get disk name from diskURI(%s) with error(%v)", diskURI, err)
+			}
+			resourceGroup, err := GetResourceGroupFromURI(diskURI)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "could not get resource group from diskURI(%s) with error(%v)", diskURI, err)
+			}
+			disk, derr := d.cloud.DisksClient.Get(ctx, resourceGroup, diskName)
```
andyzhangx, line 554: if there are 1K disks, there would be 1K get disk calls, that's too expensive, is it possible to get all resource groups and then `ListByResourceGroup`, that will save get disk calls a lot. Also would you move this code to a standalone function?

sunpa93: Thank you for pointing that out.

andyzhangx:
`/ListVolumes` still shows `k8s-master-23068744-0-etcddisk` disk which is actually not a disk PVC in current testing cluster, with this PR, I think it should filter out non disk PVC:

```
I0210 13:01:32.449448       1 utils.go:95] GRPC call: /csi.v1.Controller/ListVolumes
I0210 13:01:32.449477       1 utils.go:96] GRPC request: {}
I0210 13:01:32.478981       1 utils.go:102] GRPC response: {"entries":[{"status":{"published_node_ids":["k8s-master-23068744-0"]},"volume":{"volume_id":"/subscriptions/0e46bd28-a80f-4d3a-8200-d9eb8d80cb2e/resourceGroups/kubetest-b3c1e461-6b9e-11eb-ae9b-f6d3fab2f43c/providers/Microsoft.Compute/disks/k8s-master-23068744-0-etcddisk"}},{"status":{"published_node_ids":["k8s-agentpool1-23068744-0"]},"volume":{"volume_id":"/subscriptions/0e46bd28-a80f-4d3a-8200-d9eb8d80cb2e/resourceGroups/kubetest-b3c1e461-6b9e-11eb-ae9b-f6d3fab2f43c/providers/Microsoft.Compute/disks/shared-disk-multiple-pods"}}]}
```

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_azuredisk-csi-driver/725/pull-azuredisk-csi-driver-e2e-single-az/1359484751718649856/build-log.txt

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +518,152 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "ListVolumes failed while fetching PersistentVolumes List with error: %v", err.Error())
+	}
+
+	// get all resource groups and put them into a sorted slice
+	rgMap := make(map[string]bool)
+	volSet := make(map[string]bool)
+	numVols := 0
+	for _, pv := range pvList.Items {
+		if pv.Spec.CSI != nil && pv.Spec.CSI.Driver == "disk.csi.azure.com" {
```
andyzhangx, line 539: use `DriverName`  since it's already defined

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +518,152 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+			diskURI := pv.Spec.CSI.VolumeHandle
+			if err := isValidDiskURI(diskURI); err != nil {
+				return nil, status.Errorf(codes.InvalidArgument, "disk URI(%s) is not valid: %v", diskURI, err)
```
andyzhangx, line 542: better provide a warning and `continue`

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +518,152 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+			}
+			rg, err := GetResourceGroupFromURI(diskURI)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "could not get resource group from diskURI(%s) with error(%v)", diskURI, err)
```
andyzhangx, line 546: better provide a warning and continue

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +518,152 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
-		maxEntries = int(req.MaxEntries)
+	// loop through each resourceGroup to get disk lists
+	entries := []*csi.ListVolumesResponse_Entry{}
+	numEntries, numVisited, numNotVisited := 0, 0, 0
+	for _, resourceGroup := range resourceGroups {
+		if maxEntries > 0 && numEntries >= maxEntries {
+			numNotVisited++
+			break
+		}
+		derr := d.listVolumesByResourceGroup(ctx, resourceGroup, &entries, start, maxEntries, &numVisited, &numNotVisited, &numEntries, &volSet)
```
andyzhangx, line 576: should only list volumes which is in `pvList.Items`

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +518,152 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+	listVolumesResp := &csi.ListVolumesResponse{
+		Entries:   entries,
+		NextToken: nextTokenString,
+	}
+
+	return listVolumesResp, nil
+}
+
+// listVolumesWithoutKubeClient is a helper function for ListVolumes used for when there is no available kubeclient
+func (d *Driver) listVolumesWithoutKubeClient(ctx context.Context, start, maxEntries int) (*csi.ListVolumesResponse, error) {
```
andyzhangx, line 599: `listVolumesInNodeResourceGroup`

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +518,152 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+		return d.listVolumesWithKubeClient(ctx, start, int(req.MaxEntries))
+	}
+	return d.listVolumesWithoutKubeClient(ctx, start, int(req.MaxEntries))
+}
 
-	disks, derr := d.cloud.DisksClient.ListByResourceGroup(ctx, d.cloud.ResourceGroup)
-	if derr != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("ListVolumes on rg(%s) failed with error: %v", d.cloud.ResourceGroup, derr.Error()))
+// listVolumesWithKubeClient is a helper function for ListVolumes used for when there is an available kubeclient
+func (d *Driver) listVolumesWithKubeClient(ctx context.Context, start, maxEntries int) (*csi.ListVolumesResponse, error) {
```
andyzhangx, line 528: `listVolumesInCluster`

pkg/azuredisk/controllerserver.go
andyzhangx: I could still find etcd disks in `ListVolumes` logs, I think that's the root cause: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_azuredisk-csi-driver/725/pull-azuredisk-csi-driver-e2e-single-az/1359837206172143616/build-log.txt

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +518,152 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+			if _, visited := rgMap[rg]; visited {
+				continue
+			}
+			rgMap[rg] = true
+		}
+		numVols++
 	}
 
-	if start != 0 && start >= len(disks) {
+	if start != 0 && start > numVols {
```
nearora-msft, line 557: Can we do if start!=0 && start>len(pvList.Items) and move this before the above mentioned loop?

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +518,152 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
-		nextTokenString = strconv.Itoa(pageEnd)
-	}
-
-	listVolumesResp := &csi.ListVolumesResponse{
-		Entries:   entries,
-		NextToken: nextTokenString,
-	}
-
-	return listVolumesResp, nil
+	*numVisited, *numNotVisited = *numVisited+i, *numNotVisited-i
```
nearora-msft, line 665: Changing the values of these variables(numVisited, numNotVisited etc) by reference might lead to synchronization issues if multiple go routines are accessing these at the same time. That might make the values of these variables inconsistent across goroutines.
It would be better to pass these variables into this method by value, change(increment/decrement) the values and then send these to the calling method as return values.

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +518,152 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+		disk := disks[i]
+		// if given a set of volumes from KubeClient, only continue if the disk can be found in the set
+		if volSet != nil && (*volSet)[*disk.ID] {
+			continue
+		}
 		// HyperVGeneration property is only setup for os disks. Only the non os disks should be included in the list
-		if d.DiskProperties == nil || d.DiskProperties.HyperVGeneration == "" {
-			if d.ManagedBy != nil {
-				attachedNode, err := vmSet.GetNodeNameByProviderID(*d.ManagedBy)
+		if disk.DiskProperties == nil || disk.DiskProperties.HyperVGeneration == "" {
```
nearora-msft, line 643: If some of these disks are os disks, they will not be added to the list of entries. So the actual list of entries will be less than the maxEntries even if the len(disks) >= maxEntries. We should keep running the loop as long as there are disks to iterate over and len(entries) < maxEntries.

pkg/azuredisk/controllerserver.go
sunpa93: > should only list volumes which is in `pvList.Items`

I also thought that this was the cause and have added a `volSet` to keep track of the volumes in `pvList.Items` and only append the disk to the entry if it can be found in `volSet`. However, if etcd disks are still showing up in the results, I think the cause may be something else.

pkg/azuredisk/controllerserver.go
sunpa93: `pvList.Items` can contain volumes that are not of azure csi driver, so it is inevitable that we loop through each item and confirm that the volume is azure csi disk to get a proper tally of the total volumes `ListVolumes` will return.

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +525,167 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
 			return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%d) can not be negative", start)
 		}
 	}
+	if d.cloud.KubeClient != nil && d.cloud.KubeClient.CoreV1() != nil && d.cloud.KubeClient.CoreV1().PersistentVolumes() != nil {
+		klog.V(2).Infof("List Volumes in Cluster:")
```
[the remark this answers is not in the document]
andyzhangx: there is no `List Volumes in Cluster` in the logs https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_azuredisk-csi-driver/725/pull-azuredisk-csi-driver-e2e-windows/1362210775858417664/build-log.txt, which means e2e won't List Volumes in Cluster, need to figure out, seems `d.cloud.KubeClient.CoreV1().PersistentVolumes` is nil here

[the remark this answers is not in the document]
andyzhangx: I have pushed a fix: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/725/commits/ea802d8d85c8582302f9a326bae77d6e7800bd57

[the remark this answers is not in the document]
sunpa93: I added the log to see if `listVolumesInCluster` was being called at all during the test. This probably explains why there were etcd disks in ListVolumes logs then.

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +525,167 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+				klog.Warningf("failed to get resource group from disk uri (%s) with error(%v)", diskURI, err)
+				continue
+			}
+			volSet[diskURI] = true
+			if _, visited := rgMap[rg]; visited {
+				continue
+			}
+			rgMap[rg] = true
+		}
+		numVols++
```
nearora-msft, line 565: Should we increment this within the if block starting on line 548?

sunpa93: You are right. I don't know how the increment ended up there.

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +525,165 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+		}
+		listStatus := d.listVolumesByResourceGroup(ctx, resourceGroup, entries, start-numVisited, maxEntries-len(entries), &volSet)
+		numVisited += listStatus.numVisited
+		if listStatus.err != nil {
+			if status.Code(listStatus.err) == codes.FailedPrecondition {
+				continue
+			}
+			return nil, listStatus.err
+		}
+		entries = listStatus.entries
```
nearora-msft, line 594: Shouldn't we append the entries here instead of directly setting them. If we set it like this the entries from the previously iterated resource group will get overridden.

sunpa93: entries get passed as a parameter to `listVolumesByResourceGroup` so the returned entries from `listVolumesByResourceGroup` will be always be appended entries including those from previously iterated resource group. So it shouldn't be a problem :)

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +525,165 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+	// loop through each resourceGroup to get disk lists
+	entries := []*csi.ListVolumesResponse_Entry{}
+	numVisited := 0
+	isCompleteRun := true
+	for _, resourceGroup := range resourceGroups {
+		if !isCompleteRun || (maxEntries > 0 && len(entries) >= maxEntries) {
+			isCompleteRun = false
+			break
+		}
+		listStatus := d.listVolumesByResourceGroup(ctx, resourceGroup, entries, start-numVisited, maxEntries-len(entries), &volSet)
```
nearora-msft, line 586: I am not sure if start = start-numVisited will work as expected in all the cases. Let's assume that start = 6 and the number of disks in this resource group is 10. So by the listVolumesByResourceGroup logic, numVisited will be i-start i.e.  10-6 = 4. 

So, in the next iteration start = start- numVisited will be 6-4 = 2, while in reality it should be 0 because we have already moved 6 places ahead in the first iteration.

pkg/azuredisk/controllerserver.go
nearora-msft: Oh ofcourse! My bad.

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +525,165 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
-		d := disks[i]
-		nodeList := []string{}
+	listStatus := d.listVolumesByResourceGroup(ctx, d.cloud.ResourceGroup, entries, start, maxEntries, nil)
+	if listStatus.err != nil {
+		return nil, listStatus.err
+	}
+
+	nextTokenString := ""
+	if !listStatus.isCompleteRun {
+		nextTokenString = strconv.Itoa(listStatus.numVisited)
```
nearora-msft, line 620: I'm not sure if numVisited can work as the nextToken string in all the scenarios. 
Consider the scenario that I mentioned above when start = 6 and there's only 1 rg with 10 disks. The value of numVisited would be 4 and that will also be the next token string. When the call is made to ListVolumes the next time, it will be with startingToken=4  While actually it should be 10.

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +525,165 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+				continue
+			}
+			return nil, listStatus.err
+		}
+		entries = listStatus.entries
+		isCompleteRun = isCompleteRun && listStatus.isCompleteRun
+	}
+	nextTokenString := ""
+	if !isCompleteRun {
+		nextTokenString = strconv.Itoa(numVisited)
```
nearora-msft, line 599: I'm not sure if numVisited can work as the nextToken string in all the scenarios.
Consider the scenario that I mentioned above when start = 6 and there's only 1 rg with 10 disks. The value of numVisited would be 4 and that will also be the next token string. When the call is made to ListVolumes the next time, it will be with startingToken=4 While actually it should be 10.

andyzhangx:
at least `List Volumes in Cluster` works now, while etcd disk is still in the result, for every disk, it should exist in cluster disk PV
list, otherwise should filter out:

```
I0218 08:43:30.711123       1 utils.go:95] GRPC call: /csi.v1.Controller/ListVolumes
I0218 08:43:30.711155       1 utils.go:96] GRPC request: {}
I0218 08:43:30.711212       1 controllerserver.go:529] List Volumes in Cluster:
I0218 08:43:30.716771       1 round_trippers.go:445] GET https://10.0.0.1:443/api/v1/persistentvolumes 200 OK in 5 milliseconds
I0218 08:43:30.749000       1 utils.go:102] GRPC response: {"entries":[{"status":{"published_node_ids":["k8s-master-25577820-0"]},"volume":{"volume_id":"/subscriptions/0e46bd28-a80f-4d3a-8200-d9eb8d80cb2e/resourceGroups/kubetest-c8dc526e-71c3-11eb-9324-be3a24f54b97/providers/Microsoft.Compute/disks/k8s-master-25577820-0-etcddisk"}}]}
```

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_azuredisk-csi-driver/725/pull-azuredisk-csi-driver-e2e-windows/1362318569785790464/build-log.txt

pkg/azuredisk/controllerserver.go
sunpa93: It should be `start + numVisited`. Thanks for pointing this out :)

pkg/azuredisk/controllerserver.go
sunpa93: You are absolutely right. Start token should only be considered until the local iteration of `listVolumeByResourceGroup` finds a starting point. I am really glad that you found this.

sunpa93:
> at least `List Volumes in Cluster` works now, while etcd disk is still in the result, for every disk, it should exist in cluster disk PV
> list, otherwise should filter out:
> 
> ```
> I0218 08:43:30.711123       1 utils.go:95] GRPC call: /csi.v1.Controller/ListVolumes
> I0218 08:43:30.711155       1 utils.go:96] GRPC request: {}
> I0218 08:43:30.711212       1 controllerserver.go:529] List Volumes in Cluster:
> I0218 08:43:30.716771       1 round_trippers.go:445] GET https://10.0.0.1:443/api/v1/persistentvolumes 200 OK in 5 milliseconds
> I0218 08:43:30.749000       1 utils.go:102] GRPC response: {"entries":[{"status":{"published_node_ids":["k8s-master-25577820-0"]},"volume":{"volume_id":"/subscriptions/0e46bd28-a80f-4d3a-8200-d9eb8d80cb2e/resourceGroups/kubetest-c8dc526e-71c3-11eb-9324-be3a24f54b97/providers/Microsoft.Compute/disks/k8s-master-25577820-0-etcddisk"}}]}
> ```
> 
> https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_azuredisk-csi-driver/725/pull-azuredisk-csi-driver-e2e-windows/1362318569785790464/build-log.txt

It already filters out any volume that was not found in the pv list (https://github.com/sunpa93/azuredisk-csi-driver/blob/listvolume-fix/pkg/azuredisk/controllerserver.go#L661-L664). Should we try to manually filter out etcd disks either by string comparison or checking disk properties?

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +525,170 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+	listVolumesResp := &csi.ListVolumesResponse{
+		Entries:   listStatus.entries,
+		NextToken: nextTokenString,
+	}
+
+	return listVolumesResp, nil
+}
+
+// listVolumesByResourceGroup is a helper function that updates the ListVolumeResponse_Entry slice and returns number of total visited volumes, number of volumes that needs to be visited and an error if found
+func (d *Driver) listVolumesByResourceGroup(ctx context.Context, resourceGroup string, entries []*csi.ListVolumesResponse_Entry, start, maxEntries int, volSet *map[string]bool) listVolumeStatus {
```
andyzhangx, line 637: add comments for func parameters, especially `volSet`, and why use `*map[string]bool` (pointer), map in golang is already pass by reference

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +525,170 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+			if err := isValidDiskURI(diskURI); err != nil {
+				klog.Warningf("invalid disk uri (%s) with error(%v)", diskURI, err)
+				continue
+			}
+			rg, err := GetResourceGroupFromURI(diskURI)
+			if err != nil {
+				klog.Warningf("failed to get resource group from disk uri (%s) with error(%v)", diskURI, err)
+				continue
+			}
+			volSet[diskURI] = true
```
andyzhangx, line 558: should convert diskURI to lower case, otherwise there could be duplicated disk URIs

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +525,170 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+			if _, visited := rgMap[rg]; visited {
```
andyzhangx, line 559: also convert rg to lower case, otherwise there are duplicates

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +525,170 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
 			return nil, status.Errorf(codes.Aborted, "ListVolumes starting token(%d) can not be negative", start)
 		}
 	}
+	if d.cloud.KubeClient != nil && d.cloud.KubeClient.CoreV1() != nil && d.cloud.KubeClient.CoreV1().PersistentVolumes() != nil {
+		klog.V(6).Infof("List Volumes in Cluster:")
+		return d.listVolumesInCluster(ctx, start, int(req.MaxEntries))
+	}
+	klog.V(6).Infof("List Volumes in Node Resource Group:")
```
andyzhangx, line 532: `List Volumes in Node Resource Group: %s`, ...

pkg/azuredisk/controllerserver.go
```diff
@@ -73,6 +74,13 @@ const (
 	sourceDiskSearchMaxDepth = 10
 )
 
+type listVolumeStatus struct {
+	numVisited    int
```
andyzhangx, line 78: add comments for the fields

pkg/azuredisk/controllerserver.go
andyzhangx: I see, `nil` means it won't filter. In this case, if `volSet` is not nil and is empty, then it should return directly, not necessary to run `DisksClient.ListByResourceGroup`

pkg/azuredisk/controllerserver.go
```diff
@@ -517,60 +525,170 @@ func (d *Driver) ListVolumes(ctx context.Context, req *csi.ListVolumesRequest) (
...
+			volSet[diskURI] = true
+			if _, visited := rgMap[rg]; visited {
+				continue
+			}
+			rgMap[rg] = true
+		}
 	}
 
-	if start != 0 && start >= len(disks) {
+	if start != 0 && start > len(volSet) {
```
sunpa93, line 566: start token refers to an index in a list of all azure disks, including those that may not be csi volumes. This condition compares the start token index to the number of azure csi volumes; therefore it's an invalid comparison.

pkg/azuredisk/controllerserver.go
sunpa93: I'm sorry but I am not following this. Could you specify?

pkg/azuredisk/controllerserver.go
sunpa93: I'm sorry but I am not following this. Could you specify?

andyzhangx: there could be two `diskURI`, which has same rg, while with different case, r.g. `testrg1`, `TESTRG2`

pkg/azuredisk/controllerserver.go
sunpa93: With the current design, `DisksClient.ListByResourceGroup` would need to be called nonetheless to update numVisited because start token refers to an index in a list of all azure disks, not azure csi volumes. But if volSet is not nil and is empty, we should certainly skip the unnecessary loop and return `listVolumeStatus` with updated `numVisited` value.

sunpa93: If the start token refers to an index in a list of azure csi disk volumes, we would always have to loop from the start of the disk list returned by `DisksClient.ListByResourceGroup` in `Driver.listVolumesByResourceGroup` in order to find where to actually start adding volume entries. And I am afraid that it would make this process too expensive.

Imagine if there were 1000 azure csi disk volumes and our start token is 900. We have one resource group `rg-1`, and `DisksClient.ListByResourceGroup(ctx, "rg-1")` returns 3000 disks, among which there are 1000 csi volumes. We would then need to iterate from the very start of the list and at worst check 2900 entries to determine where our 900th csi volume is.

On the other hand, if the start token refers to an index in a list of all azure disks, we can start iterating from where the start token points to. Of course, this means that we won't be able to return an error for start token that is too large until we loop through all disks from all resource groups, but the frequency of this happening compared to the abovementioned situation seems to be much lower.

pkg/azuredisk/controllerserver.go
sunpa93: ah, I see what you mean. Thanks for bringing that up 👍

andyzhangx (approved):
/lgtm
/approve

pkg/azuredisk/mockkubeclient/interface.go
```diff
@@ -0,0 +1,706 @@
...
+	v11 "k8s.io/client-go/kubernetes/typed/authentication/v1"
+	v1beta11 "k8s.io/client-go/kubernetes/typed/authentication/v1beta1"
+	v12 "k8s.io/client-go/kubernetes/typed/authorization/v1"
+	v1beta12 "k8s.io/client-go/kubernetes/typed/authorization/v1beta1"
+	v13 "k8s.io/client-go/kubernetes/typed/autoscaling/v1"
+	v2beta1 "k8s.io/client-go/kubernetes/typed/autoscaling/v2beta1"
+	v2beta2 "k8s.io/client-go/kubernetes/typed/autoscaling/v2beta2"
+	v14 "k8s.io/client-go/kubernetes/typed/batch/v1"
+	v1beta13 "k8s.io/client-go/kubernetes/typed/batch/v1beta1"
+	v2alpha1 "k8s.io/client-go/kubernetes/typed/batch/v2alpha1"
```
andyzhangx, line 39: hi @sunpa93 where did you copy these code from? We have issue upgrading to k8s v0.21.0 lib

andyzhangx: nvm, it's fixed

PR #727, opened 2021-02-09
chore: Add side-by-side implementation of v2 driver to ease integration with ongoing v1 work

edreed:
**What type of PR is this?**

/kind cleanup

This PR adds a side-by-side implementation of the v2 driver that is conditionally compiled (through the "azurediskv2" tag) in order to ease integration with ongoing v1 work with v2 work.

Fixes #

**Requirements**:
- [X] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [X] adds unit tests

@andyzhangx This is the PR we discussed offline for staging v2 development. The v2 implementation is conditionally compiled through the "azurediskv2" tag. There is also a "--temp-use-driver-v2" command-line switch to enable side-by-side testing in the same binary. In this PR, the v1 and v2 implementations are the same (xxx_v2.go files are copies of their original counterparts). I would appreciate your feedback on this approach and your recommendations for integrating v2 testing into the integration and e2e testing pipelines.

**Release note**:
```
none
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,16 +72,31 @@ verify: unit-test
 	go vet ./pkg/...
 
 .PHONY: unit-test
-unit-test:
+unit-test: unit-test-v1 unit-test-v2
+
+.PHONY: unit-test-v1
+unit-test-v1:
 	go test -v -cover ./pkg/... ./test/utils/credentials
+
+.PHONY: unit-test-v2
+unit-test-v2:
+	go test -v -cover -tags azurediskv2 ./pkg/azuredisk --temp-use-driver-v2
 
 .PHONY: sanity-test
 sanity-test: azuredisk
 	go test -v -timeout=30m ./test/sanity
 
+.PHONY: sanity-test-v2
+sanity-test-v2: azuredisk-v2
+	go test -v -timeout=30m ./test/sanity --temp-use-driver-v2
+
 .PHONY: integration-test
 integration-test: azuredisk
 	go test -v -timeout=30m ./test/integration
+
+.PHONY: integration-test-v2
+integration-test-v2: azuredisk-v2
+	go test -v -timeout=30m ./test/integration --temp-use-driver-v2
 
 .PHONY: e2e-test
 e2e-test:
@@ -118,6 +133,10 @@ e2e-teardown:
 .PHONY: azuredisk
 azuredisk:
 	CGO_ENABLED=0 GOOS=linux go build -a -ldflags ${LDFLAGS} -mod vendor -o _output/azurediskplugin ./pkg/azurediskplugin
+
+.PHONY: azuredisk-v2
+azuredisk-v2:
+	CGO_ENABLED=0 GOOS=linux go build -a -ldflags ${LDFLAGS} -mod vendor -tags azurediskv2 -o _output/azurediskpluginv2 ./pkg/azurediskplugin
 
 .PHONY: azuredisk-windows
 azuredisk-windows:
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -38,6 +38,7 @@ import (
 	csicommon "sigs.k8s.io/azuredisk-csi-driver/pkg/csi-common"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/mounter"
 	volumehelper "sigs.k8s.io/azuredisk-csi-driver/pkg/util"
+	"sigs.k8s.io/cloud-provider-azure/pkg/provider"
 	azure "sigs.k8s.io/cloud-provider-azure/pkg/provider"
 )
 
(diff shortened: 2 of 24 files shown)


andyzhangx (commented):
/ok-to-test

pkg/azuredisk/azuredisk.go
```diff
@@ -323,3 +338,48 @@ func IsCorruptedDir(dir string) bool {
 	fmt.Printf("IsCorruptedDir(%s) returned with error: %v", dir, pathErr)
 	return pathErr != nil && mount.IsCorruptedMnt(pathErr)
 }
+
+// setControllerCapabilities sets the controller capabilities field. It is intended for use with unit tests.
+func (d *DriverCore) setControllerCapabilities(caps []*csi.ControllerServiceCapability) {
```
andyzhangx, line 343: `setXXX` func is actually refactor, can we remove such code, by that way, the v2 implementation would be much simpler to review. And I think current operator assign is ok inside this driver, if there is requirement to set ControllerCapabilities outside of this package, then we need to write `SetControllerCapabilities` func. 
`setControllerCapabilities` func is only for internal use inside current package.

edreed: The azuredisk unit tests currently have cases that call into internal methods. In order to allow the same unit tests to verify both the v1 & v2 driver, I needed to abstract them behind an interface which meant that I needed to add these accessor functions. (See FakeDriver in fake_azuredisk.go) Unfortunately as you note, this lead to more churn in this change than would otherwise be desirable. The churn in the unit test files should be limited to replacement of direct access to member with the use of these accessors.

Makefile
```diff
@@ -75,6 +75,10 @@ verify: unit-test
 unit-test:
 	go test -v -cover ./pkg/... ./test/utils/credentials
 
+.PHONY: unit-test-v2
```
nearora-msft, line 78: Are we doing this separation only for the unit tests? Or is there a plan to pass this optional parameter so that the build and make container commands also use the v2 implementation instead of v1?

andyzhangx: should combine with original `unit-test`, otherwise we are not running ut on v2 code

andyzhangx: I think we need to also add this flag to other tests, otherwise it would lose test coverage on all v2 code

edreed: I agree that the unit-tests should be combined.

For the other tests (e2e, in particular), would we need to build separate container images with v2 in the interim? Or should I just simply add duplicate targets for the v2 images? If the latter, how should we handle the image version, e.g. add a `-v2` suffix to IMAGE_VERSION or use a separate image version like `vNext`? I assume Helm charts should be added to a directory with the same name as IMAGE_VERSION to enable deployment of the `v2` binaries.

andyzhangx: one test pipeline could not test v1 & v2, need to set up another test pipeline(e.g. pull-azuredisk-csi-driver-e2e-single-az-v2). I think in the beginning, could set up sanity and integration tests against v2: https://github.com/kubernetes-sigs/azuredisk-csi-driver/tree/master/test/sanity

edreed: I've push changes to enable sanity and integration testing of the v2 driver.

edreed:
I rebased to pick up the fix for the Windows unit tests, and all unit tests are now passing.

Makefile
andyzhangx: can we enable v2 tests after running v1 tests  in sanity and integration testing, seems it's still only running v1 by default now. For e2e tests, need a new test pipeline in test-infra.

Also could you remove the Draft tag, then it will run sanity and integration testing

edreed: Sanity and integration tests are now run against both v1 & v2 by default.

test/sanity/run-test.sh
```diff
@@ -31,7 +31,11 @@ if [[ "$#" -gt 0 ]] && [[ -n "$1" ]]; then
   nodeid="$1"
 fi
 
-_output/azurediskplugin --endpoint "$endpoint" --nodeid "$nodeid" -v=5 &
+if [[ "$#" -lt 2 || "$2" != "v2" ]]; then
```
andyzhangx, line 34: should add a new `pull-azuredisk-csi-driver-sanity-v2` test pipeline in test-infra first, btw, I doubt whether v2 driver would run sanity test in the end(it could now since code is the same as v1) since v2 driver depends on CRD, while sanity test does not run inside a k8s cluster.

test/integration/run-tests-all-clouds.sh
```diff
@@ -24,4 +24,9 @@ if [[ "$#" -gt 0 ]]; then
   cloud="$1"
 fi
 
-test/integration/run-test.sh 'tcp://127.0.0.1:10000' "$nodeid" "$cloud"
+version='v1'
+if [[ "$#" -gt 1 ]]; then
```
andyzhangx, line 28: should add a new `pull-azuredisk-csi-driver-integration-v2` test pipeline in test-infra first

andyzhangx: I think @mukhoakash would be familiar with this. Could you help? Thanks.

test/sanity/run-test.sh
andyzhangx: cc @mukhoakash

edreed: I separated the v1 & v2 sanity & integration test targets again. I submitted [kubernetes/test-infra PR#20981](https://github.com/kubernetes/test-infra/pull/20981) to add the new pipelines.

@andyzhangx Agreed that we'll have challenges sanity testing the v2 driver with the CRDs. We'll need to give some thought around how we can do that. I'd really hate to lose the coverage.

Makefile
edreed: I separated the sanity & integration targets again and instead added separate pipelines for the v2 tests. See the [discussion](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/727#discussion_r581476359) below.

andyzhangx (approved):
/lgtm
/approve

PR #792, opened 2021-04-14
test: add test cases to increase coverage for util

### LINKED ISSUE #719: improve ut coverage for util.go

The line in red is missing ut test coverage:
https://coveralls.io/builds/36878386/source?filename=pkg%2Futil%2Futil.go

**Additional context**

sunpa93:
**What type of PR is this?**
/kind test

Fixes #719 

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -141,7 +141,8 @@ func TestMakeDir(t *testing.T) {
 		{
 			desc: "should not return error if path already exists",
 			setup: func() {
-				_ = os.Mkdir("./target_test", os.FileMode(0755))
+				err := os.Mkdir("./target_test", os.FileMode(0755))
+				assert.NoError(t, err, "Failed in setup: %v", err)
 			},
 			targetDir:     "./target_test",
 			expectedError: false,
@@ -149,12 +150,14 @@ func TestMakeDir(t *testing.T) {
 		{
 			desc: "[Error] existing file in target path",
 			setup: func() {
-				_, _ = os.Create("file_exists")
+				_, err := os.Create("file_exists")
+				assert.NoError(t, err, "Failed in setup: %v", err)
 			},
 			targetDir:     "./file_exists",
 			expectedError: true,
 			cleanup: func() {
-				_ = os.Remove("./file_exists")
+				err := os.Remove("./file_exists")
+				assert.NoError(t, err, "Failed in cleanup: %v", err)
 			},
 		},
 	}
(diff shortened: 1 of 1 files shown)


andyzhangx:
```
run golangci-lint
  Running [/home/runner/golangci-lint-1.29.0-linux-amd64/golangci-lint run --out-format=github-actions -E=gofmt --timeout=30m0s -E=deadcode -E=unused -E=varcheck -E=ineffassign] in [] ...
  Error: Error return value of `os.Mkdir` is not checked (errcheck)
  
  Error: issues found
```

pkg/util/util_test.go
```diff
@@ -125,23 +126,143 @@ func TestConvertTagsToMap(t *testing.T) {
...
+				err := os.Mkdir("./target_test", os.FileMode(0755))
+				assert.NoError(t, err, "Failed in setup: %v", err)
+			},
+			targetDir:     "./target_test",
+			expectedError: false,
+		},
+		{
+			desc: "[Error] existing file in target path",
+			setup: func() {
+				_, err := os.Create("./file_exists")
```
edreed, line 153: You'll need to call `Close()` on the returned `File` object in order to be able to delete it on line 159 below. I believe this is the reason the Windows unit tests are failing to delete the file with an "in-use" error.

edreed (changes requested):

pkg/util/util_test.go
sunpa93: Ah! Thank you for pointing this out. I forgot about closing the file pointer 👍

pkg/util/util_test.go
```diff
@@ -125,23 +126,144 @@ func TestConvertTagsToMap(t *testing.T) {
...
+		},
+		{
+			desc: "should not return error if path already exists",
+			setup: func() {
+				err := os.Mkdir("./target_test", os.FileMode(0755))
+				assert.NoError(t, err, "Failed in setup: %v", err)
+			},
+			targetDir:     "./target_test",
+			expectedError: false,
+		},
```
edreed, line 149: You should remove the directory after the test case has run just like in `TestMakeFile`.

edreed (changes requested):
Noticed one more thing.

pkg/util/util_test.go
sunpa93: The removal should be handled by the following block in this case:

https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/743802637524d7759b70c08979211a12b7c0bb37/pkg/util/util_test.go#L173-L177

edreed: Oh missed that. Thanks for pointing it out!

edreed (approved):

andyzhangx (approved):
/approve

PR #810, opened 2021-04-28
fail external-e2e prow job on test failure

boddumanohar:
**What type of PR is this?**
fail external-e2e prow job on test failure

/kind test

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/test/external-e2e/manifest/testdriver.yaml
+++ b/test/external-e2e/manifest/testdriver.yaml
@@ -11,7 +11,7 @@ DriverInfo:
         block: true
         fsGroup: true
         topology: true
-        controllerExpansion: true
+        controllerExpansion: false
         nodeExpansion: false
         volumeLimits: true
         snapshotDataSource: false
--- a/test/external-e2e/run.sh
+++ b/test/external-e2e/run.sh
@@ -16,6 +16,8 @@
 
 # Ensure that we have the desired version of the ginkgo test runner.
 
+set -xe
+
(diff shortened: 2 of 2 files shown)


test/external-e2e/run.sh
```diff
@@ -16,6 +16,8 @@
 
 # Ensure that we have the desired version of the ginkgo test runner.
 
+set -xe
```
andyzhangx, line 19: I think only this line is needed.

boddumanohar: I think along with this, we should trap the function `print_logs` because without `TRAP` the script will stop abruptly without printing the logs.

boddumanohar: I have updated the PR, to remove the `print_logs` and `trap`. I think now we can verify if this is working as expected. thanks

andyzhangx (commented):
/ok-to-test

test/external-e2e/manifest/testdriver.yaml
```diff
@@ -14,4 +14,4 @@ DriverInfo:
         controllerExpansion: true
```
andyzhangx, line 14: I see it failed as expected, pls set `controllerExpansion: false`, then test would pass with this PR.

andyzhangx (approved):
/lgtm
/approve

PR #812, opened 2021-04-28
chore: Fix arm64 container build

edreed:
**What type of PR is this?**

/kind cleanup

Fixes the Makefile `container-all` target to build the correct architecture binary and splits outputs into architecture-specific directories to ensure the correct binary is used when building the container image.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

The initial support for the arm64 build through `container-all` appeared to work but actually picked up the amd64 build because the incorrect `ARCH=$${arch} $(MAKE) azurefile` command would fail to re-build the binary for each architecture. The amd64 build was still in the dependency list for the `container-all` target, and subsequent container build commands would simply use the previously built amd64 binary.

**Release note**:
```
none
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ e2e-teardown:
 
 .PHONY: azuredisk
 azuredisk:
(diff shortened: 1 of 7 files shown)


andyzhangx (approved):
/lgtm
/approve

PR #824, opened 2021-05-03
Adjust PVC capacity to mirror in-tree behavior

huffmanca:
**What type of PR is this?**
/kind cleanup

Currently, if you request a 1MiB volume using the in-tree driver, then the corresponding created PVC has a capacity of 1 GiB. We currently create a backend disk that's rounded up to the nearest GiB, but the PVC uses the capacity initially specified, which is different from the in-tree driver.

This PR updates the CSi driver to 

Fixes #

**Requirements**:

**Release note**:
```
Updates the CreateVolumeResponse to use a rounded GiB of the specified request.
```


andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm
/approve
/retest

PR #843, opened 2021-05-13
feat: Update device settings to drive better performance

abhisheksinghbaghel:
In this PR:
1. Add ability for customer to enable perf optimization in storageclass.
2. Ability to select perf profile in storageclass.
3. Added scripts to run perf tests against the driver.
4. Added scripts to get sku\latency for azure disk.
5. Added unit tests.

**What type of PR is this?**

/kind feature

This PR will allow users to drive better performance to their data disks by flipping a perf optimization feature in storage class.

Fixes #

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


andyzhangx (commented):
/ok-to-test

andyzhangx:
some comments:
 - do you have data or reference about what's the perf improvement, rw iops, latency, etc.
 -  is this change specific for Ubuntu? note that we should also support RedHat
 - don't need flag in driver to enable/disable this feature, could only be one setting in storage class, that would be easier to use

abhisheksinghbaghel:
> some comments:
> 
> * do you have data or reference about what's the perf improvement, rw iops, latency, etc.
> * is this change specific for Ubuntu? note that we should also support RedHat
> * don't need flag in driver to enable/disable this feature, could only be one setting in storage class, that would be easier to use

* Forwarded you some emails with details on issues and improvements we see.
* This change itself will apply to all disks which use the storVsc driver on azure which covers HDDs and Premium SKUs today. In general I am open for broader support.
* There is some considerable amount of Init code that runs for getting SKUs etc and I plan to change that in future to become more dynamic, this is kept behind the flag. The flag is enabled by default so customers don't need to toggle it. It just gives us more flexibility for future, if something goes bad in the Init code path. We will always be able to mitigate issues without a code change.

pkg/azuredisk/device_perf_optimization_helper.go
```diff
@@ -0,0 +1,114 @@
...
+			diskIopsStr = v
+		case diskMBPSReadWriteField:
+			diskBwMbpsStr = v
+		default:
+			continue
+		}
+	}
+
+	// If perf tuning mode is not valid, set it to none
+	if !IsValidPerfTuningMode(mode) {
```
andyzhangx, line 92: should fail if it's not valid mode in `CreateVolume` validation. Silent warning is not good.

pkg/azuredisk/device_perf_optimization_helper.go
```diff
@@ -0,0 +1,114 @@
...
+		klog.Warningf("Invalid perf tuning mode %s. Defaulting to: %s", mode, PerfTuningModeNone)
+		mode = PerfTuningModeNone
+	}
+
+	// If profile is not valid, set it to default
+	if !IsValidPerfProfile(mode, profile) {
+		klog.Warningf("Invalid perf profile %s for mode %s. Defaulting to: %s", profile, mode, PerfProfileDefault)
```
andyzhangx, line 99: should fail if it's not valid mode in `CreateVolume` validation. Silent warning is not good.

pkg/azuredisk/device_perf_optimization_helper.go
```diff
@@ -0,0 +1,114 @@
...
+		profile = PerfProfileDefault
+	}
+
+	return mode, profile, accountType, diskSizeGibStr, diskIopsStr, diskBwMbpsStr
+}
+
+func AccountSupportsPerfOptimization(accountType string) bool {
+
```
andyzhangx, line 107: remove empty line

pkg/azuredisk/device_perf_optimization_helper.go
```diff
@@ -0,0 +1,114 @@
...
+	case PerfTuningModeAuto:
+		return true
+	default:
+		return false
+	}
+}
+
+// GetPerfTuningModeAndProfile gets the per tuning mode and profile set in attributes
+func GetDiskPerfAttributes(attributes map[string]string) (mode string, profile string, accountType string, diskSizeGibStr string, diskIopsStr string, diskBwMbpsStr string) {
+
```
andyzhangx, line 71: remove empty line

hack/get-disk-skus.go
```diff
@@ -0,0 +1,118 @@
...
+	azuredisk "sigs.k8s.io/azuredisk-csi-driver/pkg/azuredisk"
+)
+
+func init() {
+	klog.InitFlags(nil)
+}
+
+func main() {
+
+	skusFullfilePath := "skus-full.json"
```
andyzhangx, line 39: is it possible to define `diskSkus map[string]map[string]DiskSkuInfo` in go file directly, may reuse this map: https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/pkg/azuredisk/azure_dd_max_disk_count.go, use `populateSkuMap` to parse from one file is error prone, if it's a built-in map, then would be quite straightforward.

pkg/azuredisk/device_perf_optimization_linux.go
```diff
@@ -0,0 +1,296 @@
...
+
+type DeviceHelper struct{}
+
+func (deviceHelper *DeviceHelper) OptimizeDevicePerformance(nodeInfo *NodeInfo, diskSkus map[string]map[string]DiskSkuInfo, devicePath string, attributes map[string]string) (err error) {
+
+	klog.V(2).Infof("OptimizeDevicePerformance: Tuning settings for %s", devicePath)
+
+	if nodeInfo == nil {
+		err = fmt.Errorf("Node info is not provided. Error: invalid parameter")
+		klog.Errorf("Error: %s", err)
```
andyzhangx, line 41: no need to log error everywhere, just `return err` and could log error in outside function, then only need to write one logging which would make code clean

pkg/azuredisk/nodeserver.go
```diff
@@ -81,17 +81,33 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
...
+	}
+
+	// If perf optimizations are enabled
+	// tweak device settings to enhance performance
+	if d.getPerfOptimizationEnabled() {
+		err = d.getDeviceHelper().OptimizeDevicePerformance(d.getNodeInfo(), d.getDiskSkuInfoMap(), source, req.GetVolumeContext())
+		// In the beginning, do not fail because of perf optimization related error
+		// TODO: revisit this decision after getting some test runs
+		if err != nil {
+			klog.Errorf("NodeStageVolume: Failed to optimize device performance for target(%s) error(%s)", source, err)
```
andyzhangx, line 101: it's better return error directly, silent warning could be ignored quite easily.

deploy/csi-azuredisk-node.yaml
```diff
@@ -83,6 +83,7 @@ spec:
             - "--endpoint=$(CSI_ENDPOINT)"
             - "--nodeid=$(KUBE_NODE_NAME)"
             - "--metrics-address=0.0.0.0:29605"
+            - "--enable-perf-optimization=true"
```
andyzhangx, line 86: managed Azure Disk CSI driver on AKS does not support such customization for different clusters, so it's better put all sku configs as a map in go file, by that way, don't need to read & parse a config map, and we could leave this option to user setting (`enablePerfOptimization`) in storage class.

docs/enhancements/feat-add-ability-to-tune-azuredisk-performance-parameters.md
```diff
@@ -0,0 +1,93 @@
...
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: sc-kubestone-perf-optimized-premium-ssd-csi
+provisioner: disk.csi.azure.com
+parameters:
+  skuname: Premium_LRS
+  perftuningmode: Auto
```
andyzhangx, line 68: use camel case like `perfTuningMode`

docs/enhancements/feat-add-ability-to-tune-azuredisk-performance-parameters.md
```diff
@@ -0,0 +1,93 @@
...
+  perfprofile: Default
```
andyzhangx, line 69: why not combine these two parameters, e.g. only have `perfTuningMode`, now it has two available values: ``(empty means none, by default), `auto`, it could save one parameter!

deploy/csi-azuredisk-node.yaml
abhisheksinghbaghel: You need to take this as a feature flag. We are not asking users to tweak this setting on day to day operations, this is our may day option, so I don't see any harm in keeping this. driver is used outside of AKS, where users will be able to mitigate if it is needed.

hack/get-disk-skus.go
abhisheksinghbaghel: It's a well formed json, we are not parsing a flat file or random string. I don't think it's a good idea to hardcode an ever changing configuration. A new sku would mean we have to make a code change.
As mentioned in the other file, I plan to change this to take a dependency on the skus API.

I'll hardcode it for now, to simply deployments, since I plan to change this anyway.

docs/enhancements/feat-add-ability-to-tune-azuredisk-performance-parameters.md
abhisheksinghbaghel: @edreed @andyzhangx has an interesting idea. I know we have discussed this in the past. We can just have perf tuning mode, which can be a manual or auto mode. We will validate this in code. I think from what I was planning I can't think of a scenario where it would not work. Let me know what you think?

edreed: I agree that these can be combined into a single `perfTuningMode`. I would recommend supporting the values `none` and `auto`. The default would be `none` and implied if the parameter was not present. We could add additional properties in the future to support manual or profile based tuning.

abhisheksinghbaghel: @edreed I was more inclining towards providing the name of the profiles directly and that profile can be an inbuilt one or something provided by customer using a CRD. absence would mean no profile provided, which is same as tuning not enabled. so the current implementation will change to:
* perfprofile not present: tuning disabled
* perfprofile present but is empty or not known (today only default will be known): will result in failure.
* perfprofile present and known (today only default will be known): we will apply the profile. known profile will be combination of inbuilt profiles and customer provided profiles.

What do you think?

edreed: @abhisheksinghbaghel Yes, that would work and be extensible without additional properties in the future. I believe that we should also allow for a `none` profile to be specified in addition to omitting the parameter to allow the config to be explicit in perf tuning choice.

abhisheksinghbaghel: @edreed Thanks. I can make that happen!

pkg/azuredisk/device_perf_optimization_linux.go
abhisheksinghbaghel: Changing this behavior for places where I am returning custom errors. If I don't log the errors returned by routines then it would be hard to debug the issues. given we will not have the line no. etc in the calling function.

andyzhangx: `return fmt.Errorf("Node info should not be nil")` is ok, it already tells complete info, don't need log where error happens in the code

andyzhangx:
it returns error in e2e test, slient warning, I would suggest enable this feature in e2e test, if performance tuning failed, e2e test should fail, otherwise we don't know whether this feature has been tested in e2e test or not:
```
I0520 11:57:08.451220       1 device_perf_optimization_linux.go:37] OptimizeDevicePerformance: Tuning settings for /dev/disk/azure/scsi1/lun0
E0520 11:57:08.451232       1 device_perf_optimization_linux.go:41] Error: Node info is not provided. Error: invalid parameter
E0520 11:57:08.451242       1 nodeserver.go:101] NodeStageVolume: Failed to optimize device performance for target(/dev/disk/azure/scsi1/lun0) error(Node info is not provided. Error: invalid parameter)
```

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_azuredisk-csi-driver/843/pull-azuredisk-csi-driver-e2e-single-az/1395332748620599296/build-log.txt

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -95,6 +97,15 @@ func (d *DriverV2) Run(endpoint, kubeconfig string, disableAVSetNodes, testingMo
 		}
 	}
 
+	d.deviceHelper = NewSafeDeviceHelper()
+
+	if d.getPerfOptimizationEnabled() {
+		err = PopulateNodeAndSkuInfo(d.DriverCore)
```
andyzhangx, line 103:
```
if err :=PopulateNodeAndSkuInfo(d.DriverCore); err != nil {

}
```

pkg/azuredisk/controllerserver.go
```diff
@@ -24,7 +24,7 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-12-01/compute"
+	compute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-12-01/compute"
```
andyzhangx, line 27: not necessary change

docs/enhancements/feat-add-ability-to-tune-azuredisk-performance-parameters.md
andyzhangx: so what's the conclusion? use two parameters? and what's the default value for this driver? enable or disable?

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -95,6 +97,15 @@ func (d *DriverV2) Run(endpoint, kubeconfig string, disableAVSetNodes, testingMo
 		}
 	}
 
+	d.deviceHelper = NewSafeDeviceHelper()
+
+	if d.getPerfOptimizationEnabled() {
```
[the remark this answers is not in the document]
abhisheksinghbaghel: We are only enabling this for linux today. OptimizeDevicePerformance is a no op for unsupported OS.

pkg/azuredisk/azuredisk.go
```diff
@@ -142,7 +152,9 @@ func newDriverV1(nodeID string) *Driver {
 	if err != nil {
 		klog.Fatalf("%v", err)
 	}
-	driver.getDiskThrottlingCache = cache
+	driver.diskThrottlingCache = cache
```
andyzhangx, line 155: `getDiskThrottlingCache` is for handling `GetDisk` operation throttling, don't make other changes in this PR, focus on this  feature with  this PR.

pkg/azuredisk/device_perf_optimization_helper.go
```diff
@@ -0,0 +1,114 @@
...
+	PerfTuningModeAuto       = "auto"
+	PerfProfileDefault       = "default"
+	PremiumAccountPrefix     = "premium"
+	StandardSsdAccountPrefix = "standardssd"
+)
+
+// IsValidPerfTuningMode Checks to see if perf tuning mode passed is correct
+// Right now we only support auto perf tuning mode
+// Manual mode to come later
+func IsValidPerfTuningMode(mode string) bool {
```
andyzhangx, line 36: for these non public exposure functions, naming as `isValidPerfTuningMode` is ok, no need to expose as public function for every function, and add comment for every function

pkg/azuredisk/device_perf_optimization_helper.go
```diff
@@ -0,0 +1,114 @@
...
+	default:
+		return false
+	}
+}
+
+// IsValidPerfProfile Checks to see if perf profile passed is correct
+// Right now we are only supporing default profile
+// Other advanced profiles to come later
+func IsValidPerfProfile(mode string, profile string) bool {
+
```
andyzhangx, line 50: remove empty line for every starting line for functions

pkg/azuredisk/device_perf_optimization_linux.go
```diff
@@ -0,0 +1,296 @@
...
+		profile,
+		queueDepth,
+		nrRequests,
+		scheduler,
+		maxSectorsKb,
+		readAheadKb)
+
+	err = echoToFile(maxSectorsKb, filepath.Join("/sys/block", deviceName, "queue/max_sectors_kb"))
+	if err != nil {
+		klog.Errorf("OptimizeDevicePerformance: Could not set max_sectors_kb for device %s. Error: %v", deviceName, err)
```
andyzhangx, line 88: no need to log every errors in the function, just log outsider

pkg/azuredisk/device_perf_optimization_linux.go
```diff
@@ -0,0 +1,296 @@
...
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"k8s.io/klog/v2"
+)
+
+type DeviceHelper struct{}
+
+func (deviceHelper *DeviceHelper) OptimizeDevicePerformance(nodeInfo *NodeInfo, diskSkus map[string]map[string]DiskSkuInfo, devicePath string, attributes map[string]string) (err error) {
```
andyzhangx, line 35: `SetOptimizeDevicePerformance`

pkg/azuredisk/device_perf_optimization_linux.go
abhisheksinghbaghel: Don't we lose the error stack in the kind of logging you're suggesting? The regimen you're suggesting would result in one log at a top level function and if the error is generic enough, it would be hard to track the real root cause, line number etc of the failure. It probably makes sense to do it for smaller functions with fewer places throwing errors.

For example in this case If I don't format and log the error thrown by the exact setting, I would not know which setting actually failed. Making it harder for me to debug.

pkg/azuredisk/device_perf_optimization_linux.go
abhisheksinghbaghel: Set and Optimize are both verbs, so I don't think that would be a correct name.

If you insist I can change it to SetAndOptimize*, although I think what I have is an accurate description for the actions taken in the method. Let me know?

pkg/azuredisk/azuredisk.go
abhisheksinghbaghel: I don't intend to make any functional changes. I needed to expose this for test coverage and ran in to this name.
The go guidelines don't recommend getters to have get* naming, let alone for fields.
https://golang.org/doc/effective_go#Getters

I can revert it, if you have any concerns?

andyzhangx: that's for `Get`, not `get`, maybe `diskGetThrottlingCache` is more appropriate, anyway, this is out of this PR scope

andyzhangx (commented):
since this is a fundamental change(impact is big), need to be very cautious.
And can you fix the e2e test first? actually in e2e test logs, it failed saliently

pkg/azuredisk/azuredisk.go
abhisheksinghbaghel: Will revert. Thanks!

pkg/azuredisk/device_perf_optimization_linux.go
```diff
@@ -0,0 +1,296 @@
...
+	cmd := exec.Command("echo", content)
+	outfile, err := os.Create(filePath)
+	if err != nil {
+		return err
+	}
+	defer outfile.Close()
+
+	cmd.Stdout = outfile
+
+	err = cmd.Start()
```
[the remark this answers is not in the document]
abhisheksinghbaghel: echoToFile is similar to what Exec.Command is doing. Difference being we echo to a file vs returning the output.

May be I am not clear with what you mean here. Can you elaborate?

abhisheksinghbaghel:
> since this is a fundamental change(impact is big), need to be very cautious.
> And can you fix the e2e test first? actually in e2e test logs, it failed saliently

Yeah, we will now be failing the create\attach volume so we should be able to catch all the errors. Just iterating through changes to uncover issues. Won't check-in without an e2e test. Thanks for the word of caution!

pkg/azuredisk/device_perf_optimization_linux.go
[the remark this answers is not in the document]
andyzhangx: would you use ioutil package to open & write file?

pkg/azuredisk/device_perf_optimization_linux.go
```diff
@@ -0,0 +1,296 @@
...
+	if diskSkus == nil {
+		err = fmt.Errorf("Disk SKUs are not provided. Error: invalid parameter")
+		klog.Errorf("Error: %s", err)
+		return err
+
+	}
+
+	mode, profile, accountType, diskSizeGibStr, diskIopsStr, diskBwMbpsStr := GetDiskPerfAttributes(attributes)
+
+	if !IsPerfTuningEnabled(mode) {
```
andyzhangx, line 54: move this judgement to outside of `OptimizeDevicePerformance` func
```
if !IsPerfTuningEnabled(mode) {

...
if !AccountSupportsPerfOptimization(accountType) {
...
```

pkg/azuredisk/device_perf_optimization_linux.go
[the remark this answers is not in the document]
abhisheksinghbaghel: No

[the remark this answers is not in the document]
edreed: Why not append to the file directly rather than spawn a process to echo to a redirected stdout?

[Append Example](https://pkg.go.dev/os@go1.16.4#example-OpenFile-Append)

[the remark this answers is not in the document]
abhisheksinghbaghel: Hi @edreed One of these settings file exhibit some special behavior, echo results in toggling of this setting. Notice how we toggle between none and mq-deadline.
root@abhib-dev-ubuntu-18:/home/azureuser/go/src/sigs.k8s.io/azuredisk-csi-driver# echo mq-deadline > /sys/block/sdc/queue/scheduler
root@abhib-dev-ubuntu-18:/home/azureuser/go/src/sigs.k8s.io/azuredisk-csi-driver# cat /sys/block/sdc/queue/scheduler
[mq-deadline] none
root@abhib-dev-ubuntu-18:/home/azureuser/go/src/sigs.k8s.io/azuredisk-csi-driver# echo none > /sys/block/sdc/queue/scheduler
root@abhib-dev-ubuntu-18:/home/azureuser/go/src/sigs.k8s.io/azuredisk-csi-driver# cat /sys/block/sdc/queue/scheduler
[none] mq-deadline 
root@abhib-dev-ubuntu-18:/home/azureuser/go/src/sigs.k8s.io/azuredisk-csi-driver# ls /sys/block/sdc/queue/scheduler
/sys/block/sdc/queue/scheduler
root@abhib-dev-ubuntu-18:/home/azureuser/go/src/sigs.k8s.io/azuredisk-csi-driver# ls -l /sys/block/sdc/queue/scheduler
-rw-r--r-- 1 root root 4096 May 24 20:28 /sys/block/sdc/queue/scheduler

I did try this io* operations but I was overwriting in what I did. I am not sure why the special behavior with this file. But I can always circle back and make use of IO* operations, when I understand this FS behavior better. For now, I've just made all of it consistent with echo calls.

abhisheksinghbaghel:
> it returns error in e2e test, slient warning, I would suggest add one e2e test for this feature, if performance tuning failed, e2e test should fail, otherwise we don't know whether this feature has been tested in e2e test or not:
> 
> ```
> I0520 11:57:08.451220       1 device_perf_optimization_linux.go:37] OptimizeDevicePerformance: Tuning settings for /dev/disk/azure/scsi1/lun0
> E0520 11:57:08.451232       1 device_perf_optimization_linux.go:41] Error: Node info is not provided. Error: invalid parameter
> E0520 11:57:08.451242       1 nodeserver.go:101] NodeStageVolume: Failed to optimize device performance for target(/dev/disk/azure/scsi1/lun0) error(Node info is not provided. Error: invalid parameter)
> ```
> 
> https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_azuredisk-csi-driver/843/pull-azuredisk-csi-driver-e2e-single-az/1395332748620599296/build-log.txt

This is resolved now.

abhisheksinghbaghel:
> > since this is a fundamental change(impact is big), need to be very cautious.
> > And can you fix the e2e test first? actually in e2e test logs, it failed saliently
> 
> Yeah, we will now be failing the create\attach volume so we should be able to catch all the errors. Just iterating through changes to uncover issues. Won't check-in without an e2e test. Thanks for the word of caution!

Added e2e test which optimizes the disk.

pkg/azuredisk/nodeserver_v2.go
```diff
@@ -67,17 +67,41 @@ func (d *DriverV2) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolume
...
+
+	source, err := d.getDevicePathWithLUN(lun)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to find disk on lun %s. %v", lun, err)
+	}
+
+	// If perf optimizations are enabled
+	// tweak device settings to enhance performance
+	if d.getPerfOptimizationEnabled() {
+		profile, accountType, diskSizeGibStr, diskIopsStr, diskBwMbpsStr, err := getDiskPerfAttributes(req.GetVolumeContext())
```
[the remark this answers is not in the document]
andyzhangx: actually `diskIopsStr` is empty in static provisioning, are you going to handle such case?

docs/enhancements/feat-add-ability-to-tune-azuredisk-performance-parameters.md
```diff
@@ -0,0 +1,89 @@
...
+
+```yaml
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: sc-kubestone-perf-optimized-premium-ssd-csi
+provisioner: disk.csi.azure.com
+parameters:
+  skuName: Premium_LRS
+  perfProfile: Default
```
andyzhangx, line 66: add comments: 
`# available values: none(by default), default`

hack/gen-disk-skus-map.go
```diff
@@ -0,0 +1,333 @@
...
+	nodeMapEnd := "	}"
+
+	skusFullfilePath := "skus-full.json"
+	defer os.Remove(skusFullfilePath)
+	skusFilePath := "pkg/azuredisk/azure_skus_map.go"
+
+	skuMap := map[string]bool{}
+
+	var resources []compute.ResourceSku
+	err := getAllSkus(skusFullfilePath)
```
andyzhangx, line 91: it's more clear write like this:
```
if err := getAllSkus(skusFullfilePath); err != nil {
 ...
}
```

hack/gen-disk-skus-map.go
```diff
@@ -0,0 +1,333 @@
+/*
```
andyzhangx, line 1: add build test for this go files? it's better put into pkg folder which has unit test coverage

test/e2e/dynamic_provisioning_test.go
```diff
@@ -140,6 +140,66 @@ func (t *dynamicProvisioningTestSuite) defineTests(isMultiZone bool) {
...
+				}, isMultiZone),
+				IsWindows: isWindowsCluster,
+			},
+		}
+		test := testsuites.DynamicallyProvisionedCmdVolumeTest{
+			CSIDriver: testDriver,
+			Pods:      pods,
+			StorageClassParameters: map[string]string{
+				"skuName":     "Premium_LRS",
+				"perfProfile": "None",
```
andyzhangx, line 197: not necessary to write a `"perfProfile": "None"` since it's disabled by default, just write one e2e test case for `"perfProfile": "Default"` is ok.

docs/enhancements/feat-add-ability-to-tune-azuredisk-performance-parameters.md
andyzhangx: I found it's a little tricky that `none` is by default, while `default` is not by default. maybe `default` --> `auto` is more appropiate.

pkg/azuredisk/nodeserver_v2.go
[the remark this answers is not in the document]
abhisheksinghbaghel: changed the profile propagation to be a string vs *string.

diskIopsStr is optional, the case you're calling out is handled and unit tested. In absence of diskIopsStr in the volume attribute, we optimize the volume for max disk (managed disk limit) iops.

Only thing that is absolutely required is diskSizeGibStr. I will cut that dependency as well, once I am able to expose the disk properties and sku APIs via the cloud provider. That would cut this sku* map dependency as well.

docs/enhancements/feat-add-ability-to-tune-azuredisk-performance-parameters.md
abhisheksinghbaghel: This is a very Good point, I think we are here due to incremental changes from previous revisions. I am changing "default" to "basic" and will keep the none.

test/e2e/dynamic_provisioning_test.go
abhisheksinghbaghel: Ok. removed this one.

docs/enhancements/feat-add-ability-to-tune-azuredisk-performance-parameters.md
abhisheksinghbaghel: One param. feature will be disabled by default, with a value of perfTuningMode=None

hack/gen-disk-skus-map.go
abhisheksinghbaghel: @andyzhangx Sure. Do you have an specific thing in mind? Or may be an example?

Also, since this is not in the hot path, please allow me to address this as a chore in a separate PR. I am evaluating to see weather I should spend time importing this information dynamically via the cloud provider, so that might be a better investment of our time.

pkg/azuredisk/device_perf_optimization.go
```diff
@@ -0,0 +1,48 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package azuredisk
+
+// This is the interface for DeviceHelper
+type Interface interface {
+	DiskSupportsPerfOptimization(diskPerfProfile string, diskAccountType string) bool
```
andyzhangx, line 21: it's better write like this `diskPerfProfile, diskAccountType string`

andyzhangx (commented):
/lgtm

PR #861, opened 2021-05-25
update: fix pull policy for helm template

sunpa93:
…f image.pullPolicy which doesn't exist beyond v0.5.0

**What type of PR is this?**
/kind bug

fixes helm template for driver version v0.5.0 and above.

Helm template for `csi-azuredisk-controller`, `csi-azuredisk-node`, and `csi-azuredisk-node-windows` are using values.images.pullPolicy instead of values.images.azuredisk.pullPolicy.

This was okay until v0.5.0 because pullPolicy was value's direct child field up until v0.4.0 but since v0.5.0, pullPolicies are children of individual images under values.

So this needs to be updated in order to avoid empty pull policy.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (commented):
since that was not a big issue, would you only fix this issue in master branch.

charts/v2.0.0-alpha.1/azuredisk-csi-driver/templates/csi-azuredisk-node-windows.yaml
```diff
@@ -91,7 +91,8 @@ spec:
             - "--endpoint=$(CSI_ENDPOINT)"
             - "--nodeid=$(KUBE_NODE_NAME)"
             - "--metrics-address=0.0.0.0:{{ .Values.node.metricsPort }}"
-            - "--temp-use-driver-v2"
+            - "--temp-use-driver-v2={{ .Values.azuredisk.useV2Driver }}"
+            - "--is-node-plugin={{ .Values.azuredisk.useV2Driver }}"
```
edreed, line 95: The `--is-node-plugin` flag is not understood by the upstream "v2" implementation and should be omitted.

charts/v2.0.0-alpha.1/azuredisk-csi-driver/templates/csi-azuredisk-node.yaml
```diff
@@ -84,7 +84,8 @@ spec:
             - "--endpoint=$(CSI_ENDPOINT)"
             - "--nodeid=$(KUBE_NODE_NAME)"
             - "--metrics-address=0.0.0.0:{{ .Values.node.metricsPort }}"
-            - "--temp-use-driver-v2"
+            - "--temp-use-driver-v2={{ .Values.azuredisk.useV2Driver }}"
+            - "--is-node-plugin={{ .Values.azuredisk.useV2Driver }}"
```
edreed, line 88: Omit `--is-node-plugin`

edreed:
@sunpa93 You will also need to regenerate the packages for the charts you update.

andyzhangx:
this PR(https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/843) already has the fix on master branch, it's not necessary to fix on other releases since it's a minor issue, default value `IfNotPresent` works in most cases.

PR #866, opened 2021-05-27
test: added ZRS versions for dynamic provisioning tests

kassarl:
**What type of PR is this?**

/kind test
Creates tests to cover possible scenarios of ZRS usage 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/test/e2e/dynamic_provisioning_test.go
+++ b/test/e2e/dynamic_provisioning_test.go
@@ -737,6 +737,360 @@ func (t *dynamicProvisioningTestSuite) defineTests(isMultiZone bool) {
 		}
 		test.Run(cs, ns)
 	})
+
+	ginkgo.It("should create a ZRS volume on demand with mount options [kubernetes.io/azure-disk] [disk.csi.azure.com] [Windows] [ZRS]", func() {
+		skipIfNotInZRSRegion()
+		pvcSize := "10Gi"
+		if isMultiZone {
+			pvcSize = "1000Gi"
+		}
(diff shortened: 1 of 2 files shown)


andyzhangx (commented):
that's duplicating too many test cases, I would suggest writing like this:
```
if current region supports ZRS:
  set StorageClassParameters as {"skuName": "StandardSSD_ZRS"}
```

That would save a bunch of test codes and easier to maintain.

kassarl:
> that's duplicating too many test cases, I would suggest writing like this:
> 
> ```
> if current region supports ZRS:
>   set StorageClassParameters as {"skuName": "StandardSSD_ZRS"}
> ```
> 
> That would save a bunch of test codes and easier to maintain.

In this case, does it matter that the test cases would test either ZRS or standard LRS, and not both?

andyzhangx:
> > that's duplicating too many test cases, I would suggest writing like this:
> > ```
> > if current region supports ZRS:
> >   set StorageClassParameters as {"skuName": "StandardSSD_ZRS"}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That would save a bunch of test codes and easier to maintain.
> 
> In this case, does it matter that the test cases would test either ZRS or standard LRS, and not both?

I think it's ok since this test case would be run  multiple times since we have quite a lot of different test pipelines in test-infra, and also ZRS depends on region.

PR #877, opened 2021-06-08
feat: add helm support for existing Service Accounts

### LINKED ISSUE #709: Support pre-created Service Accounts in the Helm chart

Currently this driver would create new service accounts, we should also support pre-created Service Accounts in the Helm chart

**Additional context**

kassarl:
**What type of PR is this?**
/kind feature

Allows a user to define an pre-created Service Account in the Helm chart, a feature offered offered by other CSI drivers

Fixes #709 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


charts/latest/azuredisk-csi-driver/values.yaml
```diff
@@ -25,11 +25,20 @@ image:
     pullPolicy: IfNotPresent
 
 serviceAccount:
-  create: true
+  controller:
+    create: true # When true, a service account will be created for you. Set to false if you want to use your own.
```
andyzhangx, line 29: that's redundant, could just use `serviceAccount.controller.name` instead since it would always be provided either by default value or user.

kassarl: I was using this "create" value in each of the templates for creating the SA (e.g. serviceaccount-csi-azuredisk-controller.yaml), since that was how it was previously structured. Should I instead do a check, if the name != default then create the service account?

andyzhangx: since there is already `serviceAccount.create` and `rbac.create`, only need to specify `controller.name`, `node.name`, `snapshotcontroller.name`, if user wants to bring their own serviceAccount, they need to `--set serviceAccount.create=false --set rbac.create=false --set serviceAccount.controller.name=x --set serviceAccount.node.name=x --set serviceAccount.snapshotcontroller.name=x`

charts/latest/azuredisk-csi-driver/values.yaml
```diff
@@ -25,11 +25,17 @@ image:
     pullPolicy: IfNotPresent
 
 serviceAccount:
-  create: true
+  create: true # When true, a service account will be created for you. Set to false if you want to use your own.
+  controller:
```
andyzhangx, line 29: what about set `serviceAccount.controller: csi-azuredisk-controller-sa`, `serviceAccount.node: csi-azuredisk-node-sa`, etc. ? That would make logic more clear.

charts/latest/azuredisk-csi-driver/values.yaml
```diff
@@ -25,11 +25,14 @@ image:
     pullPolicy: IfNotPresent
 
 serviceAccount:
-  create: true
+  create: true # When true, a service account will be created for you. Set to false if you want to use your own.
+  controller: csi-azuredisk-controller-sa # Name of Service Account to be created or used
+  node: csi-azuredisk-node-sa # Name of Service Account to be created or used
+  snapshotcontroller: csi-snapshot-controller-sa # Name of Service Account to be created or used
```
edreed, line 31: Since service account creation for the snapshot controller is enabled through `snapshot.snapshotController.serviceAccount`, it would be better to group this value with that setting, e.g. `snapshot.snapshotController.serviceAccountName`

andyzhangx: good catch, I think we could remove following settings
```
  snapshotController:
    serviceAccount: true
    rbac: true
```

should all depend on
```
serviceAccount:
  create: true

rbac:
  create: true
```

kassarl: OK, I have updated it to reflect the snapshot, controller, and node all depending on serviceAccount.create, and the snapshot rbac depending on rbac.create, and removed these fields from snapshot.snapshotController. I think this is a bit clearer, otherwise I can move the snapshot sa name to snapshot.snasphotController

andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm

PR #931, opened 2021-07-09
Update the tolerations of the controller

chunliu:
**What type of PR is this?**
/kind bug

The default value of tolerations of the controller assumes that the control plane has the taint `node-role.kubernetes.io/master=true:NoSchedule`. However, in some k8s clusters, such as the clusters created with kubeadm, the taint the control plane has is `node-role.kubernetes.io/master:NoSchedule`. The current tolerations of the controller don't match it, therefore the controller cannot be scheduled on the control plane. 

This change makes the tolerations match both the taints, `node-role.kubernetes.io/master=true:NoSchedule` and `node-role.kubernetes.io/master:NoSchedule`.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx:
thanks, can you also run `helm package charts/latest/azuredisk-csi-driver -d charts/latest/` to update chart file? thanks.

andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm
/approve

PR #932, opened 2021-07-09
fix: always enable csi-snapshotter in helm install

### LINKED ISSUE #930: Snapshots don't work for a custom deployment alongside the AKS 1.21 default deployment
**What happened**: A custom CSI driver per instructions (https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/charts/README.md#install-driver-with-customized-driver-name-deployment-name) is not able to snapshot

**What you expected to happen**: I expected to be able to snapshot volumes created by the custom deployment

**How to reproduce it**: Deploy with the values in the docs, create volumesnapshotclass, create volumesnapshot. volumesnapshot never reaches readytouse

**Anything else we need to know?**: Since the snapshotter pod is already running on the hidden AKS nodes, we should not deploy it as part of the custom deployment. Therefore snapshot.enabled is false. Unfortunately, that also means that the snapshot controller container is not part of the CSI controller deployment (https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml#L80).

**Environment**:
- CSI Driver version: 1.5.0
- Kubernetes version (use `kubectl version`): 1.21

andyzhangx:
**What type of PR is this?**
/kind bug

fix: always enable csi-snapshotter in helm install

Fixes #930

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: always enable csi-snapshotter in helm install
```


scoof:
WFM

PR #939, opened 2021-07-16
feat: customize hostNetwork setting in helm install

### LINKED ISSUE #936: Support for AAD pod-managed identities
I thought I had support for pod-managed identities by providing a way to label the pods, but as I dug deeper, I realized it wasn't enough.

There are currently two things that needs to be changed to fully support pod-managed identities:
* hostNetwork: false
* Injecting azure.json with the correct identity or without identity at all.

I have this working by mounting a manually configured ConfigMap in the pod, and using the azure-cred-file configmap to point to this CM.

Disabling hostNetwork is easy enough, but injecting azure.json with at least tenant, subscription and resource-group is a bit harder, since we need to pick up those from somewhere. I don't know if the only option is to provide them as values to the helm chart and render the azure.json file manually.

This should also remove the requirement to mount any hostPath volumes in the pod.

The alternative is to keep using the node managed identity, and that would be a less secure option, since that would delegate too many rights to the CSI driver.

andyzhangx:
**What type of PR is this?**
/kind feature

feat: customize hostNetwork setting in helm install

Fixes #936

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
feat: customize hostNetwork setting in helm install
```


feiskyer (commented):
/lgtm

PR #955, opened 2021-07-30
Disable uuid checks on XFS

### LINKED ISSUE #954: Failed to mount XFS volume + its restored snapshot to the same node
**What happened**:
When a XFS volume + its restored snapshot are NodeStaged to a single node, mount of the second volume fails with:
```
  Warning  FailedMount             3s (x4 over 9s)  kubelet                  MountVolume.MountDevice failed for volume "pvc-95794d96-69b5-4fa8-8db9-51db152fa1aa" : rpc error: code = Internal desc = could not format "/dev/disk/azure/scsi1/lun1"(lun: "1"), and mount it at "/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-95794d96-69b5-4fa8-8db9-51db152fa1aa/globalmount"
```

Both volumes are mounted

**How to reproduce it**:
1. Create XFS volume and use it in a Pod.
2. Take its snapshot.
3. Restore the snapshot as a separate PV/PVC.
4. Use the snapshot restored PVC in another pod. Make sure it's scheduled to the same node as the pod from step 1. (which is still running)

-> See the mount error.

XFS does not allow to mount two block devices with the same filesystem UUID - the original + restored snapshot in this case.
Ext4 works without any issues.

**Environment**:
- CSI Driver version: 1.3.0, but it should be reproducible with master too

jsafrane:
**What type of PR is this?**
/kind bug

Add `nouuid` mount option to all XFS mounts to be able to mount a volume and its restored snapshot on the same node. Without the option, such a mount fails, because XFS detects that two different volumes with the same filesystem UUID are being mounted.

Fixes #954

**Requirements**:
- [x] includes documentation - added release note below
- [ ] adds unit tests - there are no successful / mount option related NodeStage unit tests

**Release note**:
```
It is now possible to mount a volume with XFS filesystem and its restored snapshot on the same node.
```

### DIFF
--- a/pkg/azuredisk/nodeserver.go
+++ b/pkg/azuredisk/nodeserver.go
@@ -131,7 +131,7 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 		if mnt.FsType != "" {
 			fstype = mnt.FsType
 		}
-		options = append(options, mnt.MountFlags...)
+		options = append(options, collectMountOptions(fstype, mnt.MountFlags)...)
 	}
 
 	volContextFSType := getFStype(req.GetVolumeContext())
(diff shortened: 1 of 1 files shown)


andyzhangx (approved):
/lgtm
/approve

PR #961, opened 2021-08-04
fix: Remove gen-skus-map

jsafrane:
**What type of PR is this?**
/kind cleanup

I think `pkg/tool/gen-skus-map` was checked into git accidentally. There is source just next to it. And it does not seem to be used by anything in the repo. Anyone who needs the binary can build its own (hidden behind `make verify` though)

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (approved):
/lgtm
/approve

PR #962, opened 2021-08-05
Support UserAgent Attribution

kassarl:
**What type of PR is this?**

/kind feature

Allows for customer attribution for PVCs when they are created

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
Allows UserAgent attribution on PVCs via tagging
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -17,8 +17,13 @@ limitations under the License.
 package azuredisk
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
 	"path"
 	"sort"
 	"strconv"
@@ -40,6 +45,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/optimization"
+	"sigs.k8s.io/azuredisk-csi-driver/pkg/util"
 	volumehelper "sigs.k8s.io/azuredisk-csi-driver/pkg/util"
 	"sigs.k8s.io/cloud-provider-azure/pkg/metrics"
 	azure "sigs.k8s.io/cloud-provider-azure/pkg/provider"
@@ -293,6 +299,14 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	contentSource := &csi.VolumeContentSource{}
 	for k, v := range customTagsMap {
 		tags[k] = v
+		if k == userAgentField {
+			temp := d.cloud
+			d.cloud, err = SetupNewUserAgentCloud(v, d.cloud)
+			if err != nil {
+				klog.V(2).Infof("Unable to create new cloud for UserAgent, err: (%s)", err)
+			}
+			d.cloud = temp
+		}
 	}
 
 	if strings.EqualFold(writeAcceleratorEnabled, trueValue) {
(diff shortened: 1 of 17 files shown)


pkg/azuredisk/controllerserver.go
```diff
@@ -1219,3 +1233,78 @@ func getEntriesAndNextToken(req *csi.ListSnapshotsRequest, snapshots []compute.S
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = json.Unmarshal(configContents, &config)
+	if err != nil {
+		return nil, err
+	}
+
+	config.UserAgent = userAgent
```
kassarl, line 1297: The field should have been added via my cloud-provider-azure PR

andyzhangx (commented):
would you vendor the latest cloud provider first?

pkg/azuredisk/controllerserver.go
```diff
@@ -293,6 +299,14 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	contentSource := &csi.VolumeContentSource{}
 	for k, v := range customTagsMap {
 		tags[k] = v
+		if k == userAgentField {
```
andyzhangx, line 302: user agent does not need to be set in storage class, we only need to hard code in our driver or set as driver parameter in driver start up

edreed: @andyzhangx We need a way to track Azure cloud operations initiated by partner ISVs on behalf of another customer. This is done through the user agent strings as described in [Azure customer usage attribution](https://docs.microsoft.com/en-us/azure/marketplace/azure-partner-customer-usage-attribution).

andyzhangx: I would suggest add a new parameter(`userAgent`) instead of customer tags in storage class to track csi driver operations. if not set, use hard coded userAgent in driver, the implementation in this PR is too redudunt, here is an clean example: https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/754/commits/ac18a0b4df486ad40d4479ce8d6bb2567cd03fb5

kassarl: The logic that was behind this was that ISVs would need a specific storage class created that tags useragent on operations, but wouldn't want the rest of the operations to be tagged with the Useragent. Vybava is verifying what method makes the most sense, but this was the implementation we initially decided to go for

andyzhangx: while current implementation would set useragent for all operations done by the driver since this whole driver would use only one useragent. Is there a customer asking for this feature? specify useragent in storage class(no matter it's tags or `userAgent` field) is not a good idea: 1) PV does not use this storage class would also get the same useragent 2) if driver restart, userAgent setting would be gone.

andyzhangx: I created another PR to change default userAgent(https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/969), though it's not related to this scenario.

pkg/azuredisk/controllerserver.go
```diff
@@ -201,6 +210,15 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			if strings.EqualFold(v, trueValue) {
 				enableBursting = to.BoolPtr(true)
 			}
+		case userAgentField:
+			nonUserAgentCloud = d.cloud
+			newUserAgent := v
+			d.cloud, err = SetupNewUserAgentCloud(newUserAgent, d.cloud)
```
edreed, line 216: We don't want to change the global instance of the cloud provider as we only want the custom user agent string used for the duration of this `CreateVolume` call and not for any others. Use a local variable initialized to `d.cloud` for the `CreateVolume` call and override the value here instead.

pkg/azuredisk/controllerserver.go
```diff
@@ -1219,3 +1241,139 @@ func getEntriesAndNextToken(req *csi.ListSnapshotsRequest, snapshots []compute.S
 
 	return listSnapshotResp, nil
 }
+
+// Creates a new cloud during CreateVolume call to configure the Useragent property of DiskClient
+func SetupNewUserAgentCloud(userAgent string, currentAz *azure.Cloud) (*azure.Cloud, error) {
```
edreed, line 1246: Consider `GetCloudProviderWithUserAgent` and move this and supporting functions to same file as `GetCloudProvider`. (This is currently `pkg/azuredisk/azure.go`, but #977 will be moving it to `pkg/azureutils/azure_disk_utils.go`.)

vendor/sigs.k8s.io/cloud-provider-azure/pkg/provider/azure_backoff.go
```diff
@@ -384,7 +384,7 @@ func (az *Cloud) DeletePublicIP(service *v1.Service, pipResourceGroup string, pi
 }
 
 // DeleteLB invokes az.LoadBalancerClient.Delete with exponential backoff retry
-func (az *Cloud) DeleteLB(service *v1.Service, lbName string) error {
```
edreed, line 387: You have updates to the `vendor` directory, but I don't see any corresponding updates to `go.mod`. Are you missing this file in the PR?

kassarl: I must have forgotten to add go.mod. When I updated the vendors, it changed almost all of the packages so I was trying to just include the ones relevant. I'll fix that

andyzhangx: vendor change was already done in previous PR, just rebase to current master branch

pkg/azuredisk/controllerserver.go
andyzhangx: pls reuse existing functions, and don't duplicate functions, code duplication would make it difficult to maintain

pkg/azuredisk/controllerserver.go
```diff
@@ -1219,3 +1241,139 @@ func getEntriesAndNextToken(req *csi.ListSnapshotsRequest, snapshots []compute.S
...
+	}
+	// reassign kubeClient
+	if kubeClient != nil && az.KubeClient == nil {
+		az.KubeClient = kubeClient
+	}
+	return az, nil
+}
+
+// Creates a config buffer that includes a useragent to be used for cloud creation
+func UpdateConfigUserAgent(userAgent string, credFile string) (io.Reader, error) {
```
andyzhangx, line 1292: there is a way to reset userAgent easily:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/3bac0081d7046b30e084dc5a56bbb730d0b1f26f/pkg/azuredisk/azure.go#L105-L108

I would suggest reuse `GetCloudProvider` func

kassarl: When I tried utilizing that method, upon inspection of the az object after the InitializeCloudFromConfig function has run, the initialization overwrites it with a nil value since the config does not define it. I can take another look but I'm pretty sure it just gets overwritten

andyzhangx (commented):
/ok-to-test

pkg/azuredisk/controllerserver.go
andyzhangx: actually in Line 219, it already changes the global instance. It's hard to set multiple cloud provider instances for multiple  userAgents unless we set a big lock in CreateVolume to block other create volume activities until it succeeds, while set a lock would bring performance issue.

PR #976, opened 2021-08-19
Merge v2 changes as of 08/19/2021

mukhoakash:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


.github/workflows/windows.yml
```diff
@@ -22,13 +22,4 @@ jobs:
           make azuredisk-windows
       - name: Run Windows Unit Tests
         run: |
-          # start the CSI Proxy before running tests on windows
-          Start-Job -Name CSIProxy -ScriptBlock {
-            Invoke-WebRequest https://github.com/andyzhangx/demo/raw/master/windows/csi-proxy-v1.0.0-rc.1.tar.gz -OutFile csi-proxy.tar.gz;
-            tar -xvf csi-proxy.tar.gz
-            .\bin\csi-proxy.exe --kubelet-path $pwd
```
andyzhangx, line 29: are you skipping csi-proxy based unit tests in v2 now? That would lose test coverage on Windows a lot.
cc @edreed

edreed: Unit tests run in the developer inner-loop and should exercise the code under development w/o external dependencies. Taking a dependency on the CSI Proxy makes it more difficult to run them as part of the developer inner-loop on Windows, so I eliminated that dependency for the unit tests relying on a fake instead. That way, the unit tests can be run on Windows w/o any additional set up. CSI Proxy and integration with it should get sufficient coverage from the Windows E2E tests.

andyzhangx: it's already using `https://acs-mirror.azureedge.net/csi-proxy/v1.0.0/binaries/csi-proxy-v1.0.0.tar.gz` on master branch:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/7dcacf81cb4b562d292bce6d7389f1a2b311edb9/.github/workflows/windows.yml#L27-L29

andyzhangx: maybe add a flag(e.g. `skipCSIProxyTests`) in unit test, for developing on Windows, we could set that flag.

edreed: @andyzhangx  What about reversing the sense of the switch, i.e. `--use-csi-proxy`. That way a simple `go test ./pkg/azuredisk` can be run from the developers command line to run the unit tests on Windows, and we could add the `--use-csi-proxy` command line switch in the Github action to get the additional CSI Proxy coverage?

andyzhangx: lgtm, I think you could fix that issue in another PR, not necessary to fix it in this PR.

andyzhangx:
```
cleanup snapshots: DeleteSnapshot failed: rpc error: code = Internal desc = delete snapshot error: Retriable: true, RetryAfter: 9s, HTTPStatusCode: 0, RawError: azure cloud provider throttled for operation SnapshotDelete with reason "client throttled"
```

PR #977, opened 2021-08-20
chor: Refactoring utilities to a separate folder to avoid duplication 

landreasyan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azuredisk/azuredisk.go
```diff
@@ -20,100 +20,26 @@ import (
...
 
+	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
 	csicommon "sigs.k8s.io/azuredisk-csi-driver/pkg/csi-common"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/mounter"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/optimization"
 	volumehelper "sigs.k8s.io/azuredisk-csi-driver/pkg/util"
 	azcache "sigs.k8s.io/cloud-provider-azure/pkg/cache"
 	consts "sigs.k8s.io/cloud-provider-azure/pkg/consts"
 	"sigs.k8s.io/cloud-provider-azure/pkg/provider"
-	azure "sigs.k8s.io/cloud-provider-azure/pkg/provider"
```
edreed, line 46: For consistency with other imports, we should maintain the `azure` alias and replace the direct usage of `provider` above with `azure`.

pkg/azureutils/azure_disk_constants.go
```diff
@@ -0,0 +1,77 @@
...
+limitations under the License.
+*/
+
+package azureutils
+
+import (
+	"regexp"
+)
+
+const (
```
edreed, line 23: I wonder if we should move these to their own `constants` package to consistently access them through that alias rather than both `constants` and `azureutils`?

pkg/azureutils/azure_disk_constants.go
```diff
@@ -0,0 +1,77 @@
...
+	TrueValue                     = "true"
+	VolumeAttributePartition      = "partition"
+	WellKnownTopologyKey          = "topology.kubernetes.io/zone"
+	WriteAcceleratorEnabled       = "writeacceleratorenabled"
+)
+
+var (
+	// ManagedDiskPath is described here: https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate#create-a-managed-disk-from-an-existing-managed-disk-in-the-same-or-different-subscription.
+	ManagedDiskPath   = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/disks/%s"
+	ManagedDiskPathRE = regexp.MustCompile(`(?i).*/subscriptions/(?:.*)/resourceGroups/(?:.*)/providers/Microsoft.Compute/disks/(.+)`)
```
edreed, line 76: These could also be `const`.

pkg/azureutils/azure_disk_constants.go
landreasyan: I was also thinking of making constants separate package but left it as this for now. If there are no objections for having them as separate package, I would love to make that change.

pkg/azureutils/azure_disk_constants.go
landreasyan: Unfortunately, ` regexp.MustCompile` is not a constant, so I left `ManagedDiskPath` in this section as well as a logical grouping.

pkg/azuredisk/azuredisk.go
landreasyan: I was trying to avoid renaming imports as it makes it more likely that we might unknowingly import the same package twice under different names (as seen above). I was saving package renaming only for cases where there is conflict but different linters act differently and some even directly name imports even if the name is the same...

pkg/azureutils/azure_disk_constants.go
edreed: I have no objects - I believe it would be a good thing to do. @andyzhangx What do you think?

andyzhangx: use `consts` instead of `constants` ?

pkg/azuredisk/nodeserver.go
```diff
@@ -81,7 +82,7 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 	}
 	defer d.volumeLocks.Release(diskURI)
 
-	lun, ok := req.PublishContext[LUN]
+	lun, ok := req.PublishContext[azureutils.LUN]
```
andyzhangx, line 85: `consts.LUN` ?

pkg/azuredisk/nodeserver.go
```diff
@@ -141,7 +142,7 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 	}
 
 	// If partition is specified, should mount it only instead of the entire disk.
-	if partition, ok := req.GetVolumeContext()[volumeAttributePartition]; ok {
+	if partition, ok := req.GetVolumeContext()[azureutils.VolumeAttributePartition]; ok {
```
andyzhangx, line 145: `consts.VolumeAttributePartition`

pkg/azuredisk/nodeserver.go
landreasyan: Since this file uses utilities functions from `azureutils`, I have imported the package as utilities instead of `consts`. But I will be separating the two to have a constants package and can then make this change.

pkg/azureutils/azure_disk_constants.go
landreasyan: I have refactored constants into a separate pkg.

pkg/azuredisk/azuredisk.go
```diff
@@ -20,22 +20,20 @@ import (
...
-	"github.com/pborman/uuid"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
 	"k8s.io/klog/v2"
-	"k8s.io/kubernetes/pkg/volume/util"
 	"k8s.io/mount-utils"
 
+	azureconsts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
```
edreed, line 35: Use `consts` for consistent aliasing - this is the only file to use `azureconsts`

andyzhangx (approved):
/lgtm
/approve

andyzhangx:
there is ut failure:
```
sigs.k8s.io/azuredisk-csi-driver/pkg/azuredisk [sigs.k8s.io/azuredisk-csi-driver/pkg/azuredisk.test]
pkg/azuredisk/controllerserver_test.go:136:8: undefined: logicalSectorSizeField
pkg/azuredisk/controllerserver_test.go:359:8: undefined: skuNameField
pkg/azuredisk/controllerserver_test.go:371:23: undefined: managedDiskPath
pkg/azuredisk/controllerserver_test.go:771:19: undefined: cachingModeField
```

pkg/azuredisk/azuredisk.go
landreasyan: `consts` was already taken by `consts "sigs.k8s.io/cloud-provider-azure/pkg/consts"` but I guess I can rename that import instead.

PR #990, opened 2021-08-27
feat: expose metadata for static PV attachment

andyzhangx:
**What type of PR is this?**
/kind feature

feat: expose metadata for static PV attachment

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

```console
# k get volumeattachments
NAME                                                                   ATTACHER             PV                                         NODE                           ATTACHED   AGE
csi-b848a2790e08e279bd52b47c1db0177d5d804cfbccc64e28b3392028f972527d   disk.csi.azure.com   pv-azuredisk                               aks-disk-18870927-vmss000000   true       10m

# k get volumeattachments csi-b848a2790e08e279bd52b47c1db0177d5d804cfbccc64e28b3392028f972527d -o yaml
apiVersion: storage.k8s.io/v1
kind: VolumeAttachment
metadata:
  annotations:
    csi.alpha.kubernetes.io/node-id: aks-disk-18870927-vmss000000
  creationTimestamp: "2021-09-03T10:46:59Z"
  finalizers:
  - external-attacher/disk-csi-azure-com
  name: csi-b848a2790e08e279bd52b47c1db0177d5d804cfbccc64e28b3392028f972527d
  resourceVersion: "17068153"
  uid: c4412fa9-0abc-4cf5-86d5-3f3505f2c506
spec:
  attacher: disk.csi.azure.com
  nodeName: aks-disk-18870927-vmss000000
  source:
    persistentVolumeName: pv-azuredisk
status:
  attached: true
  attachmentMetadata:
    LUN: "5"
    diskiopsreadwrite: "500"
    diskmbpsreadwrite: "60"
    networkaccesspolicy: AllowAll
    skuname: StandardSSD_LRS
```

**Release note**:
```
feat: expose metadata for static PV attachment
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -293,11 +293,9 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 				},
 			}
 
-			ctx, cancel := context.WithCancel(context.Background())
 			if sourceGiB, _ := d.GetSourceDiskSize(ctx, resourceGroup, path.Base(sourceID), 0, consts.SourceDiskSearchMaxDepth); sourceGiB != nil && *sourceGiB < int32(requestGiB) {
 				parameters[consts.ResizeRequired] = strconv.FormatBool(true)
 			}
-			cancel()
 		}
 	}
 
@@ -476,9 +474,16 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
 		klog.V(2).Infof("attach volume %q to node %q successfully", diskURI, nodeName)
 	}
 
-	pvInfo := map[string]string{consts.LUN: strconv.Itoa(int(lun))}
+	publishContext := map[string]string{consts.LUN: strconv.Itoa(int(lun))}
+	volumeContext := req.VolumeContext
+	if disk != nil && volumeContext != nil {
+		if _, ok := volumeContext[consts.RequestedSizeGib]; !ok {
+			klog.V(2).Infof("found static PV(%s), insert disk properties to volumeattachments", diskURI)
+			azureutils.InsertDiskProperties(disk, publishContext)
+		}
+	}
 	isOperationSucceeded = true
-	return &csi.ControllerPublishVolumeResponse{PublishContext: pvInfo}, nil
+	return &csi.ControllerPublishVolumeResponse{PublishContext: publishContext}, nil
 }
 
 // ControllerUnpublishVolume detach an azure disk from a required node
--- a/pkg/azuredisk/controllerserver_v2.go
+++ b/pkg/azuredisk/controllerserver_v2.go
@@ -445,9 +445,16 @@ func (d *DriverV2) ControllerPublishVolume(ctx context.Context, req *csi.Control
 		klog.V(2).Infof("attach volume %q to node %q successfully", diskURI, nodeName)
 	}
 
-	pvInfo := map[string]string{consts.LUN: strconv.Itoa(int(lun))}
+	publishContext := map[string]string{consts.LUN: strconv.Itoa(int(lun))}
+	volumeContext := req.VolumeContext
+	if disk != nil && volumeContext != nil {
+		if _, ok := volumeContext[consts.RequestedSizeGib]; !ok {
+			klog.V(2).Infof("found static PV(%s), insert disk properties to volumeattachments", diskURI)
+			azureutils.InsertDiskProperties(disk, publishContext)
+		}
+	}
 	isOperationSucceeded = true
-	return &csi.ControllerPublishVolumeResponse{PublishContext: pvInfo}, nil
+	return &csi.ControllerPublishVolumeResponse{PublishContext: publishContext}, nil
 }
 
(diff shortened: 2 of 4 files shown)


pkg/azureconstants/azure_constants.go
```diff
@@ -68,6 +68,8 @@ const (
 	VolumeAttributePartition      = "partition"
 	WellKnownTopologyKey          = "topology.kubernetes.io/zone"
 	WriteAcceleratorEnabled       = "writeacceleratorenabled"
+	CreatedByTag                  = "created-by"
+	AzureDDTag                    = "kubernetes-azure-dd"
```
shrac, line 72: There is a chance that a customer would like to take a previously created dynamic azure disk and use it as a static PV elsewhere. In this case the tag might be present even though it is being attached as a static PV. Is there a more reliable way of knowing it Like storage class being empty ?

pkg/azureutils/azure_disk_utils.go
```diff
@@ -453,3 +453,43 @@ func getKubeClient(kubeconfig string) (*kubernetes.Clientset, error) {
...
+// InsertDiskProperties: insert disk properties to map
+func InsertDiskProperties(disk *compute.Disk, publishConext map[string]string) {
+	if disk == nil || publishConext == nil {
+		return
+	}
+
+	if disk.Sku != nil {
+		publishConext[consts.SkuNameField] = string(disk.Sku.Name)
+	}
+	prop := disk.DiskProperties
```
shrac, line 478: Can we add location and maxShares properties as well to this list.

andyzhangx: even dynamic PV does not have `location` field, and `availabilityZone` is already in the `allowedTopologies` field of PV, same as dynamic PV.

added `maxShares`

pkg/azureconstants/azure_constants.go
andyzhangx: fixed by relying another field `RequestedSizeGib` which could only be created by dynamic PV

pkg/azuredisk/controllerserver.go
```diff
@@ -476,9 +474,16 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
 		klog.V(2).Infof("attach volume %q to node %q successfully", diskURI, nodeName)
 	}
 
-	pvInfo := map[string]string{consts.LUN: strconv.Itoa(int(lun))}
+	publishContext := map[string]string{consts.LUN: strconv.Itoa(int(lun))}
+	volumeConext := req.VolumeContext
```
edreed, line 478: sp: `volumeContext`

pkg/azuredisk/controllerserver_v2.go
```diff
@@ -445,9 +445,16 @@ func (d *DriverV2) ControllerPublishVolume(ctx context.Context, req *csi.Control
 		klog.V(2).Infof("attach volume %q to node %q successfully", diskURI, nodeName)
 	}
 
-	pvInfo := map[string]string{consts.LUN: strconv.Itoa(int(lun))}
+	publishContext := map[string]string{consts.LUN: strconv.Itoa(int(lun))}
+	volumeConext := req.VolumeContext
```
edreed, line 449: sp: `volumeContext`

pkg/azuredisk/controllerserver.go
```diff
@@ -476,9 +474,16 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
...
+	if disk != nil && volumeConext != nil {
+		if _, ok := volumeConext[consts.RequestedSizeGib]; !ok {
```
edreed, line 480: Is it necessary to restrict this to static provisioned PVs? In the dynamic case, the metadata could be redundant to that already in the PV but clients would take the same approach to get it (by always looking at the `VolumeAttachment`). This would avoid the need for having two different ways of getting the metadata in the dynamic and static cases.

pkg/azuredisk/controllerserver.go
andyzhangx: fixed

pkg/azuredisk/controllerserver.go
andyzhangx: As discussed, for dynamic PVs, azure back service should use metadata in PV which is more reliable and generic. Since exposing disk metadata in `volumeattachments` is only required by azure back service, I will leave this question to @shrac

shrac: We want to read VolumeAttachment objects only where required (for now static PVs) as that would mean fetching more info and might delay the validation call - Also, I could not find a way to filter a VolumeAttachment by its PV name/uuid. 

Also, there is no guarantee that VolumeAttachment would always exist. Ex: For PVs that were never attached (or) existing PVs which are already attached and the new node publish call is yet to come. As the number of such static PVs should be less we will fail validation for now but we cannot do the same for dynamic PVs.  Hence, we would continue relying on PV volume attributes for dynamic PVs.

edreed: Right. Makes sense.

PR #999, opened 2021-09-01
feat: allow storage class setting of useragent

kassarl:
**What type of PR is this?**

/kind feature
Allows useragent to be set through the storage class

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -104,8 +104,10 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		diskAccessID            string
 		maxShares               int
 		enableBursting          *bool
+		localCloud              *azure.Cloud
 	)
 
+	localCloud = d.cloud
 	tags := make(map[string]string)
 	parameters := req.GetParameters()
 	if parameters == nil {
@@ -171,12 +173,19 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			if strings.EqualFold(v, consts.TrueValue) {
 				enableBursting = to.BoolPtr(true)
 			}
+		case consts.UserAgentField:
+			newUserAgent := v
+			localCloud, err = azureutils.GetCloudProvider(d.kubeconfig, d.cloudConfigSecretName, d.cloudConfigSecretNamespace, newUserAgent)
+			if err != nil {
+				klog.V(2).Infof("Unable to create new cloud for UserAgent, err: (%s)", err)
+				localCloud = d.cloud
+			}
 		default:
 			return nil, fmt.Errorf("invalid parameter %s in storage class", k)
 		}
 	}
 
-	if azureutils.IsAzureStackCloud(d.cloud.Config.Cloud, d.cloud.Config.DisableAzureStackCloud) {
+	if azureutils.IsAzureStackCloud(localCloud.Config.Cloud, localCloud.Config.DisableAzureStackCloud) {
 		if maxShares > 1 {
 			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Invalid maxShares value: %d as Azure Stack does not support shared disk.", maxShares))
 		}
@@ -198,11 +207,11 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	diskName = azureutils.CreateValidDiskName(diskName)
 
 	if resourceGroup == "" {
-		resourceGroup = d.cloud.ResourceGroup
+		resourceGroup = localCloud.ResourceGroup
 	}
 
 	// normalize values
-	skuName, err := azureutils.NormalizeStorageAccountType(storageAccountType, d.cloud.Config.Cloud, d.cloud.Config.DisableAzureStackCloud)
+	skuName, err := azureutils.NormalizeStorageAccountType(storageAccountType, localCloud.Config.Cloud, localCloud.Config.DisableAzureStackCloud)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


pkg/azuredisk/controllerserver.go
```diff
@@ -104,8 +104,10 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		diskAccessID            string
 		maxShares               int
 		enableBursting          *bool
+		localCloud              *azure.Cloud
 	)
 
+	localCloud = d.cloud
```
edreed, line 110: I believe we also want to do this with `CreateSnapshot` as well since 3rd party backup vendors will want to attribute snapshots to their products' backup workflows.

andyzhangx (commented):
/ok-to-test

pkg/azuredisk/controllerserver.go
```diff
@@ -171,12 +173,19 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+				klog.V(2).Infof("Unable to create new cloud for UserAgent, err: (%s)", err)
+				localCloud = d.cloud
+			}
 		default:
 			return nil, fmt.Errorf("invalid parameter %s in storage class", k)
 		}
 	}
 
-	if azureutils.IsAzureStackCloud(d.cloud.Config.Cloud, d.cloud.Config.DisableAzureStackCloud) {
+	if azureutils.IsAzureStackCloud(localCloud.Config.Cloud, localCloud.Config.DisableAzureStackCloud) {
```
andyzhangx, line 188: `localCloud` may be nil if userAgent is not set, this would panic

kassarl: its set in 110

andyzhangx: ok, then it's better write like this `localCloud := d.cloud` and remove L107

pkg/azuredisk/controllerserver.go
```diff
@@ -171,12 +172,19 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			if strings.EqualFold(v, consts.TrueValue) {
 				enableBursting = to.BoolPtr(true)
 			}
+		case consts.UserAgentField:
+			newUserAgent := v
+			localCloud, err = azureutils.GetCloudProvider(d.kubeconfig, d.cloudConfigSecretName, d.cloudConfigSecretNamespace, newUserAgent)
+			if err != nil {
+				klog.V(2).Infof("Unable to create new cloud for UserAgent, err: (%s)", err)
```
andyzhangx, line 179: Should return error directly, make `CreateVolume' fail:
```
return fmt.Errorf("create cloud with UserAgent(%s) failed with: (%s)", newUserAgent, err)
```

pkg/azuredisk/controllerserver.go
```diff
@@ -838,12 +847,19 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 			}
 		case consts.ResourceGroupField:
 			resourceGroup = v
+		case consts.UserAgentField:
+			newUserAgent := v
+			localCloud, err = azureutils.GetCloudProvider(d.kubeconfig, d.cloudConfigSecretName, d.cloudConfigSecretNamespace, newUserAgent)
```
andyzhangx, line 852: Should return error directly, make `CreateSnapshot' fail:
```
return fmt.Errorf("create cloud with UserAgent(%s) failed with: (%s)", newUserAgent, err)
```

pkg/azuredisk/controllerserver.go
```diff
@@ -217,7 +225,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 
 	requirement := req.GetAccessibilityRequirements()
-	diskZone := azureutils.PickAvailabilityZone(requirement, d.cloud.Location, topologyKey)
+	diskZone := azureutils.PickAvailabilityZone(requirement, localCloud.Location, topologyKey)
```
andyzhangx, line 228: since `userAgent` is only changed, I think it's ok to use `d.cloud.Location`

pkg/azuredisk/controllerserver.go
```diff
@@ -198,11 +206,11 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	diskName = azureutils.CreateValidDiskName(diskName)
 
 	if resourceGroup == "" {
-		resourceGroup = d.cloud.ResourceGroup
```
andyzhangx, line 201: since only `userAgent` is changed, I think it's ok to use `d.cloud.ResourceGroup`

pkg/azuredisk/controllerserver.go
```diff
@@ -872,18 +888,18 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 			},
 			Incremental: &incremental,
 		},
-		Location: &d.cloud.Location,
+		Location: &localCloud.Location,
 		Tags:     tags,
 	}
 
-	mc := metrics.NewMetricContext(consts.AzureDiskCSIDriverName, "controller_create_snapshot", d.cloud.ResourceGroup, d.cloud.SubscriptionID, d.Name)
+	mc := metrics.NewMetricContext(consts.AzureDiskCSIDriverName, "controller_create_snapshot", localCloud.ResourceGroup, localCloud.SubscriptionID, d.Name)
```
andyzhangx, line 895: don't need such change since only userAgent is changed

pkg/azuredisk/controllerserver.go
edreed: Shouldn't this be the following?

```go
return status.Errorf(codes.InvalidArgument, "create cloud with UserAgent(%s) failed with: (%s)", newUserAgent, err)
```

andyzhangx (approved):
/lgtm
/approve
thanks, that's clear enough, and I also added e2e test for userAgent with another commit in this PR.

PR #1005, opened 2021-09-06
fix: enable avset setting by default

andyzhangx:
**What type of PR is this?**
/kind bug

fix: enable avset setting by default
fix the test case failure on capz standalone vm type

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: enable avset setting by default
```


ZeroMagic (approved):
/lgtm

PR #1060, opened 2021-10-29
chore: Refactoring cloud provider initialization to merge v1 and v2 driver implementations.

landreasyan:
**What type of PR is this?**
Currently, v1 and v2 drivers have separate implementation for initializing the cloud provider. This PR is to refactor and merge these differences in order to eliminate any duplicate code.  

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azureutils/azure_disk_utils.go
+++ b/pkg/azureutils/azure_disk_utils.go
@@ -29,12 +29,12 @@ import (
 	"github.com/pborman/uuid"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 	"k8s.io/mount-utils"
 
+	clientset "k8s.io/client-go/kubernetes"
 	api "k8s.io/kubernetes/pkg/apis/core"
 	volumeUtil "k8s.io/kubernetes/pkg/volume/util"
 	consts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
@@ -111,8 +111,11 @@ func GetCachingMode(attributes map[string]string) (compute.CachingTypes, error) 
 	return compute.CachingTypes(cachingMode), err
 }
 
-// GetCloudProvider get Azure Cloud Provider
-func GetCloudProvider(kubeconfig, secretName, secretNamespace, userAgent string) (*azure.Cloud, error) {
+// GetCloudProviderFromClient get Azure Cloud Provider
+func GetCloudProviderFromClient(kubeClient *clientset.Clientset, secretName string, secretNamespace string, userAgent string) (*azure.Cloud, error) {
+	var config *azure.Config
+	var fromSecret bool
+	var err error
 	az := &azure.Cloud{
 		InitSecretConfig: azure.InitSecretConfig{
 			SecretName:      secretName,
@@ -120,19 +123,6 @@ func GetCloudProvider(kubeconfig, secretName, secretNamespace, userAgent string)
 			CloudConfigKey:  "cloud-config",
 		},
 	}
-
-	kubeClient, err := getKubeClient(kubeconfig)
-	if err != nil {
-		klog.Warningf("get kubeconfig(%s) failed with error: %v", kubeconfig, err)
-		if !os.IsNotExist(err) && err != rest.ErrNotInCluster {
-			return az, fmt.Errorf("failed to get KubeClient: %v", err)
-		}
-	}
-	var (
-		config     *azure.Config
-		fromSecret bool
-	)
-
 	if kubeClient != nil {
 		klog.V(2).Infof("reading cloud config from secret %s/%s", az.SecretNamespace, az.SecretName)
 		az.KubeClient = kubeClient
@@ -161,13 +151,13 @@ func GetCloudProvider(kubeconfig, secretName, secretNamespace, userAgent string)
 
 		credFileConfig, err := os.Open(credFile)
 		if err != nil {
-			klog.Warningf("load azure config from file(%s) failed with %v", credFile, err)
-		} else {
-			defer credFileConfig.Close()
-			klog.V(2).Infof("read cloud config from file: %s successfully", credFile)
-			if config, err = azure.ParseConfig(credFileConfig); err != nil {
-				klog.Warningf("parse config file(%s) failed with error: %v", credFile, err)
-			}
+			klog.Errorf("Failed to load config from file: %s", credFile)
+			return nil, fmt.Errorf("Failed to load config from file: %s, cloud not get azure cloud provider", credFile)
+		}
+		defer credFileConfig.Close()
+		klog.V(2).Infof("read cloud config from file: %s successfully", credFile)
+		if config, err = azure.ParseConfig(credFileConfig); err != nil {
+			klog.Warningf("parse config file(%s) failed with error: %v", credFile, err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/azureutils/azure_disk_utils.go
```diff
@@ -194,6 +185,44 @@ func GetCloudProvider(kubeconfig, secretName, secretNamespace, userAgent string)
...
+		return nil, fmt.Errorf("failed to get KubeClient: %v", err)
+	}
+	az, err := GetCloudProviderFromClient(kubeClient, secretName, secretNamespace, userAgent)
+	return az, err
+}
+
+// GetKubeConfig gets config object from config file
+func GetKubeConfig(kubeconfig string) (config *rest.Config, err error) {
+	// attempt to read the in-cluster config
+	config, err = rest.InClusterConfig()
```
andyzhangx, line 201: this func is changing the original behavior, any reason why change this?

https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/36b2a9a078fb25728bb909045926e6abfa45565a/pkg/azureutils/azure_disk_utils.go#L445-L461

pkg/azureutils/azure_disk_utils.go
```diff
@@ -194,6 +185,44 @@ func GetCloudProvider(kubeconfig, secretName, secretNamespace, userAgent string)
...
+	// if we couldn't get the in-cluster config
+	// get kubeconfig path from environment variable
+	if err != nil {
+		if kubeconfig == "" {
+			kubeconfig = os.Getenv("KUBECONFIG")
+			if kubeconfig == "" {
+				kubeconfig = filepath.Join(os.Getenv("HOME"), ".kube", "config")
```
andyzhangx, line 208: `HOME` path is actually inside driver container, would v2 driver use `~/.kube/config` as kubeconfig? I doubt that.

pkg/azureutils/azure_disk_utils.go
landreasyan: I am trying to merge the behaviors we had in the v1 and v2 drivers. The recommended approach is to read the in cluster config and then fall back to kubeconfig if it fails. The behavior is essentially  the same in the old implementation as well. Was there a reason to prioritize kubeconfig over the in cluster one?

pkg/azureutils/azure_disk_utils.go
landreasyan: @mukhoakash I think this was originally your change. I have kept it as an additional fall back scenario but was there any other reason for adding this?

pkg/azureutils/azure_disk_utils.go
edreed: If `kubeConfig` is not empty, it means that it was specified on the command line. In this case, it should probably take precedence here.

pkg/azureutils/azure_disk_utils.go
edreed: I suspect this was done to ease running the driver locally on a dev box since this would pick up whatever the current Kubernetes cluster context had been set to.

pkg/azureutils/azure_disk_utils.go
landreasyan: But if we are running inside a cluster and our driver runs inside a pod, is there a scenario we where we would prefer the `kubeconfig` rather than the in-cluster config? From what I understand, the `kubeconfig` is mostly important for testing when the suite does not actually run inside a pod.

andyzhangx: if `kubeconfig` is set by user, it must be honored first:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/c1e583969e90b35140dea33d41ba5e2380b158ad/pkg/azurediskplugin/main.go#L44

pkg/azureutils/azure_disk_utils.go
andyzhangx: that would require running this driver as a binary instead of container, that's very rare case, if really needed, just set `kubeconfig` env var. I would suggest keep original logic:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/36b2a9a078fb25728bb909045926e6abfa45565a/pkg/azureutils/azure_disk_utils.go#L445-L461

pkg/azureutils/azure_disk_utils.go
landreasyan: I have reverted the changes to the getter and kept the old logic.

pkg/azureutils/azure_disk_utils.go
```diff
@@ -194,6 +184,39 @@ func GetCloudProvider(kubeconfig, secretName, secretNamespace, userAgent string)
 	return az, nil
 }
 
+// GetCloudProviderFromConfig get Azure Cloud Provider
+func GetCloudProvider(kubeConfig, secretName, secretNamespace, userAgent string) (*azure.Cloud, error) {
+	kubeClient, err := GetKubeClient(kubeConfig)
+	if err != nil && !os.IsNotExist(err) && err != rest.ErrNotInCluster {
+		return nil, fmt.Errorf("failed to get KubeClient: %v", err)
+	}
+	az, err := GetCloudProviderFromClient(kubeClient, secretName, secretNamespace, userAgent)
```
andyzhangx, line 193: `return GetCloudProviderFromClient(kubeClient, secretName, secretNamespace, userAgent)`

pkg/azureutils/azure_disk_utils.go
```diff
@@ -111,28 +111,18 @@ func GetCachingMode(attributes map[string]string) (compute.CachingTypes, error)
...
 		InitSecretConfig: azure.InitSecretConfig{
 			SecretName:      secretName,
 			SecretNamespace: secretNamespace,
 			CloudConfigKey:  "cloud-config",
 		},
 	}
-
-	kubeClient, err := getKubeClient(kubeconfig)
-	if err != nil {
-		klog.Warningf("get kubeconfig(%s) failed with error: %v", kubeconfig, err)
```
andyzhangx, line 126: could you keep this warning `klog.Warningf("get kubeconfig(%s) failed with error: %v", kubeconfig, err)` in L190?

pkg/azureutils/azure_disk_utils.go
```diff
@@ -194,6 +184,39 @@ func GetCloudProvider(kubeconfig, secretName, secretNamespace, userAgent string)
...
+	kubeClient, err := GetKubeClient(kubeConfig)
+	if err != nil && !os.IsNotExist(err) && err != rest.ErrNotInCluster {
+		return nil, fmt.Errorf("failed to get KubeClient: %v", err)
+	}
+	az, err := GetCloudProviderFromClient(kubeClient, secretName, secretNamespace, userAgent)
+	return az, err
+}
+
+// GetKubeConfig gets config object from config file
+func GetKubeConfig(kubeconfig string) (config *rest.Config, err error) {
```
andyzhangx, line 198: is this func `GetKubeConfig` needed? I thought only `GetKubeClient` is needed.

landreasyan: We need this function in our V2 implementation.

pkg/azureutils/azure_disk_utils.go
```diff
@@ -161,13 +151,13 @@ func GetCloudProvider(kubeconfig, secretName, secretNamespace, userAgent string)
...
 		if err != nil {
-			klog.Warningf("load azure config from file(%s) failed with %v", credFile, err)
-		} else {
-			defer credFileConfig.Close()
-			klog.V(2).Infof("read cloud config from file: %s successfully", credFile)
-			if config, err = azure.ParseConfig(credFileConfig); err != nil {
-				klog.Warningf("parse config file(%s) failed with error: %v", credFile, err)
-			}
+			klog.Errorf("Failed to load config from file: %s", credFile)
+			return nil, fmt.Errorf("Failed to load config from file: %s, cloud not get azure cloud provider", credFile)
```
edreed, line 155: The previous implementation would ignore the error to open the cloud config file and rely on the ultimate fallback of initializing using `az.InitializeCouldFromConfig()`. Why change this behavior? If there is a failure to parse and existing file, that is ignored in the following statements. It isn't clear to me why we would consider a missing or inaccessible file differently from an un-parsable one here.

pkg/azureutils/azure_disk_utils.go
mukhoakash: > @mukhoakash I think this was originally your change. I have kept it as an additional fall back scenario but was there any other reason for adding this?

Yes @landreasyan it was kept for easing the driver run from dev box env which @edreed mentioned.

pkg/azureutils/azure_disk_utils.go
landreasyan: This is the situation where `config==nil` and we fail to get the `config` from `config, err = azure.ParseConfig(credFileConfig)` due to missing cred file. The fall back to `az.InitializeCouldFromConfig()` only happens when `config !=nil` which does not cover this case. If we don't fail here azure.cloud is not initialized resulting in a null pointer exception later in the code.

landreasyan: This check is probably not necessary either: 
`if config == nil {
		klog.V(2).Infof("no cloud config provided, error: %v, driver will run without cloud config", err)
	}`

edreed: Ah, yes. I misread the code below. It's initializing the cloud provider from the config we load here, not supplying a default on failure.

andyzhangx (approved):
/lgtm
/approve

PR #1083, opened 2021-11-14
chore: MCR pipeline support for V2 driver build

edreed:
**What type of PR is this?**

/kind cleanup

The pipeline used to build and publish container images to the MCR supports the use of environment variables, but requires their value be used to customize the build and not just their presence. In order to support building both the V1 & V2 driver from the same pipeline, this PR changes the Makefile to rely on the value of the `BUILD_V2` environment variable to determine whether to build the V1 or V2 driver.

Although the V2 driver will be built from the `main_v2` branch, this change is necessary on master to correctly build the V1 driver and avoid inadvertently building and publish an incomplete V2 driver image from master when the pipeline definition is updated.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (approved):
/lgtm
/approve

PR #1110, opened 2021-12-11
bug: Adding supported VolumeCapabilites verification check to CSI RPC requests.

landreasyan:
**What type of PR is this?**

This PR adds a verification check for supported volume capabilities to CSI RPC requests.
Our driver doesn't support RWX access mode for filesystem volumes, yet we didn't have a specific check for preventing the use of such configuration.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -25,12 +25,10 @@ import (
 	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-12-01/compute"
-	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -41,7 +39,6 @@ import (
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 	consts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
-	"sigs.k8s.io/azuredisk-csi-driver/pkg/optimization"
 	volumehelper "sigs.k8s.io/azuredisk-csi-driver/pkg/util"
 	"sigs.k8s.io/cloud-provider-azure/pkg/metrics"
 	azure "sigs.k8s.io/cloud-provider-azure/pkg/provider"
@@ -61,7 +58,11 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		klog.Errorf("invalid create volume req: %v", req)
 		return nil, err
 	}
-
+	params := req.GetParameters()
+	diskParams, err := azureutils.ParseDiskParameters(params)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Failed parsing disk parameters: %v", err)
+	}
 	name := req.GetName()
 	if len(name) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "CreateVolume Name must be provided")
@@ -71,7 +72,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Error(codes.InvalidArgument, "CreateVolume Volume capabilities must be provided")
 	}
 
-	if !azureutils.IsValidVolumeCapabilities(volCaps) {
+	if !azureutils.IsValidVolumeCapabilities(volCaps, diskParams.MaxShares) {
 		return nil, status.Error(codes.InvalidArgument, "Volume capability not supported")
 	}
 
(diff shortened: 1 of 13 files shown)


andyzhangx (commented):
/ok-to-test

pkg/azureutils/azure_disk_utils.go
```diff
@@ -355,14 +355,28 @@ func IsValidDiskURI(diskURI string) error {
...
+	for _, c := range volCaps {
+		blockVolume := c.GetBlock()
+		mountVolume := c.GetMount()
+		accessMode := c.GetAccessMode().GetMode()
+
+		if (blockVolume == nil && mountVolume == nil) ||
+			(blockVolume != nil && mountVolume != nil) {
+			return false
+		}
+		if mountVolume != nil && accessMode == csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER {
```
andyzhangx, line 370: for file system, it does not support following 3 access modes:
```
	VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY VolumeCapability_AccessMode_Mode = 3
	VolumeCapability_AccessMode_MULTI_NODE_SINGLE_WRITER VolumeCapability_AccessMode_Mode = 4
	VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER VolumeCapability_AccessMode_Mode = 5
```

landreasyan: Oh, good to know! I'll add those as well.

test/e2e/testsuites/specs.go
```diff
@@ -201,12 +202,12 @@ func (pod *PodDetails) SetupStatefulset(client clientset.Interface, namespace *v
...
 	cleanupFuncs = append(cleanupFuncs, tsc.Cleanup)
 	ginkgo.By("setting up the PVC")
-	tpvc := NewTestPersistentVolumeClaim(client, namespace, volume.ClaimSize, volume.VolumeMode, &createdStorageClass)
+	tpvc := NewTestPersistentVolumeClaim(client, namespace, volume.ClaimSize, volume.VolumeMode, volume.VolumeAccessMode, &createdStorageClass)
 	storageClassName := ""
 	if tpvc.storageClass != nil {
 		storageClassName = tpvc.storageClass.Name
 	}
-	tpvc.requestedPersistentVolumeClaim = generateStatefulSetPVC(tpvc.namespace.Name, storageClassName, tpvc.claimSize, tpvc.volumeMode, tpvc.dataSource)
+	tpvc.requestedPersistentVolumeClaim = generatePVC(tpvc.namespace.Name, storageClassName, tpvc.claimSize, "pvc", tpvc.volumeMode, tpvc.accessMode, tpvc.dataSource)
```
edreed, line 210: Claim size and name ("pvc") order is reversed here.

landreasyan: Ah, yes! Thank you!

pkg/azuredisk/controllerserver.go
```diff
@@ -194,18 +198,6 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		}
 	}
 
-	if maxShares < 2 {
```
andyzhangx, line 197: I think the reasonable logic is like this:
if maxShares >=2 {
  allow muti-node read/write for raw block device
} 
always allow single-node read/write no matter it's shared disk or not.

landreasyan: I wanted to keep the volume capabilities check separate from maxshares values as we per [CSI documentation](https://kubernetes-csi.github.io/docs/raw-block.html?search=ValidateVolumeCapabilities) we need to validate volume capabilities in the following calls where we don't necessarily want to parse the maxshares values: 
CreateVolume() (multiple capabilities)
ControllerPublishVolume()
ValidateVolumeCapabilities() (multiple capabilities)
GetCapacity() (see below)
NodeStageVolume()
NodePublishVolume()

Do you think we should allow multi-node read/write for file system if the maxshares is < 2? That's going to be the only difference in the current and suggested implementation. 
We could replace the `IsValidVolumeCapabilities` with this logic here but then again, I think it makes it hard to maintain and keep the validation logic consistent.

andyzhangx: This PR actually changed the logic: it allows RWX with raw block device now (no matter it's shared disk or not), original design is that only shared disk would supports RWX with raw block device

landreasyan: If the disk is not shared, does allowing or not allowing RWX make any difference since sharing is already disabled? Alternatively, do you think we should be relying on maxshares for verifying volume capabilities in all the following functions?: 
CreateVolume() (multiple capabilities)
ControllerPublishVolume()
ValidateVolumeCapabilities() (multiple capabilities)
GetCapacity() (see below)
NodeStageVolume()
NodePublishVolume()

andyzhangx: if disk is not shared, we should not allow RWX. I think we need to keep original design, only shared disk would support RWX, RWO with raw block device, and yes, this driver should also rely on maxshares, and if `maxshares` is not provided by user, then it's not shared disk by default.

landreasyan: I have added maxshares parameter to the verification logic.

pkg/azuredisk/controllerserver.go
```diff
@@ -847,58 +760,37 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
...
-			newUserAgent := v
-			localCloud, err = azureutils.GetCloudProvider(d.kubeconfig, d.cloudConfigSecretName, d.cloudConfigSecretNamespace, newUserAgent, d.allowEmptyCloudConfig)
-			if err != nil {
-				return nil, fmt.Errorf("create cloud with UserAgent(%s) failed with: (%s)", newUserAgent, err)
-
-			}
-		default:
-			return nil, fmt.Errorf("AzureDisk - invalid option %s in VolumeSnapshotClass", k)
+	params := req.GetParameters()
+	diskParams, err := azureutils.ParseDiskParameters(params)
```
andyzhangx, line 766: could you only make changes to volume cap verification? `ParseDiskParameters` func implementation is mainly for CreateVolume, and it does not apply to CreateSnapshot and ExpandVolume since different CSI interface allows different parameters

pkg/azureutils/azure_disk_utils.go
```diff
@@ -426,6 +501,101 @@ func NormalizeStorageAccountType(storageAccountType, cloud string, disableAzureS
 	return "", fmt.Errorf("azureDisk - %s is not supported sku/storageaccounttype. Supported values are %s", storageAccountType, supportedSkuNames)
 }
 
+func ParseDiskParameters(parameters map[string]string) (DiskParameters, error) {
```
andyzhangx, line 504: this function could not be combined for CreateVolume, CreateSnapshot and ExpandVolume since different CSI interface accepts different parameters, cannot use one function to validate all parameters.

landreasyan: This function was called in CreateVolume and CreateSnapshot functions only which as far as I could tell were validating a subset of the same parameters. I have reverted the change to CreateSnapshot, so this parsing is now only used for CreateVolume.

andyzhangx, line 504: mainly LGTM, could you write a UT for this func, thanks.

andyzhangx (approved):
/lgtm
/approve

PR #1177, opened 2022-02-14
Changing the example from LRS to ZRS in order to showcase Zone Redundancy for managed disks.

naioja:
**What type of PR is this?**
/kind documentation

Fixes #
Uses the ZRS option for managed disks instead of LRS which was the default.

**Requirements**:
  
- [x] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/deploy/example/topology/README.md
+++ b/deploy/example/topology/README.md
@@ -20,7 +20,7 @@ metadata:
   name: managed-csi
 provisioner: disk.csi.azure.com
 parameters:
-  skuName: StandardSSD_LRS  # available values: Standard_LRS, Premium_LRS, StandardSSD_LRS, UltraSSD_LRS, Premium_ZRS, StandardSSD_ZRS
+  skuName: StandardSSD_ZRS  # available values: Standard_LRS, Premium_LRS, StandardSSD_LRS, UltraSSD_LRS, Premium_ZRS, StandardSSD_ZRS
 reclaimPolicy: Delete
 volumeBindingMode: Immediate
 ```


andyzhangx:
ZRS disk is available only in the West US 2, West Europe, North Europe, and France Central regions, so it's not good idea to make it default: https://docs.microsoft.com/en-us/azure/virtual-machines/disks-redundancy

naioja:
Indeed ZRS disks have a limited availability at the moment, but the yaml is used as an example how to have achieve workload flexibility where pods using PVC can being created in any of the availability zones and have their disks follow. 
In contrast using LRS when a PV/PVC is created in a specific zone for example AZ1 and then the POD that uses that PVC moves from the host in AZ1 to a host in AZ2 the pod creation will fail.
image
I thought the idea behind this example was to be able to have PVC and used them in any of the availability zones.

andyzhangx:
> Indeed ZRS disks have a limited availability at the moment, but the yaml is used as an example how to achieve workload flexibility where pods using PVC can being created in any of the availability zones and have their disks follow. In contrast using LRS when a PV/PVC is created in a specific zone for example AZ1 and then the POD that uses that PVC moves from the host in AZ1 to a host in AZ2 the pod creation will fail. image I thought the idea behind this example was to be able to have PVCs and used them in any of the availability zones, if that's not the case then I would draft another Storage Class to showcase ZRS disk capabilities.

@naioja I agree this is a bad example, would you just remove this example, and put the storage class example to the ZRS disk section?

naioja:
It seems we already have documentation/articles pointing to this repo : 
https://azure.microsoft.com/en-gb/blog/improve-availability-with-zoneredundant-storage-for-azure-disk-storage/ 
image
Should we just go ahead and create a new directory for zrs and showcase the SC there anyway ?

andyzhangx:
> It seems we already have documentation/articles pointing to this repo : https://azure.microsoft.com/en-gb/blog/improve-availability-with-zoneredundant-storage-for-azure-disk-storage/ image Should we just go ahead and create a new directory for zrs and showcase the SC there anyway ?

@naioja I have worked out a new PR: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1185, PTAL, thanks.

andyzhangx:
already addressed by https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1185

PR #1197, opened 2022-02-28
Skip volume expansion if block node

derek-pryor:
**What type of PR is this?**
/kind bug

No expansion on the node should be necessary if device is of type raw block

**Release note**:
```
Skip volume expansion if device is of type raw block
```


andyzhangx:
actually we need to rescan devices when volume expansion happens on block volume(only supported on v1.13 now), otherwise the driver could not get the expanded size after volume expansion:

https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/3b4f6445db686f23e1b39114b1fd9b5cb9ada878/pkg/azuredisk/nodeserver.go#L501-L509

derek-pryor:
@andyzhangx Thanks for that information.
PR not needed and I'll update our bug tracker.

derek-pryor:
@andyzhangx No, I was implementing a change that was/is being applied to many of the CSI drivers. Whoever opened the issue for Azure Disk didn't know that this change was not needed.

PR #1200, opened 2022-03-02
fix: GetNodeInfo should succeed when IMDS is unavailable

### LINKED ISSUE #1198: Instance metadata access from a Windows node
**What happened**:
The instance metadata endpoint can't be accessed from a Windows node because of the Windows container not running in the same network as the host, we have applied a workaround with an internal patch where we use the kube API server to fetch some of the info needed by CSI's `NodeGetInfo`, with this patch we went from not being able to fulfill the request (which previously showed these logs):

```
azuredisk-plugin I0218 00:03:35.165776    9572 utils.go:84] GRPC response: {"name":"disk.csi.azure.com","vendor_version":"v1.11.0"}
azuredisk-plugin I0218 00:03:36.911998    9572 utils.go:77] GRPC call: /csi.v1.Node/NodeGetInfo
azuredisk-plugin I0218 00:03:36.911998    9572 utils.go:78] GRPC request: {}
azuredisk-plugin W0218 00:03:57.924669    9572 nodeserver.go:334] get instance type(w4abj7whz000000) failed with: Get "http://169.254.169.254/metadata/instance?api-version=2019-03-11&format=json": dial tcp 169.254.169.254:80: conne
ctex: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
azuredisk-plugin W0218 00:04:18.951384    9572 nodeserver.go:358] get zone(w4abj7whz000000) failed with: Get "http://169.254.169.254/metadata/instance?api-version=2019-03-11&format=json": dial tcp 169.254.169.254:80: connectex: A c
onnection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
azuredisk-plugin I0218 00:04:18.951384    9572 nodeserver.go:387] not found a matching size in getMaxDataDiskCount, VM Size: , use default volume limit: 16
azuredisk-plugin I0218 00:04:18.951384    9572 utils.go:84] GRPC response: {"accessible_topology":{"segments":{"topology.disk.csi.azure.com/zone":""}},"max_volumes_per_node":16,"node_id":"w4abj7whz000000"}
```

To the patched version which is able to pull the necessary info from the kube API server

```
    azuredisk-plugin I0223 20:44:13.764467   14248 utils.go:77] GRPC call: /csi.v1.Node/NodeGetInfo
    azuredisk-plugin I0223 20:44:13.764467   14248 utils.go:78] GRPC request: {}
    azuredisk-plugin I0223 20:44:13.777684   14248 nodeserver.go:373] Attempting to get the Zone from the kube API server
    azuredisk-plugin I0223 20:44:13.777684   14248 nodeserver.go:386] NodeGetInfo, nodeName: w4abj7whz000000, failureDomain: eastus-1, region: eastus
    azuredisk-plugin I0223 20:44:13.777684   14248 nodeserver.go:400] Attempting to get the InstanceType from the kube API server
    azuredisk-plugin I0223 20:44:13.777684   14248 nodeserver.go:429] got a matching size in getMaxDataDiskCount, VM Size: STANDARD_B2S, MaxDataDiskCount: 4
    azuredisk-plugin I0223 20:44:13.777684   14248 utils.go:84] GRPC response: {"accessible_topology":{"segments":{"topology.disk.csi.azure.com/zone":"eastus-1","topology.kubernetes.io/zone":"eastus-1"}},"max_volumes_per_node":4,"node_
    id":"w4abj7whz000000"}
```

The patch that we have introduced internally conflicts with the new flag `--supportZone` added recently in https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1196

I'm going to create a dummy PR with the patch and link it to this issue, please let me know your thoughts

andyzhangx:
**What type of PR is this?**
/kind bug

fix: GetNodeInfo should succeed when IMDS is unavailable

`get-node-info-from-labels` flag is added, if it's set as `true`, only get zone and instance type from node labels, if it's `false`, get zone and instance type from IMDS first, and fall back to get zone and instance type from node labels if IMDS failed.

Fixes #1198

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: GetNodeInfo should succeed when IMDS is unavailable
```


test/external-e2e/run.sh
```diff
@@ -34,7 +34,7 @@ setup_e2e_binaries() {
     tar -xvf e2e-tests.tar.gz && rm e2e-tests.tar.gz
 
     # test on alternative driver name
-    export EXTRA_HELM_OPTIONS="--set controller.runOnMaster=true --set controller.replicas=1 --set driver.name=$DRIVER.csi.azure.com --set controller.name=csi-$DRIVER-controller --set linux.dsName=csi-$DRIVER-node --set windows.dsName=csi-$DRIVER-node-win"
+    export EXTRA_HELM_OPTIONS="--set controller.disableAvailabilitySetNodes=true --set controller.replicas=1 --set driver.name=$DRIVER.csi.azure.com --set controller.name=csi-$DRIVER-controller --set linux.dsName=csi-$DRIVER-node --set windows.dsName=csi-$DRIVER-node-win"
```
mauriciopoppe, line 37: is `controller.disableAvailabilitySetNodes=true` part of #1198?

pkg/azuredisk/nodeserver.go
```diff
@@ -324,6 +324,11 @@ func (d *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRequest) (
 		Segments: map[string]string{topologyKey: ""},
 	}
 
+	failureDomain, instanceType, err := getNodeInfoFromAPIServer(ctx, d.NodeID, d.cloud.KubeClient)
```
mauriciopoppe, line 327: this impl is always trying to get the API server, in the patch I sent I think it was behind a flag

pkg/azuredisk/nodeserver.go
```diff
@@ -335,25 +340,28 @@ func (d *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRequest) (
 			zone, zoneError = d.cloud.GetZone(ctx)
```
mauriciopoppe, line 340: /hold

if `d.supportZone` is true and we're in a Windows node that doesn't have access to the metadata server then we'll always attempt to make this request (which times out but after some time), this means that the NodeGetInfo call will not succeed after some time has passed.

In the patch I made sure that if we get info from the API server that we no longer make requests to the metadata server, therefore NodeGetInfo succeeded with a single request

Setting `d.supportZone` to false through a cli flag would skip this block but the impl above is already trying to fetch info from the metadata server anyways

Is there a case where we would need `d.supportZone` given that the node component can get it from the kube API server?

pkg/azuredisk/nodeserver.go
```diff
@@ -363,10 +371,13 @@ func (d *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRequest) (
 				return nil, status.Error(codes.Internal, "failed to get instances from cloud provider")
```
mauriciopoppe, line 371: same comment as https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1200/files#r817990940

test/external-e2e/run.sh
andyzhangx: this is another commit: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1200/commits/d2084568588093c749378ad0dadc9fd0e3a95ea9, you could ignore this

pkg/azuredisk/nodeserver.go
andyzhangx: from the logs, I could see after timeout, it could proceed in `NodeGetInfo`, since `NodeGetInfo` is one time operation, I would suggest always get zone from instance metadata first (it's always accurate), and if failed with timeout, then get zone info from node labels.

```
azuredisk-plugin I0218 00:03:36.911998    9572 utils.go:77] GRPC call: /csi.v1.Node/NodeGetInfo
azuredisk-plugin I0218 00:03:36.911998    9572 utils.go:78] GRPC request: {}
azuredisk-plugin W0218 00:03:57.924669    9572 nodeserver.go:334] get instance type(w4abj7whz000000) failed with: Get "http://169.254.169.254/metadata/instance?api-version=2019-03-11&format=json": dial tcp 169.254.169.254:80: conne
ctex: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
azuredisk-plugin W0218 00:04:18.951384    9572 nodeserver.go:358] get zone(w4abj7whz000000) failed with: Get "http://169.254.169.254/metadata/instance?api-version=2019-03-11&format=json": dial tcp 169.254.169.254:80: connectex: A c
onnection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
azuredisk-plugin I0218 00:04:18.951384    9572 nodeserver.go:387] not found a matching size in getMaxDataDiskCount, VM Size: , use default volume limit: 16
```

pkg/azuredisk/nodeserver.go
andyzhangx: what about this? if get failureDomain and instanceType failed from instance metadata, then we always fall back to get from apiserver? use a flag to control this behavior is difficult to use (always need manual config). And even get from instance metadata timeout, it would proceed to get from apiserver.

mauriciopoppe: > if get failureDomain and instanceType failed from instance metadata

If it's not available then the call will block for some time before it times out and continues, it'd be great if we could instead continue immediately if we detected that the metadata server is unavailable, my concern is with the time NodeGetInfo takes before it succeeds

pkg/azuredisk/nodeserver.go
mauriciopoppe: Please look at https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1200#discussion_r818392423, the concern is the blocking call, in the example it took around 40s to timeout (I guess 20s per blocking call) before continuing

pkg/azuredisk/nodeserver.go
andyzhangx: sure, I will change the logic a little, and also I found this requires nodes get role: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1200/commits/f2014c514fb7e3e6952d8502b1459145a38a39de

pkg/azuredisk/nodeserver.go
andyzhangx: already added `get-node-info-from-labels` flag, if it's set, then only get node info from node labels. PTAL.

pkg/azuredisk/nodeserver.go
```diff
@@ -326,20 +326,28 @@ func (d *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRequest) (
...
-			zone      cloudprovider.Zone
-			zoneError error
+			zone cloudprovider.Zone
+			err  error
 		)
-		if runtime.GOOS == "windows" && (!d.cloud.UseInstanceMetadata || d.cloud.Metadata == nil) {
-			zone, zoneError = d.cloud.VMSet.GetZoneByNodeName(d.NodeID)
+
+		if d.getNodeInfoFromLabels {
+			zone.FailureDomain, _, err = getNodeInfoFromLabels(ctx, d.NodeID, d.cloud.KubeClient)
```
mauriciopoppe, line 334: This is a nit that was noticed by @msau42 while reviewing the internal PR, we're making 2 calls to the API server, in the proposed WIP PR I think I made a call to it once and with the returned node object I the properties needed e.g. the failure domain label here and the instance type later.

An additional call to the API server is not a concern though.

pkg/azuredisk/nodeserver.go
```diff
@@ -349,23 +357,34 @@ func (d *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRequest) (
...
+				} else {
+					instanceType, err = instances.InstanceType(ctx, types.NodeName(d.NodeID))
+				}
 			}
-			var err error
-			if instanceType, err = instances.InstanceType(ctx, types.NodeName(d.NodeID)); err != nil {
+			if err != nil {
 				klog.Warningf("get instance type(%s) failed with: %v", d.NodeID, err)
 			}
+			if instanceType == "" {
```
mauriciopoppe, line 381: this is interesting, in which scenarios would this if block execute? I guess if `d.getNodeInfoFromLabels` is false and somehow the other instance metadata requests timed out right?

andyzhangx: yes, when instance metadata requests timed out, fall back to get instance type from node labels

pkg/azuredisk/nodeserver.go
andyzhangx: with https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1200/commits/8bfdfe4db6cb0937391070fadaf256da10cbd489, there is only one call to api server now, PTAL.

pkg/azurediskplugin/main.go
```diff
@@ -45,6 +45,7 @@ var (
 	driverName                 = flag.String("drivername", consts.DefaultDriverName, "name of the driver")
 	volumeAttachLimit          = flag.Int64("volume-attach-limit", -1, "maximum number of attachable volumes per node")
 	supportZone                = flag.Bool("support-zone", true, "boolean flag to get zone info in NodeGetInfo")
+	getNodeInfoFromLabels      = flag.Bool("get-node-info-from-labels", false, "boolean flag to get zone info from node labels in NodeGetInfo")
```
mauriciopoppe, line 48: I'm wondering if this flag should be set to true in the Windows image cli flags too

andyzhangx: I don't think so, we already have fall back in this PR which could ensure if IMDS does not work, get zone and instance type still works. I would work out another PR to support this setting in helm chart install.

pkg/azuredisk/nodeserver.go
```diff
@@ -324,22 +324,32 @@ func (d *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRequest) (
...
+	var err error
+
 	if d.supportZone {
-		var (
-			zone      cloudprovider.Zone
-			zoneError error
-		)
-		if runtime.GOOS == "windows" && (!d.cloud.UseInstanceMetadata || d.cloud.Metadata == nil) {
-			zone, zoneError = d.cloud.VMSet.GetZoneByNodeName(d.NodeID)
+		var zone cloudprovider.Zone
```
mauriciopoppe, line 331: I believe that the code is correct however I think it's a little bit hard to read, it might be a good idea to refactor this into a function so that we can return the zone as soon as it's available (same thing for the instance type), I don't think I have a concern about the logic though.

Thanks for making the change.

pkg/azuredisk/nodeserver.go
mauriciopoppe: thanks for making the changes, I added this comment https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1200#discussion_r819297469 and I think that the logic is good.

/remove-hold

PR #1201, opened 2022-03-03
WIP: Run Windows containers as host-process containers

marosset:
**What type of PR is this?**

/kind feature

This PR allows for running the Windows azuredisk csi driver containers as host-process containers.
This simplifies the deployment of the drivers and aligns with the long-term CSI plugin goals of the Kubernetes community.

This also works around https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/2132 allowing azuredisk csi drivers to be used in CAPZ clusters.

Note: this still requires csi-proxy to be installed and running on Windows nodes in the cluster

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```

```


marosset:
I will work on updating the helm charts now too.

marosset:
It looks like `pull-azuredisk-csi-driver-e2e-windows-containerd` is using v1.20 bits for the nodes because 1.20 is no longer supported.

This should probably be updated...

```
2022/03/10 22:06:26 e2e.go:496: Cluster nodes:
6144k8s000              Ready   agent    43s   v1.20.15
6144k8s001              Ready   <none>   15s   v1.20.15
k8s-master-61449729-0   Ready   master   90s   v1.20.15
2022/03/10 22:06:26 process.go:153: Running: kubectl --match-server-version=false version
WARNING: version difference between client (1.23) and server (1.20) exceeds the supported minor version skew of +/-1
```

marosset:
Let me fix the verify job and then squash all my commits.

jsturtevant:
I don't have a ton of experience with the CSI drivers but generally lgtm and the test `pull-kubernetes-e2e-capz-azure-disk-windows` is passing using this template.

marosset:
@andyzhangx it looks like `pull-kubernetes-e2e-capz-azure-disk` is timing out running windows test cases 

ex: `   Should create and attach a volume with basic perfProfile [enableBursting][disk.csi.azure.com] [Windows] [It]`

Do we need to set ginkgo.skip somewhere?

andyzhangx:
> @andyzhangx it looks like `pull-kubernetes-e2e-capz-azure-disk` is timing out running windows test cases
> 
> ex: ` Should create and attach a volume with basic perfProfile [enableBursting][disk.csi.azure.com] [Windows] [It]`
> 
> Do we need to set ginkgo.skip somewhere?

@marosset that's zone issue:
```
"Disk /subscriptions/0e46bd28-a80f-4d3a-8200-d9eb8d80cb2e/resourceGroups/CAPZ-9B3W2C/providers/Microsoft.Compute/disks/reattach-disk-multiple-nodes cannot be attached to the VM because it is not in zone '1'."
```

currently master node is in zone 1, while agent node are non-zone, the test case would create a PV disk first with zone 1, and then it could not be attached to non-zone node, I will check how to fix it later

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node-windows-hostprocess.yaml
```diff
@@ -0,0 +1,161 @@
...
+        nodeAffinity:
+{{ toYaml .Values.windows.nodeAffinity | indent 10 }}
+      priorityClassName: system-node-critical
+      {{- include "azuredisk.pullSecrets" . | indent 6 }}
+      securityContext:
+        windowsOptions:
+          hostProcess: true
+          runAsUserName: "NT AUTHORITY\\SYSTEM"
+      hostNetwork: true
+      initContainers:
```
andyzhangx, line 41: is it possible to combine these settings with original non-hostprocess deployment? looks like the main change is `initContainers`, otherwise we need to maintain two copies of windows daemonset in long term.

marosset: Are you suggesting to update the csi-azuredisk-node-windows.yaml deployment use host-process and not introduce another file?

I don't think AKS supports host-process containers right now but if we are OK with saying only older existing versions
of the driver work on AKS I can do this.

I was thinking we could have both types of deployments exist at the same time for a while in case we find any issues with the host-process container deployments and then delete them at a later time.
Let me know what you think.

andyzhangx: correct, update `charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node-windows.yaml` since chart config supports if-else, that would be easier to maintain in the future. And for `deploy/csi-azuredisk-node-windows.yaml`, looks like we have to create a new file for host-process containers

marosset: Sounds good. I will do this today.

marosset: I merged the two windows node helm template files - PTAL.
I left them in a sperate commit in case reviewer do not like the changes (I think it makes the helm templates a little bit more complicated).

marosset: Hmm, it looks like there is a check to ensure the number of .yaml files in the chart template directory matches the number of files in /deploy

```
yaml file num(10) under deploy/ not equal to num(9) under charts/latest/azuredisk-csi-driver/templates
make: *** [Makefile:85: verify] Error 1
```

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node-windows.yaml
```diff
@@ -33,16 +33,52 @@ spec:
...
         - name: liveness-probe
-          volumeMounts:
-            - mountPath: C:\csi
-              name: plugin-dir
 {{- if hasPrefix "/" .Values.image.livenessProbe.repository }}
           image: "{{ .Values.image.baseRepo }}{{ .Values.image.livenessProbe.repository }}:{{ .Values.image.livenessProbe.tag }}"
 {{- else }}
           image: "{{ .Values.image.livenessProbe.repository }}:{{ .Values.image.livenessProbe.tag }}"
 {{- end }}
+          imagePullPolicy: {{ .Values.image.livenessProbe.pullPolicy }}
```
andyzhangx, line 69: I thought it was a small diff, while it looks like these two deployments are quite different now, then I think we should keep two deployment files, otherwise it may be hard to understand and maintain.

marosset: No problem. I dropped the commit to split merge the manifests.
Let me rebase again.
The manifest do have many subtle differences.

marosset:
@andyzhangx do you have any other concerns or feedback here?
I'd like to try get CAPZ + azuredisk + Windows periodic jobs enabled.

andyzhangx (approved):
/lgtm
/approve

PR #1204, opened 2022-03-04
feat: add getNodeInfoFromLabels in chart config

andyzhangx:
**What type of PR is this?**
/kind feature

feat: add getNodeInfoFromLabels in chart config

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
feat: add getNodeInfoFromLabels in chart config
```


andyzhangx:
cc @mauriciopoppe

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node-windows.yaml
```diff
@@ -108,6 +108,7 @@ spec:
             - "--user-agent-suffix={{ .Values.driver.userAgentSuffix }}"
             - "--allow-empty-cloud-config={{ .Values.node.allowEmptyCloudConfig }}"
             - "--support-zone={{ .Values.node.supportZone }}"
+            - "--get-node-info-from-labels={{ .Values.windows.getNodeInfoFromLabels }}"
```
mauriciopoppe, line 111: This is the location where I was thinking that the value should be true, but I guess it depends on the setup and on having a component that can act as an intermediary to make the instance metadata calls (e.g. a container that intercepts the request and forwards it to the host binary that does it on behalf of the CSI Driver)

mauriciopoppe (commented):
should the files in `deploy/` be updated too?

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node-windows.yaml
andyzhangx: if you are using helm chart install, you could add `--set windows.getNodeInfoFromLabels=true`, while it's better not change default value as true here

andyzhangx:
> should the files in `deploy/` be updated too?

@mauriciopoppe fixed.

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node-windows.yaml
mauriciopoppe: sounds good, thanks

deploy/csi-azuredisk-node.yaml
```diff
@@ -91,6 +91,7 @@ spec:
             - "--metrics-address=0.0.0.0:29605"
             - "--enable-perf-optimization=true"
             - "--allow-empty-cloud-config=true"
+            - "--get-node-info-from-labels=false"
```
mauriciopoppe, line 94: mmm I thought these files were generated by running the chart against values and dumping them here, it looks like they're updated manually because this file doesn't have the other flags in the chart.

if we want to deploy a release which files should we use?

- the files in `deploy/<release>`? if so, are these files the ones tested?
- the result of applying some values to the chart from the tag <release>? if so, are these files the ones tested?

andyzhangx: there are two ways to install this driver:
 - kubectl apply `deploy/<release>`
 - helm chart install
pls always use helm chart install for customization: https://github.com/kubernetes-sigs/azuredisk-csi-driver/tree/master/charts#v1-parameters

mauriciopoppe: understood, thanks

mauriciopoppe:
/lgtm
/hold

Feel free to remove the hold, pull-kubernetes-e2e-capz-azure-disk-windows didn't succeed yet

PR #1220, opened 2022-03-14
[V2] test: add test for unplanned failover scenario

sunpa93:
**What type of PR is this?**
/kind test

* adds test that emulates unplanned  failover scenario for v2 driver
* fixes test cases with invalid assumptions

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/test/e2e/dynamic_provisioning_test.go
+++ b/test/e2e/dynamic_provisioning_test.go
@@ -48,22 +48,21 @@ var _ = ginkgo.Describe("Dynamic Provisioning", func() {
 	ginkgo.Context("[multi-az]", func() {
 		t.defineTests(true, scheduler)
 	})
+
+	ginkgo.Context("[unplanned]", func() {
+		t.defineUnplannedTests(false, "csi-azuredisk-scheduler-extender")
+	})
 })
 
 type dynamicProvisioningTestSuite struct {
+	cs                    clientset.Interface
+	ns                    *v1.Namespace
+	snapshotrcs           restclientset.Interface
+	testDriver            driver.PVTestDriver
 	allowedTopologyValues []string
 }
 
-func (t *dynamicProvisioningTestSuite) defineTests(isMultiZone bool, schedulerName string) {
-	f := framework.NewDefaultFramework("azuredisk")
-
-	var (
-		cs          clientset.Interface
-		ns          *v1.Namespace
-		snapshotrcs restclientset.Interface
-		testDriver  driver.PVTestDriver
-	)
-
+func (t *dynamicProvisioningTestSuite) beforeEach(f *framework.Framework, isMultiZone bool) {
 	ginkgo.BeforeEach(func() {
 		checkPodsRestart := testutil.TestCmd{
 			Command:  "bash",
@@ -73,18 +72,18 @@ func (t *dynamicProvisioningTestSuite) defineTests(isMultiZone bool, schedulerNa
 		}
 		testutil.ExecTestCmd([]testutil.TestCmd{checkPodsRestart})
 
-		cs = f.ClientSet
-		ns = f.Namespace
+		t.cs = f.ClientSet
+		t.ns = f.Namespace
 
 		var err error
-		snapshotrcs, err = testutil.RestClient(testconsts.SnapshotAPIGroup, testconsts.APIVersionv1)
+		t.snapshotrcs, err = testutil.RestClient(testconsts.SnapshotAPIGroup, testconsts.APIVersionv1)
 		if err != nil {
 			ginkgo.Fail(fmt.Sprintf("could not get rest clientset: %v", err))
 		}
 
 		// Populate allowedTopologyValues from node labels fior the first time
 		if isMultiZone && len(t.allowedTopologyValues) == 0 {
-			nodes, err := cs.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+			nodes, err := t.cs.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
 			framework.ExpectNoError(err)
 			allowedTopologyValuesMap := make(map[string]bool)
 			for _, node := range nodes.Items {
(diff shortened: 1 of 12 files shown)


edreed:
Instead of a separate test, could we do this as part of the E2E tests by adding support for this scenario to `PodFailoverWithReplicas`?

sunpa93:
> Instead of a separate test, could we do this as part of the E2E tests by adding support for this scenario to PodFailoverWithReplicas?

I was originally hesitant to add this to e2e test suite because e2e test runs very often and this test case by itself would consume ~15 minutes to complete but if this is no issue, I can go ahead and move this to be a part of the e2e test suite.

edreed:
> I was originally hesitant to add this to e2e test suite because e2e test runs very often and this test case by itself would consume ~15 minutes to complete but if this is no issue, I can go ahead and move this to be a part of the e2e test suite.

That is something to consider. We could use a separate context to prevent running by default and only add the context in to the test run that we want.

jmclong:
I tested out the changes for TestStatefulSet here, and it fixes the issue with AllPods not being populated when calling the Create method. I'll wait for this change prior submitting a fix for #1244.

edreed:
The Trivy failures are fixed by #1293 .

test/e2e/dynamic_provisioning_test.go
```diff
@@ -1652,7 +1725,7 @@ func (t *dynamicProvisioningTestSuite) defineTests(isMultiZone bool, schedulerNa
 		storageClassParameters := map[string]string{"skuName": skuName, "maxShares": "3", "cachingMode": "None"}
```
jmclong, line 1726: Hey,

I was looking at the test failure that you encountered, and the symptom seems like it was pretty similar to an issue I was facing earlier. I think that you need to set `VolumeBindingMode` to immediate, otherwise polling for the `multi-az` case doesn't work. The `PersistentVolume` for the volumes of that pod will be `nil`. Here's a snippet of the code that I had written to do that:



```go
if testutil.IsZRSSupported(location) {
	immediate := storagev1.VolumeBindingImmediate
	for i := range pod.Volumes {
		pod.Volumes[i].VolumeBindingMode = &immediate
	}
}
```

You have to add this in after the call to `NormalizeVolumes` because that changes the volume binding mode on the volumes of the pod. 

If this is indeed the issue, we might want to take a look at usage of `VerifySuccessfulAzVolumeAttachments` after this PR to see if the ```if volume.PersistentVolume != nil``` inside of it should actually be an error condition because it's tricky behavior. 

Joshua

sunpa93: Hey Joshua. Sorry for the delay; I was addressing problems from other PR yesterday and wasn't able to get to this promptly.

And the problem indeed seems to be caused because of volume binding mode being set to `WaitForFirstCustomer` when it should be `VolumeBindingImmediate` as you suggested. For ZRS disks, `GetDynamicProvisioningStorageClass` function is responsible for setting the volume binding mode to `VolumeBindingImmediate`.
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/04aba1e20ce3598d77e359198e4579335eb27024/test/e2e/driver/azuredisk_driver.go#L62-L64

The problem seems to be that the `SkuNameField` key in the storage class parameter is hard coded and is in camel case when it should not be (`SkuNameField` defined in azure constants package is all lower cased).
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/04aba1e20ce3598d77e359198e4579335eb27024/test/e2e/dynamic_provisioning_test.go#L1652

Actually, the `SkuNameField` key value seems to be inconsistent throughout the entire test suite (some are camel cased and some aren't). So I have raised a PR (https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1305) to address this as this make it difficult to detect bugs.

jmclong: Thanks for finding the root cause!

jmclong: I'm still seeing this issue after correcting the SKU field key. It seems that `NormalizeVolumes` sets the Volume's binding mode to `storagev1.VolumeBindingWaitForFirstConsumer` in the multi-az case. This results in the following check failing and the `PersistentVolume` field of `tpvc` never being set:

https://github.com/jmclong/azuredisk-csi-driver/blob/88123cd850a0082c59ba016d7eccdc04223a0f8e/test/resources/volumes.go#L101:L101

sunpa93: That's probably because `SetUpDynamicPersistentVolumeClaim` relies on `VolumeDetail.BindingMode` instead of correctly set storage class's binding mode.
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/04aba1e20ce3598d77e359198e4579335eb27024/test/resources/volumes.go#L101

sunpa93: I just made the abovementioned changes and added it to the https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1305 as a supplemental change.

And I have made a few changes in this PR to `VerifySuccessfulAzVolumeAttachments`, `SetupWithDynamicVolumes`, and `SetupWithDynamicMultipleVolumes` to address issues with `PodDetails.Volumes[n].PersistentVolume` being `nil`, which I will be pushing shortly.

jmclong: Nice! I was about to validate that solution. (using `storageClass.VolumeBindingMode` instead of `volume.VolumeBindingMode`).

sunpa93:
Test results dependency on https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1305

sunpa93:
This PR closed until https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1305 gets merged due to dependency.

PR #1216, opened 2022-03-11
[V2] chore: Merge master to main v2

mukhoakash:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:
- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #1237, opened 2022-03-23
[V2] chore: Allow for sync or async AttachDisk operation

edreed:
**What type of PR is this?**

/kind cleanup

Brings the V2 driver up to parity with allowing for synchronous or asynchronous `AttachDisk` operation.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/provisioner/cloudprovisioner.go
```diff
@@ -340,7 +340,12 @@ func (c *CloudProvisioner) PublishVolume(
 			return nil, err
 		}
 
-		lun, err = c.cloud.AttachDisk(ctx, true, diskName, volumeID, nodeName, cachingMode, disk)
+		asyncAttach := true
```
nearora-msft, line 343: Should we update the driver-parameters.md file to specify  that the default value of this parameter for V2 driver is true?

edreed: This PR is more or less so we can run the scale tests with synchronous `AttachDisk` to see its impact on the scale-out times. The full (and updated) change for this support in V1 will be brought in as part of #1216 .

nearora-msft: If this is just for testing, will this be reverted after #1216 is merged?

edreed: We'll still need the updated `cloud-provisioner-azure` brought in by this change to enable sync/async operation via configuration to maintain feature parity with the v1 driver.

nearora-msft (approved):

pkg/provisioner/cloudprovisioner.go
edreed: I updated the V2 driver parameter documentation.

PR #1244, opened 2022-03-25
[V2] feat: add replica creation after volumes are garbage collected 

kassarl:
**What type of PR is this?**

/kind feature
After volumes are garbage collected and attachment capacity opens up on a node, prompts an attempt to create previously failed replicas

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/replica.go
+++ b/pkg/controller/replica.go
@@ -21,11 +21,15 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	diskv1beta1 "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/azuredisk/v1beta1"
 	consts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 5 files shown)


sunpa93 (changes requested):

pkg/controller/replica.go
```diff
@@ -171,6 +175,34 @@ func (r *ReconcileReplica) triggerGarbageCollection(volumeName string) {
...
+
+	go func(ctx context.Context) {
+		err := wait.PollImmediate(deletionPollingInterval, 10*time.Minute, func() (bool, error) {
+			volRequirement, err := azureutils.CreateLabelRequirements(consts.VolumeNameLabel, selection.Equals, volumeName)
+			if err != nil {
+				return false, err
+			}
+			labelSelector := labels.NewSelector().Add(*volRequirement)
+
+			_, err = r.controllerSharedState.azClient.DiskV1beta1().AzVolumeAttachments(r.controllerSharedState.objectNamespace).List(ctx, metav1.ListOptions{LabelSelector: labelSelector.String()})
```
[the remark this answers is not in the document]
sunpa93: As this is in a wait loop, stale cache shouldn't be a problem. Cache will eventually be in the right state (i.e. AzVolumeAttachments for the volume all deleted), at which point controller can recreate failed replica attachments.

pkg/controller/replica.go
```diff
@@ -171,6 +175,34 @@ func (r *ReconcileReplica) triggerGarbageCollection(volumeName string) {
...
+			if err != nil {
+				if apiErrors.IsNotFound(err) {
+					return true, nil
+				}
+				klog.Errorf("failed to get AzVolumeAttachments: %v", err)
+				return false, err
+			}
+			return false, nil
```
sunpa93, line 195: We should return true if `len(AzVolumeAttachmentList.Items) == 0`

pkg/controller/replica.go
```diff
@@ -171,6 +175,37 @@ func (r *ReconcileReplica) triggerGarbageCollection(volumeName string) {
...
+				if apiErrors.IsNotFound(err) {
+					return true, nil
+				}
+				klog.Errorf("failed to get AzVolumeAttachments: %v", err)
+				return false, err
+			}
+			if len(azVolumeAttachmentList.Items) == 0 {
+				return true, nil
+			}
+			return false, nil
```
sunpa93, line 198: np: we can probably just do `return len(azVolumeAttachmentList.Items) == 0, nil` instead of having a separate if clause for this.

nearora-msft (approved):

jmclong:
The check for the stateful set fails because the` AllPod` property is not populated by the `Create` method. This is currently fixed as a part of #1220. 

Additionally, the `NewPod` created by the test will be created with a volume with `VolumeBindingMode` set to `VolumeBindingWaitForFirstConsumer` for the multi-az test. This results in the `PersistentVolume` of the volume of `NewPod` being set to `nil`. The `VerifySuccessfulAzVolumeAttachments` method will see that `PersistentVolume` is `nil` without checking to see if the primary and replica attachments are successful. This causes the expected "failure" to pass and the test to fail for the multi-az case. Other tests change the `VolumeBindingMode` to `VolumeBindingImmediate`, so that may be an option to fix that issue.

Finally, I noticed that the test will fail if you run it multiple times. This is because ./hack/verify_examples.sh is called after the test suite. It creates a stateful set and deployment that are not cleaned up. This won't interfere with the test when running in the cluster, because it runs after the suite, but can be confusing when running the test interactively.

jmclong:
Hi @edreed,

Please close this pull request, as I do not have write-access to the branch.

I will create a new one, as #1305 has now been merged into main_v2.

Thanks!

edreed:
This PR will be replaced by a new updated PR from @jmclong .

PR #1245, opened 2022-03-25
[V2] chore: Adding flags for qps and disabling listvolume by default.

landreasyan:
**What type of PR is this?**

Increasing the default rest client QPS and disabling listVolume capability by default to optimize performance of the driver.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -57,6 +57,9 @@ type DriverOptions struct {
 	UseCSIProxyGAInterface     bool
 	EnableDiskOnlineResize     bool
 	AllowEmptyCloudConfig      bool
+	EnableListVolumes          bool
+	EnableListSnapshots        bool
+	RestClientQPS              int
 }
 
 // CSIDriver defines the interface for a CSI driver.
@@ -91,6 +94,8 @@ type DriverCore struct {
 	useCSIProxyGAInterface     bool
 	enableDiskOnlineResize     bool
 	allowEmptyCloudConfig      bool
+	enableListVolumes          bool
+	enableListSnapshots        bool
 }
 
 // Driver is the v1 implementation of the Azure Disk CSI Driver.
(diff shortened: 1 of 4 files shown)


sunpa93 (approved):
lgtm

pkg/azuredisk/azuredisk.go
```diff
@@ -57,6 +57,9 @@ type DriverOptions struct {
 	UseCSIProxyGAInterface     bool
 	EnableDiskOnlineResize     bool
 	AllowEmptyCloudConfig      bool
+	EnableListVolumes          bool
+	EnableListSnapshots        bool
+	RestClientQps              int
```
sunpa93, line 62: apparently this has to be spelled QPS

landreasyan: Yes, you are right. For some reason I thought go was always mixedCaps even for acronyms but apparently those should have a consistent case.

pkg/azurediskplugin/main.go
```diff
@@ -53,6 +53,9 @@ var (
 	useCSIProxyGAInterface     = flag.Bool("use-csiproxy-ga-interface", true, "boolean flag to enable csi-proxy GA interface on Windows")
 	enableDiskOnlineResize     = flag.Bool("enable-disk-online-resize", true, "boolean flag to enable disk online resize")
 	allowEmptyCloudConfig      = flag.Bool("allow-empty-cloud-config", true, "Whether allow running driver without cloud config")
+	enableListVolumes          = flag.Bool("enable-list-volumes", false, "boolean flag to enable ListVolumes on controller")
+	enableListSnapshots        = flag.Bool("enable-list-snapshots", false, "boolean flag to enable ListSnapshots on controller")
+	restClientQPS              = flag.Int("client-qps", 15, "QPS for the rest client. Defaults to 15.")
```
edreed, line 58: Let's use `kubeClientQPS` and `kube-client-qps` to more clearly indicate this is for K8s API access.

PR #1247, opened 2022-03-26
[V2] fix: Avoid calling detach for ControllerUnpublish Volume

nearora-msft:
**What type of PR is this?**
/kind bug

Issue: When a ControllerPublishVolume call comes for a replica attachment, it is moved to demoted state. If another ControllerPublishVolume  comes for the same attachment, the current implementation deletes the replica attachment because of this condition that whether the volume attachment is a replica or not becomes true here the 2nd time around: https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/467066e583ee720abc58b66a95d3332269753112/pkg/provisioner/crdprovisioner.go#L631

Fix: Remove this condition from the UnpublishVolume function and create a separate function for deatching the volume

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/provisioner/crdprovisioner.go
+++ b/pkg/provisioner/crdprovisioner.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/klog/v2"
 	diskv1beta1 "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/azuredisk/v1beta1"
 	azDiskClientSet "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/client/clientset/versioned"
+	"sigs.k8s.io/azuredisk-csi-driver/pkg/apis/client/clientset/versioned/typed/azuredisk/v1beta1"
 	azurediskInformers "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/client/informers/externalversions"
 	consts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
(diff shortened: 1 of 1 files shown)


edreed:
@sunpa93 Could you please take a look at this change?

pkg/provisioner/crdprovisioner.go
```diff
@@ -627,61 +628,74 @@ func (c *CrdProvisioner) UnpublishVolume(
...
+	if azVolumeInstance.Spec.MaxMountReplicaCount == 0 {
 		// if volume's maxMountReplicaCount == 0, detach and wait for detachment to complete
-		// if AzVolumeAttachment instance indicates that previous attachment request was successful, annotate the CRI with detach request so that the underlying volume attachment can be properly detached.
-		if azVolumeAttachmentInstance.Status.Detail != nil {
-			// if detachment is pending, return to prevent duplicate request
-			if azVolumeAttachmentInstance.Status.State == diskv1beta1.Detaching {
-				return status.Errorf(codes.Aborted, "detachment still in process for volume (%s) and node (%s)", volumeName, nodeID)
-			}
+		return c.detachVolume(ctx, azVAClient, azVolumeAttachmentInstance)
+	}
```
sunpa93, line 636: I think it would be more consistent to refactor this into a separate `demote` function.

PR #1250, opened 2022-03-30
[V2] bug: UnpublishVolume should not delete the CRI without detaching

sunpa93:
**What type of PR is this?**
/kind bug

* With the asynchronous attach feature, it is possible for `ControllerUnpublishVolume` request to be issued while the volume attachment operation is happening asynchronously in the background.
* In this case, the `AzVolumeAttachment.Status.Detail` will not be populated and therefore, the `CrdProvisioner` will make a deletion request for the CRI without making a detachment request.
* In which case, the volume attachment gets lost and its associated volume will not be able to be deleted.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/provisioner/crdprovisioner.go
+++ b/pkg/provisioner/crdprovisioner.go
@@ -658,20 +658,30 @@ func (c *CrdProvisioner) detachVolume(ctx context.Context, azVAClient v1beta1.Az
 		return err
 	}
 
-	if azVolumeAttachment.Status.Detail != nil {
-		// if detachment is pending, return to prevent duplicate request
-		if azVolumeAttachment.Status.State == diskv1beta1.Detaching {
-			return status.Errorf(codes.Aborted, "detachment still in process for volume (%s) and node (%s)", volumeName, nodeName)
+	updateFunc := func(obj interface{}) error {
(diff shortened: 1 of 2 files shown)


pkg/provisioner/crdprovisioner.go
```diff
@@ -658,20 +658,29 @@ func (c *CrdProvisioner) detachVolume(ctx context.Context, azVAClient v1beta1.Az
...
+			return err
 		}
-
-		klog.Infof("Requesting AzVolumeAttachment (%s) deletion", attachmentName)
-
-		updateFunc := func(obj interface{}) error {
+	case diskv1beta1.DetachmentFailed:
+		// if detachment failed, reset error and state to retrigger operation
+		updateFunc = func(obj interface{}) error {
+			_ = updateFunc(obj)
```
edreed, line 679: Won't this end up being a recursive call since the closure captures `updateFunc` by reference and not by value?

sunpa93: Oh, you are right. Good catch :)

PR #1257, opened 2022-03-31
[V2] chore: make CRD status a sub-resource

sunpa93:
**What type of PR is this?**
/kind cleanup

* Currently, CRD status is not a sub-resource and making it into a subresource would ideally allow us to avoid granting too much permission to services that only require updating the CRI status field (e.g. currently, the node server service principal has the full permission to update the entire AzVolumeAttachment object).

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


edreed:
Remove file `charts/latest-v2/charts/azuredisk-csi-driver-v2.0.0-beta.1.tgz`

pkg/provisioner/crdprovisioner.go
```diff
@@ -643,9 +642,35 @@ func (c *CrdProvisioner) demoteVolume(ctx context.Context, azVAClient v1beta1.Az
...
+		updateInstance.Labels[consts.RoleLabel] = string(diskv1beta1.ReplicaRole)
+		updateInstance.Labels[consts.RoleChangeLabel] = consts.Demoted
 		return nil
 	}
-	return azureutils.UpdateCRIWithRetry(ctx, c.conditionWatcher.informerFactory, nil, c.azDiskClient, azVolumeAttachment, updateFunc, consts.NormalUpdateMaxNetRetry)
+	return azureutils.UpdateCRIWithRetry(ctx, c.conditionWatcher.informerFactory, nil, c.azDiskClient, azVolumeAttachment, updateFunc, consts.NormalUpdateMaxNetRetry, azureutils.UpdateAll)
+}
+
+func (c *CrdProvisioner) DetachVolume(ctx context.Context, volumeID string,
+	nodeID string) error {
```
edreed, line 658: Keep parameters on one line.

pkg/provisioner/crdprovisioner.go
```diff
@@ -643,9 +642,35 @@ func (c *CrdProvisioner) demoteVolume(ctx context.Context, azVAClient v1beta1.Az
...
+
+		updateInstance.Labels[consts.RoleLabel] = string(diskv1beta1.ReplicaRole)
+		updateInstance.Labels[consts.RoleChangeLabel] = consts.Demoted
 		return nil
 	}
-	return azureutils.UpdateCRIWithRetry(ctx, c.conditionWatcher.informerFactory, nil, c.azDiskClient, azVolumeAttachment, updateFunc, consts.NormalUpdateMaxNetRetry)
+	return azureutils.UpdateCRIWithRetry(ctx, c.conditionWatcher.informerFactory, nil, c.azDiskClient, azVolumeAttachment, updateFunc, consts.NormalUpdateMaxNetRetry, azureutils.UpdateAll)
+}
+
+func (c *CrdProvisioner) DetachVolume(ctx context.Context, volumeID string,
```
edreed, line 657: I find it very confusing to have both an `UnpublishVolume` and `DetachVolume` method. Which do I use when?

I recommend adding an `UnpublishMode` enum parameter to `UnpublishVolume` to tailor its behavior to the scenario. We can start with values `DemoteOrDetach` for the normal case and `Detach` for the new case. Later we can add `ForceDetach` when we understand the scenario in which that could be useful.

I also recommend refactoring the `maxShares` check into a separate internal method `func (c *CrdProvisioner) shouldDemote(volumeName string, mode UnpublishMode) bool` that only gets the `AzVolume` if `mode == DemoteOrDetach`.

pkg/azureutils/azure_disk_utils.go
```diff
@@ -117,6 +117,14 @@ const (
 	Uncached
 )
 
+type CRIUpdateMode uint
+
+const (
+	UpdateCRI       CRIUpdateMode = 0001
```
edreed, line 123: In Golang, an [integer literal](https://go.dev/ref/spec#Integer_literals) with a leading `0` indicates an octal number. Given you're use of '0' & '1' here, I suspect you wanted binary? If so, you need to prefix with `0b`.

pkg/azureutils/azure_disk_utils.go
```diff
@@ -927,16 +935,22 @@ func UpdateCRIWithRetry(ctx context.Context, informerFactory azurediskInformers.
 
 		switch target := copyForUpdate.(type) {
 		case *diskv1beta1.AzVolume:
-			if cachedClient == nil {
+			if updateMode&UpdateCRIStatus > 0 && !reflect.DeepEqual(objForUpdate.(*diskv1beta1.AzVolume).Status, target.Status) {
```
edreed, line 938: It's conventional to wrap the `&` expression in parentheses and use `!= 0` for flags checks, e.g. `(updateMode&UpdateCRIStatus) != 0`.

pkg/controller/attach_detach.go
```diff
@@ -496,8 +442,7 @@ func updateState(azVolumeAttachment *diskv1beta1.AzVolumeAttachment, state diskv
 		return nil, status.Errorf(codes.FailedPrecondition, "function `updateState` requires non-nil AzVolumeAttachment object.")
 	}
 	if mode == normalUpdate {
-		expectedStates := allowedTargetAttachmentStates[string(azVolumeAttachment.Status.State)]
-		if !containsString(string(state), expectedStates) {
+		if expectedStates, exists := allowedTargetAttachmentStates[string(azVolumeAttachment.Status.State)]; exists && !containsString(string(state), expectedStates) {
```
edreed, line 445: If the current state is not in the map, we allow the state transition. Shouldn't the logic be `!exists || !containsString(...)`?

sunpa93: That was the intention. Given that we cannot prefill our CRI with `OperationPendingState` on creation, we have to allow transition from an empty string to `Creating` or `Attaching` state. I thought about adding this relationship to the map but wasn't sure.

pkg/controller/attach_detach.go
```diff
@@ -554,42 +499,51 @@ func (r *ReconcileAttachDetach) recreateAzVolumeAttachment(ctx context.Context,
...
+			azVolumeAttachment.Status = diskv1beta1.AzVolumeAttachmentStatus{
+				State: azureutils.GetAzVolumeAttachmentState(volumeAttachment.Status),
+			}
+			if azVolumeAttachment.Status.State == diskv1beta1.Attached {
+				azVolumeAttachment.Status.Detail = &diskv1beta1.AzVolumeAttachmentStatusDetail{
+					Role: diskv1beta1.PrimaryRole,
+				}
+			}
+			// update status
+			_, err = r.controllerSharedState.azClient.DiskV1beta1().AzVolumeAttachments(r.controllerSharedState.objectNamespace).UpdateStatus(ctx, azVolumeAttachment, metav1.UpdateOptions{})
```
edreed, line 544: Now that create and status update are two separate operations, the controller may see the created object *before* the status update occurs. In this case, it is ok because this method is only called during recovery and the reconciler will ignore the creation event. However, are there other cases where this would be problematic?

pkg/controller/azvolume.go
```diff
@@ -155,7 +155,10 @@ func (r *ReconcileAzVolume) triggerCreate(ctx context.Context, azVolume *diskv1b
 		defer cloudCancel()
 
 		var updateFunc func(interface{}) error
+		// UpdateCRIWithRetry should be called on a context w/o timeout when called in a separate goroutine as it is not going to be retriggered and leave the CRI in unrecoverable transient state instead.
+		updateCtx := context.Background()
```
edreed, line 159: Why move this here when it was near where it was being used before? It's not being used elsewhere in the PR.

pkg/controller/azvolume.go
```diff
@@ -476,10 +463,10 @@ func (r *ReconcileAzVolume) recoverAzVolume(ctx context.Context, recoveredAzVolu
 				var err error
 				azv := obj.(*diskv1beta1.AzVolume)
 				// add a recover annotation to the CRI so that reconciliation can be triggered for the CRI even if CRI's current state == target state
-				if azv.ObjectMeta.Annotations == nil {
-					azv.ObjectMeta.Annotations = map[string]string{}
+				if azv.Status.Annotations == nil {
+					azv.Status.Annotations = map[string]string{}
 				}
-				azv.ObjectMeta.Annotations[consts.RecoverAnnotation] = "azVolume"
+				azv.Status.Annotations[consts.RecoverAnnotation] = "azVolume"
```
edreed, line 469: It would be worth adding `getStatusAnnotations` and `setStatusAnnotations` methods the wrap common handling of a `nil` map.

pkg/controller/common.go
```diff
@@ -1744,47 +1742,29 @@ func objectDeletionRequested(obj runtime.Object) bool {
...
-		}
-	}
-	return false
+	return volume != nil && volume.Status.Annotations != nil && contains(volume.Status.Annotations, consts.PreProvisionedVolumeCleanupAnnotation)
 }
 
-func labelExists(labels map[string]string, label string) bool {
-	if labels != nil {
-		_, ok := labels[label]
+func contains(mmap map[string]string, key string) bool {
```
edreed, line 1765: Consider `statusAnnotationsContains`

pkg/provisioner/crdprovisioner.go
```diff
@@ -229,10 +230,7 @@ func (c *CrdProvisioner) CreateVolume(
 				Secrets:                   secrets,
 				ContentVolumeSource:       volumeContentSource,
 				AccessibilityRequirements: accessibilityReq,
-			},
-			Status: diskv1beta1.AzVolumeStatus{
-				PersistentVolume: parameters[consts.PvNameKey],
-				State:            diskv1beta1.VolumeOperationPending,
+				PersistentVolume:          parameters[consts.PvNameKey],
```
edreed, line 233: What guarantees do we have that the `consts.PvNameKey` is set in this map? If it's not there, we'll panic. We should fail gracefully if it is missing.

pkg/controller/attach_detach.go
edreed: Let's be explicit in the map. I worry that if we add a new state but forget to update the map, we may allow an arbitrary and invalid state transition.

pkg/provisioner/crdprovisioner.go
sunpa93: https://kubernetes-csi.github.io/docs/external-provisioner.html#persistentvolumeclaim-and-persistentvolume-parameters

This gets added by the external-provisioner when `--extra-create-metadata` is set when installing the provisioner.

pkg/controller/attach_detach.go
sunpa93: Will do 👍

pkg/provisioner/crdprovisioner.go
edreed: Right. However, it would result in a panic if this were omitted in the `csi-provisioner`. Let's be defensive against this case and log an appropriate error.

pkg/controller/common.go
sunpa93: Turned it into a generic function called `mapContains` for `map[string]string`

PR #1263, opened 2022-04-01
doc: refine static PV example

andyzhangx:
**What type of PR is this?**
/kind documentation

doc: refine static PV example

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,66 @@
+# For most projects, this workflow file will not need changing; you simply need
```
edreed, line 1: This addition seems unrelated to the PV example update. Shouldn't this be added in a separate PR?

PR #1273, opened 2022-04-08
[V2] feat: add end to end workflow logging

sunpa93:
**What type of PR is this?**
/kind feature

* Given the v2 driver’s current logging pattern, it is difficult to track the entire end to end workflow, adversely affecting the logs’ readability and the overall debugging experience.
* This PR leverages CRI annotations and golang context to enhance the logging pattern so that each workflow can be tracked end to end more easily.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/Azure/go-autorest/autorest/date v0.3.0
 	github.com/Azure/go-autorest/autorest/to v0.4.0
 	github.com/container-storage-interface/spec v1.5.0
+	github.com/go-logr/logr v1.2.2
 	github.com/golang/mock v1.6.0
 	github.com/golang/protobuf v1.5.2
 	github.com/google/gofuzz v1.2.0 // indirect
@@ -67,7 +68,6 @@ require (
 	github.com/fatih/color v1.13.0 // indirect
 	github.com/felixge/httpsnoop v1.0.1 // indirect
 	github.com/fsnotify/fsnotify v1.5.1 // indirect
-	github.com/go-logr/logr v1.2.2 // indirect
 	github.com/go-openapi/jsonpointer v0.19.5 // indirect
 	github.com/go-openapi/jsonreference v0.19.5 // indirect
 	github.com/go-openapi/spec v0.19.3 // indirect
--- a/pkg/azureconstants/azure_constants.go
+++ b/pkg/azureconstants/azure_constants.go
@@ -110,9 +110,16 @@ const (
 	VolumeDetachRequestAnnotation         = "disk.csi.azure.com/volume-detach-request"
 	RecoverAnnotation                     = "disk.csi.azure.com/recovery" // used to ensure reconciliation is triggered for recovering CRIs
 	VolumeNameLabel                       = "disk.csi.azure.com/volume-name"
+	VolumeIDLabel                         = "disk.csi.azure.com/volume-id"
 	InlineVolumeAnnotation                = "disk.csi.azure.com/inline-volume"
+	PodNameKey                            = "disk.csi/azure.com/pod-name"
 	PreProvisionedVolumeAnnotation        = "disk.csi.azure.com/pre-provisioned"
 	PreProvisionedVolumeCleanupAnnotation = "disk.csi.azure.com/pre-provisioned-clean-up"
+	RequestIDKey                          = "disk.csi.azure.com/request-id"
+	RequestStartimeKey                    = "disk.csi.azure.com/request-starttime"
+	RequestTimeFormat                     = time.RFC3339Nano
+	RequesterKey                          = "disk.csi.azure.com/requester-name"
+	WorkflowKey                           = "disk.csi.azure.com/requester-name"
 
 	ControllerClusterRoleName         = "azuredisk-external-provisioner-role"
 	ControllerClusterRoleBindingName  = "azuredisk-csi-provisioner-binding"
@@ -130,6 +137,7 @@ const (
 	EnableAsyncAttachField  = "enableasyncattach"
 	TooManyRequests         = "TooManyRequests"
 	ClientThrottled         = "client throttled"
+	Latency                 = "latency"
 	// define different sleep time when hit throttling
 	SnapshotOpThrottlingSleepSec = 50
 
@@ -138,6 +146,7 @@ const (
 
 	ReplicaAttachmentFailedEvent  = "ReplicaAttachmentFailed"
 	ReplicaAttachmentSuccessEvent = "ReplicaAttachmentSucceeded"
+	ClientFailedGetEvent          = "ClientFailedToGetObject"
 )
 
 type UnpublishMode int
--- a/pkg/azuredisk/azuredisk_v2.go
+++ b/pkg/azuredisk/azuredisk_v2.go
@@ -52,6 +52,7 @@ import (
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/optimization"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/provisioner"
 	volumehelper "sigs.k8s.io/azuredisk-csi-driver/pkg/util"
+	"sigs.k8s.io/azuredisk-csi-driver/pkg/workflow"
 	azurecloudconsts "sigs.k8s.io/cloud-provider-azure/pkg/consts"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
@@ -368,19 +369,22 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
 	// Leader controller manager should recover CRI if possible and clean them up before exiting.
 	go func() {
 		<-mgr.Elected()
+		var errors []error
+		ctx, w := workflow.New(ctx)
+		defer func() { w.Finish(err) }()
 		// recover lost states if necessary
-		klog.Infof("Elected as leader; initiating CRI recovery...")
+		w.Logger().Infof("Elected as leader; initiating CRI recovery...")
 		if err := azvReconciler.Recover(ctx); err != nil {
-			klog.Warningf("Failed to recover AzVolume: %v", err)
+			errors = append(errors, err)
 		}
 		if err := attachReconciler.Recover(ctx); err != nil {
-			klog.Warningf("Failed to recover AzVolumeAttachments: %v.", err)
+			errors = append(errors, err)
 		}
 		if err := pvReconciler.Recover(ctx); err != nil {
-			klog.Warningf("Failed to restore shared claimToVolumeMap and volumeToClaimMap: %v.", err)
+			errors = append(errors, err)
 		}
 		if err := podReconciler.Recover(ctx); err != nil {
-			klog.Warningf("Failed to recover replica AzVolumeAttachments: %v.", err)
+			errors = append(errors, err)
 		}
 		sharedState.MarkRecoveryComplete()
 	}()
--- a/pkg/azureutils/azure_disk_utils.go
+++ b/pkg/azureutils/azure_disk_utils.go
@@ -57,6 +57,7 @@ import (
 	consts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/optimization"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/util"
+	"sigs.k8s.io/azuredisk-csi-driver/pkg/workflow"
 	"sigs.k8s.io/cloud-provider-azure/pkg/azureclients"
 	azure "sigs.k8s.io/cloud-provider-azure/pkg/provider"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -883,8 +884,10 @@ func GetAzVolumeAttachmentState(volumeAttachmentStatus storagev1.VolumeAttachmen
 }
 
 func UpdateCRIWithRetry(ctx context.Context, informerFactory azurediskInformers.SharedInformerFactory, cachedClient client.Client, azDiskClient azDiskClientSet.Interface, obj client.Object, updateFunc func(interface{}) error, maxNetRetry int, updateMode CRIUpdateMode) error {
+	var err error
 	objName := obj.GetName()
-	klog.Infof("Initiating update with retry for %v (%s)", reflect.TypeOf(obj), objName)
+	ctx, w := workflow.New(ctx)
+	defer func() { w.Finish(err) }()
 
 	conditionFunc := func() error {
 		var err error
@@ -920,7 +923,7 @@ func UpdateCRIWithRetry(ctx context.Context, informerFactory azurediskInformers.
 		}
 
 		if err != nil {
-			klog.Errorf("failed to get %v (%s): %v", reflect.TypeOf(obj), objName, err)
+			err = status.Errorf(codes.Internal, "failed to get object: %v", err)
 			return err
 		}
 
@@ -970,7 +973,7 @@ func UpdateCRIWithRetry(ctx context.Context, informerFactory azurediskInformers.
 		return false
 	}
 
-	err := retry.OnError(
+	err = retry.OnError(
 		wait.Backoff{
 			Duration: consts.CRIUpdateRetryDuration,
 			Factor:   consts.CRIUpdateRetryFactor,
@@ -979,9 +982,6 @@ func UpdateCRIWithRetry(ctx context.Context, informerFactory azurediskInformers.
 		isRetriable,
 		conditionFunc,
 	)
-	if err != nil {
-		klog.Errorf("failed to update %v (%s): %v", reflect.TypeOf(obj), objName, err)
-	}
 
 	// if encountered net error from api server unavailability, exit process
 	ExitOnNetError(err)
--- a/pkg/controller/attach_detach.go
+++ b/pkg/controller/attach_detach.go
@@ -22,16 +22,17 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/go-logr/logr"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/klog/v2"
 	diskv1beta1 "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/azuredisk/v1beta1"
 	consts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/util"
+	"sigs.k8s.io/azuredisk-csi-driver/pkg/workflow"
 
 	volerr "k8s.io/cloud-provider/volume/errors"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -58,6 +59,7 @@ Attach Detach controller is responsible for
 	3. detaching volume upon deletions marked with certain annotations
 */
 type ReconcileAttachDetach struct {
+	logger                logr.Logger
 	crdDetacher           CrdDetacher
 	cloudDiskAttacher     CloudDiskAttachDetacher
 	controllerSharedState *SharedState
@@ -90,35 +92,36 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
 		return reconcileReturnOnSuccess(request.Name, r.retryInfo)
 	} else if err != nil {
 		azVolumeAttachment.Name = request.Name
-		return reconcileReturnOnError(azVolumeAttachment, "get", err, r.retryInfo)
+		return reconcileReturnOnError(ctx, azVolumeAttachment, "get", err, r.retryInfo)
 	}
+
+	ctx, _ = workflow.GetWorkflowFromObj(ctx, azVolumeAttachment)
 
 	// if underlying cloud operation already in process, skip until operation is completed
 	if isOperationInProcess(azVolumeAttachment) {
-		klog.V(5).Infof("Another operation (%s) is already in process for the AzVolumeAttachment (%s). Will be requeued once complete.", azVolumeAttachment.Status.State, azVolumeAttachment.Name)
 		return reconcileReturnOnSuccess(azVolumeAttachment.Name, r.retryInfo)
 	}
 
 	// detachment request
 	if objectDeletionRequested(azVolumeAttachment) {
 		if err := r.triggerDetach(ctx, azVolumeAttachment); err != nil {
-			return reconcileReturnOnError(azVolumeAttachment, "detach", err, r.retryInfo)
+			return reconcileReturnOnError(ctx, azVolumeAttachment, "detach", err, r.retryInfo)
 		}
 		// attachment request
 	} else if azVolumeAttachment.Status.Detail == nil {
 		if err := r.triggerAttach(ctx, azVolumeAttachment); err != nil {
-			return reconcileReturnOnError(azVolumeAttachment, "attach", err, r.retryInfo)
+			return reconcileReturnOnError(ctx, azVolumeAttachment, "attach", err, r.retryInfo)
 		}
 		// promotion request
 	} else if azVolumeAttachment.Spec.RequestedRole != azVolumeAttachment.Status.Detail.Role {
 		switch azVolumeAttachment.Spec.RequestedRole {
 		case diskv1beta1.PrimaryRole:
 			if err := r.promote(ctx, azVolumeAttachment); err != nil {
-				return reconcileReturnOnError(azVolumeAttachment, "promote", err, r.retryInfo)
+				return reconcileReturnOnError(ctx, azVolumeAttachment, "promote", err, r.retryInfo)
 			}
 		case diskv1beta1.ReplicaRole:
 			if err := r.demote(ctx, azVolumeAttachment); err != nil {
-				return reconcileReturnOnError(azVolumeAttachment, "demote", err, r.retryInfo)
+				return reconcileReturnOnError(ctx, azVolumeAttachment, "demote", err, r.retryInfo)
 			}
 		}
 	}
@@ -127,20 +130,27 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
 }
 
 func (r *ReconcileAttachDetach) triggerAttach(ctx context.Context, azVolumeAttachment *diskv1beta1.AzVolumeAttachment) error {
+	var err error
+	ctx, w := workflow.New(ctx)
+	defer func() { w.Finish(err) }()
+
 	// requeue if AzVolumeAttachment's state is being updated by a different worker
 	defer r.stateLock.Delete(azVolumeAttachment.Name)
 	if _, ok := r.stateLock.LoadOrStore(azVolumeAttachment.Name, nil); ok {
-		return getOperationRequeueError("attach", azVolumeAttachment)
+		err = getOperationRequeueError("attach", azVolumeAttachment)
+		return err
 	}
 
-	if azVolume, err := azureutils.GetAzVolume(ctx, r.controllerSharedState.cachedClient, r.controllerSharedState.azClient, strings.ToLower(azVolumeAttachment.Spec.VolumeName), r.controllerSharedState.objectNamespace, true); err != nil {
+	var azVolume *diskv1beta1.AzVolume
+	if azVolume, err = azureutils.GetAzVolume(ctx, r.controllerSharedState.cachedClient, r.controllerSharedState.azClient, strings.ToLower(azVolumeAttachment.Spec.VolumeName), r.controllerSharedState.objectNamespace, true); err != nil {
 		if errors.IsNotFound(err) {
-			klog.Infof("Aborting attach operation for AzVolumeAttachment (%s): AzVolume (%s) not found", azVolumeAttachment.Name, azVolumeAttachment.Spec.VolumeName)
+			w.Logger().V(5).Infof("Aborting attach operation for AzVolumeAttachment (%s): AzVolume (%s) not found", azVolumeAttachment.Name, azVolumeAttachment.Spec.VolumeName)
+			err = nil
 			return nil
 		}
 		return err
 	} else if objectDeletionRequested(azVolume) {
-		klog.Infof("Aborting attach operation for AzVolumeAttachment (%s): AzVolume (%s) scheduled for deletion", azVolumeAttachment.Name, azVolumeAttachment.Spec.VolumeName)
+		w.Logger().V(5).Infof("Aborting attach operation for AzVolumeAttachment (%s): AzVolume (%s) scheduled for deletion", azVolumeAttachment.Name, azVolumeAttachment.Spec.VolumeName)
 		return nil
 	}
 
@@ -151,48 +161,51 @@ func (r *ReconcileAttachDetach) triggerAttach(ctx context.Context, azVolumeAttac
 		_, derr := updateState(azv, diskv1beta1.Attaching, normalUpdate)
 		return derr
 	}
-	if err := azureutils.UpdateCRIWithRetry(ctx, nil, r.controllerSharedState.cachedClient, r.controllerSharedState.azClient, azVolumeAttachment, updateFunc, consts.NormalUpdateMaxNetRetry, azureutils.UpdateCRIStatus); err != nil {
+	if err = azureutils.UpdateCRIWithRetry(ctx, nil, r.controllerSharedState.cachedClient, r.controllerSharedState.azClient, azVolumeAttachment, updateFunc, consts.NormalUpdateMaxNetRetry, azureutils.UpdateCRIStatus); err != nil {
 		return err
 	}
 
-	klog.Infof("Attaching volume (%s) to node (%s)", azVolumeAttachment.Spec.VolumeID, azVolumeAttachment.Spec.NodeName)
-
-	// initiate goroutine to attach volume
-	//nolint:contextcheck // Attach is performed asynchronously by the reconciler; context is not inherited by design
+	w.Logger().Info("Attaching volume")
+	waitCh := make(chan goSignal)
+	//nolint:contextcheck // call is asynchronous; context is not inherited by design
 	go func() {
-		cloudCtx, cloudCancel := context.WithTimeout(context.Background(), cloudTimeout)
+		var attachErr error
+		_, goWorkflow := workflow.New(ctx)
+		defer func() { goWorkflow.Finish(attachErr) }()
+		waitCh <- goSignal{}
+
+		goCtx := goWorkflow.SaveToContext(context.Background())
+		cloudCtx, cloudCancel := context.WithTimeout(goCtx, cloudTimeout)
 		defer cloudCancel()
 
 		// attempt to attach the disk to a node
-		response, attachErr := r.attachVolume(cloudCtx, azVolumeAttachment.Spec.VolumeID, azVolumeAttachment.Spec.NodeName, azVolumeAttachment.Spec.VolumeContext)
+		var response map[string]string
+		response, attachErr = r.attachVolume(cloudCtx, azVolumeAttachment.Spec.VolumeID, azVolumeAttachment.Spec.NodeName, azVolumeAttachment.Spec.VolumeContext)
 		// if the disk is attached to a different node
 		if danglingAttachErr, ok := attachErr.(*volerr.DanglingAttachError); ok {
 			// get disk, current node and attachment name
-			diskName, err := azureutils.GetDiskName(azVolumeAttachment.Spec.VolumeID)
-			if err != nil {
-				klog.Warningf("failed to extract disk name from volumeID (%s): %v", azVolumeAttachment.Spec.VolumeID, err)
-			}
 			currentNodeName := string(danglingAttachErr.CurrentNode)
-			currentAttachmentName := azureutils.GetAzVolumeAttachmentName(diskName, currentNodeName)
+			currentAttachmentName := azureutils.GetAzVolumeAttachmentName(azVolumeAttachment.Spec.VolumeName, currentNodeName)
+			goWorkflow.Logger().Infof("Dangling attach detected for %s", currentNodeName)
 
 			// check if AzVolumeAttachment exists for the existing attachment
-			_, err = r.controllerSharedState.azClient.DiskV1beta1().AzVolumeAttachments(r.controllerSharedState.objectNamespace).Get(cloudCtx, currentAttachmentName, metav1.GetOptions{})
+			_, err := r.controllerSharedState.azClient.DiskV1beta1().AzVolumeAttachments(r.controllerSharedState.objectNamespace).Get(cloudCtx, currentAttachmentName, metav1.GetOptions{})
 			var detachErr error
 			if errors.IsNotFound(err) {
 				// AzVolumeAttachment doesn't exist so we only need to detach disk from cloud
-				detachErr = r.cloudDiskAttacher.UnpublishVolume(ctx, azVolumeAttachment.Spec.VolumeID, currentNodeName)
+				detachErr = r.cloudDiskAttacher.UnpublishVolume(goCtx, azVolumeAttachment.Spec.VolumeID, currentNodeName)
 				if detachErr != nil {
-					klog.Errorf("failed to detach volume (%s) from node (%s). Error: %v", azVolumeAttachment.Spec.VolumeID, currentNodeName, err)
+					goWorkflow.Logger().Errorf(detachErr, "failed to detach dangling volume (%s) from node (%s). Error: %v", azVolumeAttachment.Spec.VolumeID, currentNodeName, err)
 				}
 			} else {
 				// AzVolumeAttachment exist so we need to detach disk through crdProvisioner
-				detachErr = r.crdDetacher.UnpublishVolume(ctx, azVolumeAttachment.Spec.VolumeID, currentNodeName, make(map[string]string), consts.Detach)
+				detachErr = r.crdDetacher.UnpublishVolume(goCtx, azVolumeAttachment.Spec.VolumeID, currentNodeName, make(map[string]string), consts.Detach)
 				if detachErr != nil {
-					klog.Errorf("failed to make a unpublish request AzVolumeAttachment for volume (%s) and node (%s). Error: %v", azVolumeAttachment.Spec.VolumeID, currentNodeName, err)
+					goWorkflow.Logger().Errorf(detachErr, "failed to make a unpublish request dangling AzVolumeAttachment for volume (%s) and node (%s)", azVolumeAttachment.Spec.VolumeID, currentNodeName)
 				}
-				detachErr = r.crdDetacher.WaitForDetach(ctx, azVolumeAttachment.Spec.VolumeID, currentNodeName)
+				detachErr = r.crdDetacher.WaitForDetach(goCtx, azVolumeAttachment.Spec.VolumeID, currentNodeName)
 				if detachErr != nil {
-					klog.Errorf("failed to unpublish AzVolumeAttachment for volume (%s) and node (%s). Error: %v", azVolumeAttachment.Spec.VolumeID, currentNodeName, err)
+					goWorkflow.Logger().Errorf(detachErr, "failed to unpublish dangling AzVolumeAttachment for volume (%s) and node (%s)", azVolumeAttachment.Spec.VolumeID, currentNodeName)
 				}
 			}
 			// attempt to attach the disk to a node after detach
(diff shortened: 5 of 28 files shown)


sunpa93:
WIP. Currently running more tests to verify logs consistency.

pkg/azureutils/logger.go
```diff
@@ -0,0 +1,189 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2/klogr"
+	diskv1beta1 "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/azuredisk/v1beta1"
+)
+
+type contextKey int
+
+const workflowKey contextKey = 1
+
+type Workflow struct {
```
edreed, line 21: I recommend placing `Workflow` and related functions in a separate `workflow.go` file.

pkg/azureutils/logger.go
```diff
@@ -0,0 +1,189 @@
...
+		startTime: time.Now(),
+		logger:    Logger{klogr.New()},
+	}
+	w.rootWorkflow = &w
+	return w
+}
+
+func (w Workflow) Finish(err error) {
+	latency := time.Since(w.startTime)
+	w.logger.WithValues(consts.Latency, latency, "error", err, "workflow", w.name).WithCallDepth(1).Info("Workflow completed.")
```
edreed, line 91: We should log error status through the `Error` function and omit a nil error in the success case.

pkg/provisioner/crdprovisioner.go
```diff
@@ -154,40 +155,29 @@ func (c *CrdProvisioner) CreateVolume(
...
-		}
-		return false, nil
-	})
-	if err != nil {
-		return nil, err
-	}
-	defer waiter.Close()
-
-	azVolumeInstance, err := lister.Get(azVolumeName)
+	ctx, workflow := azureutils.StartWorkflow(ctx, "CrdProvisioner.CreateVolume", consts.VolumeNameLabel, volumeName, consts.PvNameKey, parameters[consts.PvNameKey])
```
edreed, line 158: Rather than require the caller to pass in the function name, we could use [`runtime.Caller(1)`](https://pkg.go.dev/runtime#Caller) and [`runtime.FuncForPC()`](https://pkg.go.dev/runtime#FuncForPC) to discover the calling function name. We would end up with a name like `provisioner.(*CrdProvisioner).CreateVolume`.

pkg/provisioner/crdprovisioner.go
```diff
@@ -212,19 +202,18 @@ func (c *CrdProvisioner) CreateVolume(
 		}
 		// if the error was caused by errors other than IsNotFound, return failure
 	} else if !apiErrors.IsNotFound(err) {
-		klog.Error("failed to get AzVolume (%s): %v", azVolumeName, err)
+		workflow.Logger().Error(err, "failed to get AzVolume CRI")
```
edreed, line 205: We should consider converting this to an internal error and allow `Finish` to log it, e.g.:

```go
err = status.Errorf(codes.Internal, "failed to get AzVolume CRI: %v", err)
```

pkg/provisioner/crdprovisioner.go
```diff
@@ -239,36 +228,55 @@ func (c *CrdProvisioner) CreateVolume(
...
+	if err != nil {
+		return nil, err
 	}
+	defer waiter.Close()
 
-	obj, err := waiter.Wait(ctx)
+	var obj runtime.Object
+	obj, err = waiter.Wait(ctx)
 	if obj == nil || err != nil {
+		workflow.Logger().Error(err, "failed to create volume")
```
edreed, line 263: This seems unnecessary. We should be able to infer the volume creation failure from the `Finish` log.

pkg/azureutils/logger.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func (w Workflow) AddRequestToObject(obj runtime.Object) {
```
edreed, line 94: Consider `AnnotateObject`, e.g.

```go
workflow.AnnotateObject(avVolume)
```
```

File: pkg/azureutils/logger.go [hunk 7]
```diff
@@ -0,0 +1,189 @@
...
+
+func (w Workflow) AddDetailToLogger(details ...interface{}) {
+	w.logger = Logger{w.logger.WithValues(details...)}
+}
+
+func (w Workflow) SaveToContext(ctx context.Context) context.Context {
+	return context.WithValue(ctx, workflowKey, &w)
+}
+
+func WorkflowFromContext(ctx context.Context) (Workflow, bool) {
```
edreed, line 132: Consider `GetWorkflowFromContext` for consistency with the other `GetWorkflow[FromObj]` methods.

pkg/provisioner/crdprovisioner.go
```diff
@@ -154,40 +155,29 @@ func (c *CrdProvisioner) CreateVolume(
...
+	defer func() { workflow.Finish(err) }()
 
+	var azVolumeInstance *diskv1beta1.AzVolume
+	azVolumeInstance, err = azVLister.Get(azVolumeName)
 	if err == nil {
 		if azVolumeInstance.Status.Detail != nil {
 			// If current request has different specifications than the existing volume, return error.
 			if !isAzVolumeSpecSameAsRequestParams(azVolumeInstance, maxMountReplicaCount, capacityRange, parameters, secrets, volumeContentSource, accessibilityReq) {
-				return nil, status.Errorf(codes.AlreadyExists, "Volume with name (%s) already exists with different specifications", volumeName)
+				err = status.Errorf(codes.AlreadyExists, "Volume with name (%s) already exists with different specifications", volumeName)
```
edreed, line 167: Do not include information in the error message that is already available in the workflow details.

pkg/provisioner/crdprovisioner.go
```diff
@@ -288,17 +297,34 @@ func (c *CrdProvisioner) DeleteVolume(ctx context.Context, volumeID string, secr
...
 
-	waiter, err := c.conditionWatcher.newConditionWaiter(ctx, azVolumeType, azVolumeName, func(obj interface{}, objectDeleted bool) (bool, error) {
+	var azVolumeInstance *diskv1beta1.AzVolume
+	azVolumeInstance, err = lister.Get(azVolumeName)
+	ctx, workflow := azureutils.StartWorkflow(ctx, "CrdProvisioner.DeleteVolume", azureutils.GetObjectDetails(azVolumeInstance)...)
+	defer func() { workflow.Finish(err) }()
+
+	if err != nil {
+		if apiErrors.IsNotFound(err) {
+			workflow.Logger().Logger.WithValues(consts.VolumeNameLabel, volumeName).Info("Deletion successful: could not find the AzVolume CRI")
```
edreed, line 319: Hasn't the volume name already been added to the details by `azureutils.GetObjectDetails()`?

pkg/provisioner/crdprovisioner.go
```diff
@@ -600,29 +647,36 @@ func (c *CrdProvisioner) UnpublishVolume(
...
-	azVolumeAttachmentInstance, err := azVALister.Get(attachmentName)
+	var azVolumeAttachmentInstance *diskv1beta1.AzVolumeAttachment
+	azVolumeAttachmentInstance, err = azVALister.Get(attachmentName)
+	ctx, workflow := azureutils.StartWorkflow(ctx, "CrdProvisioner.UnpublishVolume", azureutils.GetObjectDetails(azVolumeAttachmentInstance)...)
+	defer func() { workflow.Finish(err) }()
+
 	if err != nil {
 		if apiErrors.IsNotFound(err) {
-			klog.Infof("AzVolumeAttachment(%s) deleted: volume (%s) has been successfully detached from node (%s)", attachmentName, volumeName, nodeID)
+			workflow.Logger().WithValues(consts.VolumeNameLabel, volumeName, consts.NodeNameLabel, nodeID).Info("Volume detach successful")
```
edreed, line 670: Weren't volume and node name already added by `azureutils.GetObjectDetails()`?

pkg/provisioner/crdprovisioner.go
```diff
@@ -644,10 +698,12 @@ func (c *CrdProvisioner) shouldDemote(volumeName string, mode consts.UnpublishMo
 }
 
 func (c *CrdProvisioner) demoteVolume(ctx context.Context, azVAClient v1beta1.AzVolumeAttachmentInterface, azVolumeAttachment *diskv1beta1.AzVolumeAttachment) error {
-	klog.Infof("Requesting AzVolumeAttachment (%s) demotion", azVolumeAttachment.Name)
+	ctx, workflow := azureutils.GetWorkflowFromObj(ctx, azVolumeAttachment)
+	workflow.Logger().Infof("Requesting AzVolumeAttachment (%s) demotion", azVolumeAttachment.Name)
```
edreed, line 702: Consider making this a child workflow and removing the intermediate logging.

pkg/azureutils/logger.go
```diff
@@ -0,0 +1,189 @@
...
+	return []interface{}{}
+}
+
+func StartWorkflow(ctx context.Context, name string, details ...interface{}) (context.Context, Workflow) {
+	w := Workflow{
+		requestID: GenerateRequestID(),
+	}
+	w.logger = Logger{klogr.New().WithValues(consts.RequestIDKey, w.requestID).WithValues(details...)}
+	w.rootWorkflow = &w
+	return w.StartChildWorkflow(ctx, name)
```
edreed, line 66: Shouldn't this be `w.SaveToContext(cts), w`?

pkg/azureutils/logger.go
```diff
@@ -0,0 +1,189 @@
...
+			return []interface{}{consts.VolumeNameLabel, target.Spec.VolumeName, consts.NodeNameLabel, target.Spec.NodeName, consts.RoleLabel, target.Spec.RequestedRole}
+		}
+	}
+	return []interface{}{}
+}
+
+func StartWorkflow(ctx context.Context, name string, details ...interface{}) (context.Context, Workflow) {
+	w := Workflow{
+		requestID: GenerateRequestID(),
+	}
```
edreed, line 63: Shouldn't we set `name` here?

pkg/azureutils/logger.go
```diff
@@ -0,0 +1,189 @@
...
+	meta, _ := meta.Accessor(obj)
+	if meta == nil {
+		return
+	}
+	annotations := meta.GetAnnotations()
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	annotations[consts.RequestIDKey] = w.requestID
+	annotations[consts.RequestStartimeKey] = w.startTime.Format(consts.RequestTimeFormat)
```
edreed, line 104: Should we add the workflow name as well so we can recover it in `GetWorkflowFromObj`?

pkg/azureutils/logger.go
```diff
@@ -0,0 +1,189 @@
...
+func (w Workflow) Name() string {
+	return w.name
+}
+
+func (w Workflow) AddDetailToLogger(details ...interface{}) {
+	w.logger = Logger{w.logger.WithValues(details...)}
+}
+
+func (w Workflow) SaveToContext(ctx context.Context) context.Context {
+	return context.WithValue(ctx, workflowKey, &w)
```
edreed, line 129: Consider saving the logger to the `Context` as well so that logr/klogr clients will can log through our logger if they're context-aware, e.g.

```go
return logr.NewContext(context.WithValue(ctx, workflowKey, &w), w.logger)
```

pkg/azureutils/logger.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2/klogr"
+	diskv1beta1 "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/azuredisk/v1beta1"
+)
+
+type contextKey int
+
+const workflowKey contextKey = 1
```
edreed, line 19: I believe you could use an empty struct of unique type for the key. This is what [`logr.NewContext()`](https://github.com/go-logr/logr/blob/47e013cee9b1f91c987cc70a218639655431b607/logr.go#L395) does.

pkg/controller/attach_detach.go
```diff
@@ -90,35 +90,37 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
...
-		return reconcileReturnOnError(azVolumeAttachment, "get", err, r.retryInfo)
+		return reconcileReturnOnError(ctx, azVolumeAttachment, "get", err, r.retryInfo)
 	}
 
+	ctx, workflow := azureutils.GetWorkflowFromObj(ctx, azVolumeAttachment)
+
 	// if underlying cloud operation already in process, skip until operation is completed
 	if isOperationInProcess(azVolumeAttachment) {
-		klog.V(5).Infof("Another operation (%s) is already in process for the AzVolumeAttachment (%s). Will be requeued once complete.", azVolumeAttachment.Status.State, azVolumeAttachment.Name)
+		workflow.Logger().V(5).Infof("Another operation (%s) is already in process for the AzVolumeAttachment (%s). Will be requeued once complete.", azVolumeAttachment.Status.State, azVolumeAttachment.Name)
```
edreed, line 100: If not for this verbose log, we wouldn't need to object the workflow in the `Reconcile` function. Since we expect the reconciler to be invoked for intermediate object updates, is this log ever useful? Let's remove it if not.

pkg/azureutils/logger.go
```diff
@@ -0,0 +1,189 @@
...
+func StartWorkflow(ctx context.Context, name string, details ...interface{}) (context.Context, Workflow) {
+	w := Workflow{
+		requestID: GenerateRequestID(),
+	}
+	w.logger = Logger{klogr.New().WithValues(consts.RequestIDKey, w.requestID).WithValues(details...)}
+	w.rootWorkflow = &w
+	return w.StartChildWorkflow(ctx, name)
+}
+
+func (w Workflow) StartChildWorkflow(ctx context.Context, name string) (context.Context, Workflow) {
```
edreed, line 69: Consider having a single `NewWorkflow(ctx context.Context, name string)` method that implicitly creates a child workflow if a workflow is already associated with the context and a root workflow if none. Calls into the `CrdProvisioner` and `CloudProvisioner` made from the controllers will use workflow appropriately attached to the controllers' workflow w/o having to make assumptions about whether they're called in a root or child context.

pkg/controller/attach_detach.go
```diff
@@ -151,48 +166,47 @@ func (r *ReconcileAttachDetach) triggerAttach(ctx context.Context, azVolumeAttac
...
-			_, err = r.controllerSharedState.azClient.DiskV1beta1().AzVolumeAttachments(r.controllerSharedState.objectNamespace).Get(cloudCtx, currentAttachmentName, metav1.GetOptions{})
+			_, err := r.controllerSharedState.azClient.DiskV1beta1().AzVolumeAttachments(r.controllerSharedState.objectNamespace).Get(cloudCtx, currentAttachmentName, metav1.GetOptions{})
 			var detachErr error
 			if errors.IsNotFound(err) {
 				// AzVolumeAttachment doesn't exist so we only need to detach disk from cloud
-				detachErr = r.cloudDiskAttacher.UnpublishVolume(ctx, azVolumeAttachment.Spec.VolumeID, currentNodeName)
+				detachErr = r.cloudDiskAttacher.UnpublishVolume(goCtx, azVolumeAttachment.Spec.VolumeID, currentNodeName)
 				if detachErr != nil {
-					klog.Errorf("failed to detach volume (%s) from node (%s). Error: %v", azVolumeAttachment.Spec.VolumeID, currentNodeName, err)
+					workflow.Logger().Errorf(err, "failed to detach dangling volume (%s) from node (%s). Error: %v", azVolumeAttachment.Spec.VolumeID, currentNodeName, err)
```
edreed, line 199: We should be logging `detachErr` here since we know `err` is "not found". Same with lines 205 & 209.

pkg/controller/attach_detach.go
```diff
@@ -203,16 +217,16 @@ func (r *ReconcileAttachDetach) triggerAttach(ctx context.Context, azVolumeAttac
...
-				klog.Infof("failed to get pods for volume (%s). Error: %v", azVolumeAttachment.Spec.VolumeName, err)
+				workflow.Logger().Error(err, "failed to list pods for volume")
 			}
 		}
 
 		// update AzVolumeAttachment CRI with the result of the attach operation
 		var updateFunc func(interface{}) error
 		if attachErr != nil {
-			klog.Errorf("failed to attach volume %s to node %s: %v", azVolumeAttachment.Spec.VolumeName, azVolumeAttachment.Spec.NodeName, attachErr)
+			workflow.Logger().Errorf(attachErr, "attach volume failed")
```
edreed, line 229: Error will be logged in `workflow.Finish`

pkg/controller/attach_detach.go
```diff
@@ -227,7 +241,7 @@ func (r *ReconcileAttachDetach) triggerAttach(ctx context.Context, azVolumeAttac
 				return uerr
 			}
 		} else {
-			klog.Infof("successfully attached volume (%s) to node (%s) and update status of AzVolumeAttachment (%s)", azVolumeAttachment.Spec.VolumeName, azVolumeAttachment.Spec.NodeName, azVolumeAttachment.Name)
+			workflow.Logger().Infof("attach volume succeeded")
```
edreed, line 244: Success will be logged in `workflow.Finish`

pkg/controller/attach_detach.go
```diff
@@ -245,14 +259,23 @@ func (r *ReconcileAttachDetach) triggerAttach(ctx context.Context, azVolumeAttac
 		}
 
 		// UpdateCRIWithRetry should be called on a context w/o timeout when called in a separate goroutine as it is not going to be retriggered and leave the CRI in unrecoverable transient state instead.
-		updateCtx := context.Background()
+		updateCtx := workflow.SaveToContext(goCtx)
```
edreed, line 262: Does `goCtx` already contain the workflow?

pkg/controller/attach_detach.go
```diff
@@ -272,23 +295,28 @@ func (r *ReconcileAttachDetach) triggerDetach(ctx context.Context, azVolumeAttac
 			return err
 		}
 
-		klog.Infof("Detaching volume (%s) from node (%s)", azVolumeAttachment.Spec.VolumeID, azVolumeAttachment.Spec.NodeName)
-
+		workflow.Logger().Info("Detaching volume")
```
edreed, line 298: Let's rely on workflow logging.

pkg/provisioner/conditionwaiter.go
```diff
@@ -52,13 +55,15 @@ func (w *conditionWaiter) Wait(ctx context.Context) (runtime.Object, error) {
 		}
 	}
 
+	workflow := azureutils.GetWorkflow(ctx, obj)
+
 	// if not wait for the event handler signal
 	select {
 	case <-ctx.Done():
-		klog.Errorf("context timeout for %s (%s)", w.objType, w.objName)
+		workflow.Logger().Error(status.Error(codes.DeadlineExceeded, "context timeout"), "context timeout")
```
edreed, line 63: We should log `ctx.Err()` since it could have canceled.

pkg/provisioner/conditionwaiter.go
```diff
@@ -52,13 +55,15 @@ func (w *conditionWaiter) Wait(ctx context.Context) (runtime.Object, error) {
...
 		return nil, ctx.Err()
 	case waitResult := <-w.entry.waitChan:
-		klog.Infof("received result for %s (%s): error :%v", w.objType, w.objName, waitResult.err)
+		workflow.Logger().Infof("received result for %s (%s): error :%v", w.objType, w.objName, waitResult.err)
```
edreed, line 66: This should be a verbose log, if logged at all.

sunpa93: I am still working on the verbosities. There still seem to remain many log lines to be removed.

pkg/azureutils/azure_disk_utils.go
```diff
@@ -884,7 +884,8 @@ func GetAzVolumeAttachmentState(volumeAttachmentStatus storagev1.VolumeAttachmen
 
 func UpdateCRIWithRetry(ctx context.Context, informerFactory azurediskInformers.SharedInformerFactory, cachedClient client.Client, azDiskClient azDiskClientSet.Interface, obj client.Object, updateFunc func(interface{}) error, maxNetRetry int, updateMode CRIUpdateMode) error {
 	objName := obj.GetName()
-	klog.Infof("Initiating update with retry for %v (%s)", reflect.TypeOf(obj), objName)
+	workflow, _ := WorkflowFromContext(ctx)
+	workflow.Logger().Info("Initiating update with retry")
```
edreed, line 888: This should be a verbose log.

pkg/provisioner/conditionwaiter.go
edreed: Consider making object type and name workflow (logger) details instead of including in the message string. We should also log errors as errors and omit a `nil` error in the success case.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -368,19 +368,20 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
 	// Leader controller manager should recover CRI if possible and clean them up before exiting.
 	go func() {
 		<-mgr.Elected()
+		ctx, workflow := azureutils.StartWorkflow(ctx, "Recover CRIs")
 		// recover lost states if necessary
```
edreed, line 372: defer `Finish`

pkg/provisioner/crdprovisioner.go
sunpa93: The values need to be manually added to the logger when object is not found because object details cannot be extracted from nil objects.

pkg/provisioner/crdprovisioner.go
sunpa93: Same as for when `AzVolume` is not found, when `AzVolumeAttachment` is not found, logger needs to be manually added values.

pkg/azureutils/logger.go
sunpa93: This was to refactor portions `StartWorkflow` and `Workflow.StartChildWorkflow` both share but now that they are unified under a single `NewWorkflow` function, this should be good.

pkg/provisioner/conditionwaiter.go
sunpa93: I removed the log portion and instead started a child workflow in `conditionWaiter` so that we can infer the results from the `workflow.Finish()`

pkg/controller/attach_detach.go
sunpa93: I agree that this log line is better off removed but what would the rationale be for not objecting the workflow in `Reconcile` function?

pkg/controller/attach_detach.go
```diff
@@ -127,20 +127,36 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
 }
 
 func (r *ReconcileAttachDetach) triggerAttach(ctx context.Context, azVolumeAttachment *diskv1beta1.AzVolumeAttachment) error {
+	var err error
+	var waitForGoroutine bool
+	var aborted bool
+	ctx, workflow := azureutils.NewWorkflow(ctx)
```
nearora-msft, line 133: Should we pass the AzVolumeAttachment name/id as details while creating NewWorkflow(), so that we could also search  the logs by that Id?

pkg/controller/azvolume.go
```diff
@@ -98,44 +97,52 @@ func (r *ReconcileAzVolume) Reconcile(ctx context.Context, request reconcile.Req
...
-		// azVolume released, so clean up replica Attachments (primary Attachment should be deleted via UnpublishVolume request to avoid race condition)
 	}
 
 	return reconcileReturnOnSuccess(azVolume.Name, r.retryInfo)
 }
 
 func (r *ReconcileAzVolume) triggerCreate(ctx context.Context, azVolume *diskv1beta1.AzVolume) error {
+	var err error
+	var waitForGoroutine bool
+	ctx, workflow := azureutils.NewWorkflow(ctx)
```
nearora-msft, line 134: Similar comment as before, Should we pass AzVolume name as details param to NewWorkflow so that we can search the logs for that specific volume?

pkg/azureconstants/azure_constants.go
```diff
@@ -63,6 +63,7 @@ const (
 	PvcNameTag                    = "kubernetes.io-created-for-pvc-name"
 	PvNameKey                     = "csi.storage.k8s.io/pv/name"
 	PvNameTag                     = "kubernetes.io-created-for-pv-name"
+	PodKey                        = "disk.csi.azure.com/podName"
```
edreed, line 66: Is the use case for this different than `PodNameKey` below? It should be `disk.csi.azure.com/pod-name` for consistency, but that would duplicate the constant below.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -368,19 +370,29 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
...
 	go func() {
 		<-mgr.Elected()
+		var errors []error
+		ctx, workflow := azureutils.NewWorkflow(ctx)
+		defer func() {
+			if len(errors) == 0 {
+				return
+			}
+			errMsg := fmt.Sprintf("%v+", errors)
+			err := status.Error(codes.Internal, errMsg)
```
edreed, line 380: Use `status.Errorf()` instead of `fmt.Sprintf()` and `status.Error()` separately.

pkg/azureutils/logger.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (l Logger) Error(err error, msg string) {
+	l.WithCallDepth(1).Error(err, msg)
+}
+
+func (l Logger) Errorf(err error, msg string, params ...interface{}) {
+	l.WithCallDepth(1).Error(err, fmt.Sprintf(msg, params...))
+}
+
+func GetRequestID(obj runtime.Object) string {
```
edreed, line 37: This should be moved to `workflow.go` since its related to workflow and not logging.

pkg/controller/attach_detach.go
sunpa93: Given that the logger details contain the volume name and node name, wouldn't it be sufficient to query with the volume `AND` node keyword?

pkg/controller/azvolume.go
sunpa93: We add the volume name and pv name to workflow details at the beginning of reconciliation extracting them from the CRI. So our workflow should contain the volume name in every logs it outputs.
https://github.com/sunpa93/azuredisk-csi-driver/blob/be8c10064ae1de99e92c17d5bff6a12a7483a934/pkg/azureutils/workflow.go#L51

pkg/azureconstants/azure_constants.go
sunpa93: I totally forgot that we had `podNameKey`. Will switch to that.

pkg/azureutils/workflow.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	if parentWorkflow, ok := GetWorkflowFromContext(ctx); ok {
+		w.requestID = parentWorkflow.requestID
+		w.rootWorkflow = &parentWorkflow
+		w.logger = Logger{parentWorkflow.logger.WithValues(details...)}
+	} else {
+		w.requestID = GenerateRequestID()
+		w.rootWorkflow = &w
+		w.logger = Logger{klogr.New().WithValues(consts.RequestIDKey, w.requestID).WithValues(details...)}
```
edreed, line 73: Consider using `append` like in line 37 to avoid calling `WithValues` twice.

pkg/controller/azdrivernode.go
```diff
@@ -46,28 +47,29 @@ func (r *ReconcileAzDriverNode) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{Requeue: true}, nil
 	}
 
-	klog.V(2).Info("Checking to see if node (%v) exists.", request.NamespacedName)
+	ctx, workflow := azureutils.NewWorkflow(ctx, consts.NodeNameLabel, request.Name)
```
edreed, line 50: Missing `defer func () { workflow.Finish() }`?

pkg/controller/azvolume.go
```diff
@@ -535,25 +569,27 @@ func NewAzVolumeController(mgr manager.Manager, volumeProvisioner VolumeProvisio
...
 	}
 
-	klog.V(2).Info("Controller set-up successful.")
+	logger.V(2).Info("Controller set-up successful.")
 
 	return &reconciler, nil
 }
 
 func (c *SharedState) createAzVolumeFromPv(ctx context.Context, pv v1.PersistentVolume, annotations map[string]string) error {
+	ctx, workflow := azureutils.NewWorkflow(ctx)
```
edreed, line 591: Missing defer?

pkg/controller/common.go
```diff
@@ -680,6 +689,9 @@ func (p *interPodAffinityFilter) setup(pods []v1.Pod, persistentVolumes []*v1.Pe
 }
 
 func (p *interPodAffinityFilter) filter(ctx context.Context, nodes []v1.Node) ([]v1.Node, error) {
+	ctx, workflow := azureutils.NewWorkflow(ctx)
+	workflow.AddDetailToLogger("filter-plugin", p.name())
```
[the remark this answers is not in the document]
sunpa93: Oh it actually does already take variadic arguments. I missed refactoring this portion as I was unifying `StartWorkflow` and `StartChildWorkflow` functions to one.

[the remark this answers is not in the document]
sunpa93: Functional options sounds like a clean way to wrap things too.

pkg/controller/common.go
```diff
@@ -1887,14 +1973,16 @@ func (vq *VolumeReplicaRequestsPriorityQueue) DrainQueue() []*ReplicaRequest {
 
 ///Removes replica requests from the priority queue and adds to operation queue.
 func (c *SharedState) tryCreateFailedReplicas(ctx context.Context, requestor operationRequester) error {
+	workflow, _ := azureutils.GetWorkflowFromContext(ctx)
```
edreed, line 1976: This seems like a standalone workflow triggered from a specific event. I believe it make sense to create a new workflow here rather than rely on the caller to provide it.

pkg/controller/common.go
```diff
@@ -1903,11 +1991,13 @@ func (c *SharedState) tryCreateFailedReplicas(ctx context.Context, requestor ope
 }
 
 func (c *SharedState) garbageCollectReplicas(ctx context.Context, volumeName string, requester operationRequester) {
+	workflow, _ := azureutils.GetWorkflowFromContext(ctx)
```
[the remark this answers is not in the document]
sunpa93: In resonance with the previous comment, wouldn't pv controller and replica controller be the initiators?

pkg/controller/common.go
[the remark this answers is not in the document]
edreed: Ah, I see `NewContext` already does accept variadic args for details. Why not use it here? Still, I believe using the Functional Options Pattern to specify details would be a better approach.

pkg/controller/node_availability.go
```diff
@@ -43,6 +44,8 @@ func (r *ReconcileNodeAvailability) Reconcile(ctx context.Context, request recon
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	ctx, workflow := azureutils.NewWorkflow(ctx, consts.NodeNameLabel, request.Name)
```
edreed, line 47: Missing defer? But do we really need to create a workflow here if we create on in `tryCreateFailedReplicas`?

pkg/controller/pod.go
```diff
@@ -50,19 +51,21 @@ func (r *ReconcilePod) Reconcile(ctx context.Context, request reconcile.Request)
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	ctx, workflow := azureutils.NewWorkflow(ctx, consts.PodNameKey, request.Name)
```
edreed, line 54: Missing defer?

pkg/controller/pv.go
```diff
@@ -53,31 +53,34 @@ func (r *ReconcilePV) Reconcile(ctx context.Context, request reconcile.Request)
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	ctx, workflow := azureutils.NewWorkflow(ctx, consts.PvNameKey, request.Name)
```
edreed, line 56: Missing defer?

pkg/controller/node_availability.go
sunpa93: I was treating the node availability as the workflow initiator to trigger the "recreate failed replicas` workflow. In such case, the workflow does not finish at the end of `Reconcile` function.

pkg/controller/replica.go
```diff
@@ -50,35 +49,36 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
 	if !r.controllerSharedState.isRecoveryComplete() {
 		return reconcile.Result{Requeue: true}, nil
 	}
+	ctx, workflow := azureutils.NewWorkflow(ctx)
```
edreed, line 52: Missing defer?

pkg/controller/pod.go
sunpa93: Same as the comment above. I wonder if we should treat pod controller as a distinct workflow initiator.

pkg/controller/replica.go
```diff
@@ -137,23 +138,23 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
 	return reconcile.Result{}, nil
 }
 
-func (r *ReconcileReplica) triggerGarbageCollection(volumeName string) {
-	emptyCtx := context.Background()
-	deletionCtx, cancelFunc := context.WithCancel(emptyCtx)
+//nolint:contextcheck // Garbage collection is asynchronous; context is not inherited by design
+func (r *ReconcileReplica) triggerGarbageCollection(ctx context.Context, volumeName string) {
+	workflowCtx, workflow := azureutils.NewWorkflow(ctx)
```
edreed, line 143: Missing defer?

pkg/controller/node_availability.go
edreed: > I was treating the node availability as the workflow initiator to trigger the "recreate failed replicas `workflow`.

Right now, the node availability controller has only one task, but in the future may have more. In general, we should feel free to view a reconciler as the workflow initiator when has it has one purpose. However, this reconciler starts an asynchronous to its execution and I view it as more akin to an event trigger than an initiator of activitiy. Also, we are unconditionally creating a new workflow when we only need to if we actually call `triggerCreateFailedReplicas`.

In the future, we may also want to trigger replica recreation from other controllers on different events than schedulability. In that case, starting the workflow in `triggerCreateFailedReplicas` makes more sense.

> In such case, the workflow does not finish at the end of `Reconcile` function.

We shouldn't rely on other functions to complete the workflow. It makes tracking down the lifetime difficult.

sunpa93: Makes sense 👍

pkg/controller/common.go
edreed: See my comment in the node availability controller.

pkg/controller/common.go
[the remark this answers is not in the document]
edreed: I see them more as event triggers than initiators since they trigger other workflows as well.

pkg/azureutils/workflow.go
```diff
@@ -0,0 +1,178 @@
+package azureutils
```
edreed, line 1: I wonder if we should move `Workflow` and `Logger` to its own package, `workflow`. That way, creating a new workflow could be written a little more naturally as:

```go
w := workflow.New(ctx, workflow.WithDetails(...))
```

It would mean renaming all of the existing `workflow` variables, though. ☹️

sunpa93: Wouldn't be a problem :D

pkg/controller/azvolume.go
nearora-msft: Makes sense. Since we are creating new worklfow from the object in this function, I wonder if it would be more clear to name it NewWorkflowFromObj  instead?

sunpa93: Actually, we are creating a child workflow of the root workflow that has been stored to context in `Reconcile` function.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -368,19 +371,28 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
 	// Leader controller manager should recover CRI if possible and clean them up before exiting.
 	go func() {
 		<-mgr.Elected()
+		var errors []error
+		ctx, w := workflow.New(ctx)
+		defer func() {
+			if len(errors) == 0 {
+				return
```
edreed, line 378: Shouldn't we `w.Finish(nil)` here?

pkg/workflow/workflow.go
```diff
@@ -0,0 +1,232 @@
...
+	startTime      time.Time
+	requestID      string
+	logger         Logger
+	parentWorkflow *Workflow
+	numWait        int32 // numWait should always be number of children + 1(itself)
+}
+
+type Option func(*Workflow)
+
+func WithDetails(details ...interface{}) Option {
```
edreed, line 36: Let's keep options functions grouped with `New` for easy location and reference.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -368,19 +371,28 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
...
+			}
+			err := status.Errorf(codes.Internal, "%v+", errors)
```
edreed, line 380: You've added support to `Finish` to handle multiple errors, so this is no longer necessary.

pkg/workflow/workflow.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	if w.numWait > 0 {
+		return
+	}
+
+	latency := time.Since(w.startTime)
+	errs = w.errors()
+	if len(errs) == 0 {
+		err := status.Errorf(codes.Internal, "%+v", errs)
+		w.logger.WithValues(consts.Latency, latency, "workflow", w.name).WithCallDepth(1).Error(err, "Workflow completed with an error.")
```
edreed, line 122: "workflow" should have a constant.

pkg/workflow/workflow.go
```diff
@@ -0,0 +1,232 @@
...
+	return w
+}
+
+func (w Workflow) Finish(errs ...error) {
+	if len(errs) > 0 {
+		for _, err := range errs {
+			w.errSet.Store(err, nil)
+		}
+	}
+	if w.numWait > 0 {
```
edreed, line 114: Since the `numWait` count includes this instance, should the atomic decrement happen here?

We should defend against calling `Finish` too many times and `panic` is the count goes negative.

pkg/workflow/workflow.go
```diff
@@ -0,0 +1,232 @@
...
+	errs = w.errors()
+	if len(errs) == 0 {
+		err := status.Errorf(codes.Internal, "%+v", errs)
+		w.logger.WithValues(consts.Latency, latency, "workflow", w.name).WithCallDepth(1).Error(err, "Workflow completed with an error.")
+	} else {
+		w.logger.WithValues(consts.Latency, latency, "workflow", w.name).WithCallDepth(1).Info("Workflow completed with success.")
+	}
+
+	if w.parentWorkflow != nil {
+		_ = atomic.AddInt32(&w.parentWorkflow.numWait, -1)
```
edreed, line 128: I believe the decrement should happen in the `Finish` method itself.

pkg/workflow/workflow.go
```diff
@@ -0,0 +1,232 @@
...
+type workflowKey struct{}
+
+type Workflow struct {
+	name           string
+	errSet         *sync.Map
+	startTime      time.Time
+	requestID      string
+	logger         Logger
+	parentWorkflow *Workflow
+	numWait        int32 // numWait should always be number of children + 1(itself)
```
edreed, line 31: Consider `pendingCount` since we're not actually waiting on anything.

pkg/controller/common.go
```diff
@@ -284,7 +286,9 @@ func (c *SharedState) createOperationQueue(volumeName string) {
 	_, _ = c.volumeOperationQueues.LoadOrStore(volumeName, newLockableEntry(newOperationQueue()))
 }
 
-func (c *SharedState) addToOperationQueue(volumeName string, requester operationRequester, operationFunc func() error, isReplicaGarbageCollection bool) {
+func (c *SharedState) addToOperationQueue(ctx context.Context, volumeName string, requester operationRequester, operationFunc func(context.Context) error, isReplicaGarbageCollection bool) {
+	ctx, w := workflow.New(ctx, workflow.WithDetails(consts.VolumeNameLabel, volumeName))
```
edreed, line 290: While it is terribly convenient and reasonable to create the workflow here, `addToOperationQueue` will be recorded as the initiator rather than the caller (i.e. `createReplicas` and `triggerManageReplicas`) which isn't very information. Consider adding a `workflow.WithCaller(skip int)` option that would allow us to capture a caller explicitly here, e.g.:

```go
ctx, w := workflow.New(cts, workflow.WithCaller(1), workflow.WithDetails(consts.VolumeNameLabel, volumeName))
```

pkg/workflow/workflow.go
sunpa93: Makes sense.

pkg/workflow/workflow.go
```diff
@@ -0,0 +1,232 @@
...
+	case *diskv1beta1.AzVolumeAttachment:
+		if target != nil {
+			return []interface{}{consts.VolumeNameLabel, target.Spec.VolumeName, consts.NodeNameLabel, target.Spec.NodeName, consts.RoleLabel, target.Spec.RequestedRole}
+		}
+	}
+	return []interface{}{}
+}
+
+// New starts a new child workflow if there is already a workflow saved to the context otherwise a new workflow
+func New(ctx context.Context, opts ...Option) (context.Context, Workflow) {
```
edreed, line 73: When using the Functional Options Pattern, I like to structure my `New` function so that options are processed first and default values are applied afterward. This way, you avoid duplicating work if an option overrides a value. You can look at my implementation of [`batch.NewProcessor`](https://github.com/edreed/cloud-provider-azure/blob/817d5114fc89f8b9e8204c4138a8c4f42b093d7d/pkg/batch/batch.go#L350) as an example.

pkg/controller/replica.go
```diff
@@ -137,24 +126,39 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
 	return reconcile.Result{}, nil
 }
 
-func (r *ReconcileReplica) triggerGarbageCollection(volumeName string) {
-	emptyCtx := context.Background()
-	deletionCtx, cancelFunc := context.WithCancel(emptyCtx)
+//nolint:contextcheck // context is not inherited by design
+func (r *ReconcileReplica) triggerManageReplica(ctx context.Context, volumeName string) {
+	manageReplicaCtx, w := workflow.New(context.Background(), workflow.WithDetails(consts.VolumeNameLabel, volumeName))
```
edreed, line 131: Since `addToOperationQueue` creates workflow, is it necessary to also create one here?

pkg/controller/common.go
edreed: Oh, I see now that the callers both create parent workflows and this simply creates a child workflow for the queueing operation. Setting the caller explicitly here wouldn't make sense in that case.

pkg/controller/replica.go
sunpa93: As there are several controllers that can trigger `manageReplica` workflow, I thought it would make a useful log to see which one the workflow came from.

pkg/controller/pod.go
```diff
@@ -51,18 +55,19 @@ func (r *ReconcilePod) Reconcile(ctx context.Context, request reconcile.Request)
...
 		if errors.IsNotFound(err) {
-			klog.V(5).Infof("Failed to reconcile pod %s. Pod was deleted.", request.Name)
+			logger.V(5).Info("No need to reconcile. Pod was deleted.")
 			if err := r.controllerSharedState.deletePod(ctx, podKey); err != nil {
 				return reconcile.Result{Requeue: true}, err
 			}
 			return reconcile.Result{}, nil
 		}
-		klog.Errorf("Error getting the pod %s. Error: %v", podKey, err)
+		r.controllerSharedState.eventRecorder.Eventf(pod.DeepCopyObject(), corev1.EventTypeWarning, consts.ClientFailedGetEvent, "Failed to get pod (%s): %v", request.NamespacedName.String(), err)
```
edreed, line 70: I don't believe we should create an event with the error against the pod since it is not indicative of the pod's condition, but rather an error between this controller and the API server. We should just log an error here like before.

pkg/controller/replica.go
edreed: Yeah, I see that now.

go.mod
```diff
@@ -45,6 +45,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require github.com/go-logr/logr v1.2.2
```
edreed, line 48: Please move this to the first `require` section.

pkg/workflow/workflow.go
```diff
@@ -0,0 +1,235 @@
...
+type Option func(*Workflow)
+
+func WithDetails(details ...interface{}) Option {
+	return func(w *Workflow) {
+		w.logger = Logger{w.logger.WithValues(details...)}
+	}
+}
+
+// New starts a new child workflow if there is already a workflow saved to the context otherwise a new workflow
+func New(ctx context.Context, opts ...Option) (context.Context, Workflow) {
```
edreed, line 76: We should manage `Workflow` by-reference rather than by-value. Otherwise, the completion won't work correctly in the multi-child case since each child would get new copy of the parent from `GetWorkflowFromContext`.

pkg/workflow/workflow_test.go
```diff
@@ -0,0 +1,60 @@
...
+			description: "Parent workflow should only finish when child workflow finishes",
+			verifyFunc: func() {
+				ctx, parentWorkflow := New(context.Background())
+				_, childWorkflow := New(ctx)
+				parentWorkflow.Finish()
+				require.Equal(t, atomic.LoadInt32(parentWorkflow.pendingCount), int32(1))
+				childWorkflow.Finish()
+				require.Equal(t, atomic.LoadInt32(parentWorkflow.pendingCount), int32(0))
+			},
+		},
```
edreed, line 51: It would be good to add a test case for multiple children.

edreed:
@sunpa93 Is this change ready? If so, could you remove the "Draft" status?

sunpa93:
> @sunpa93 Is this change ready? If so, could you remove the "Draft" status?

It should be. I have removed the "Draft" status.

edreed:
@sunpa93 The `go.mod` file needs to be updated:

```shell
go mod tidy
go mod vendor
```

sunpa93:
> > @sunpa93 The `go.mod` file needs to be updated:
> > ```shell
> > go mod tidy
> > go mod vendor
> > ```
> 
> Just realized that I only made the change locally and did not push. Pushed.

edreed:
The linter is failing the static checks due to unrelated code.

PR #1290, opened 2022-04-19
[V2] fix: Fix verify-examples.sh leak resources created 🐞

### LINKED ISSUE #1288: [V2] verify-examples.sh leaks deployment and stateful set when running e2e tests
**What happened**:
When running e2e tests, `/hack/verify_examples.sh` runs as a part of the `ginkgo.AfterSuite` test hook in `test/e2e/suite_test.go`. This script leaks a deployment and stateful set that can interfere to subsequent test runs that expect a clean system.

I expect there to be no deployment or stateful set left over after running an e2e test. 

**How to reproduce it**:
1. Run an e2e test
2. Execute `kubectl get all`
3. Observer that there is a deployment and stateful set in the default namespace after the test run.

This is part of the testing infrastructure and not related to features or functionality.

**Environment**:
- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

jmclong:
**What type of PR is this?**
/kind cleanup

When running e2e tests interactively, `/hack/verify-examples.sh` runs as a part of the `ginkgo.AfterSuite` test hook in `test/e2e/suite_test.go`. This script leaks a deployment and stateful set that can interfere to subsequent test runs that expect a clean system.

Fixes #1288 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/hack/verify-examples.sh
+++ b/hack/verify-examples.sh
@@ -76,8 +76,8 @@ fi
 
 cleanup_examples() {
     for EXAMPLE in "${EXAMPLES[@]}"; do
-        echo "Attempting deletion of \"EXAMPLE\""
-        kubectl delete -f "$EXAMPLE"
+        echo "Attempting deletion of \"$EXAMPLE\""
+        kubectl delete -f "$EXAMPLE" --ignore-not-found
     done
 }
 
--- a/hack/verify-examples.sh
+++ b/hack/verify-examples.sh
@@ -74,8 +74,17 @@ if [[ "$1" == "windows" ]]; then
     )
 fi
 
+cleanup_examples() {
+    for EXAMPLE in "${EXAMPLES[@]}"; do
+        echo "Attempting deletion of \"EXAMPLE\""
+        kubectl delete -f "$EXAMPLE"
+    done
+}
+
+trap cleanup_examples EXIT
+
 for EXAMPLE in "${EXAMPLES[@]}"; do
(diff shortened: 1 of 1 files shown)


hack/verify-examples.sh
```diff
@@ -74,8 +74,17 @@ if [[ "$1" == "windows" ]]; then
...
+        echo "Attempting deletion of \"$EXAMPLE\""
+        kubectl delete -f "$EXAMPLE" --ignore-not-found
+    done
+}
+
+trap cleanup_examples EXIT
+
 for EXAMPLE in "${EXAMPLES[@]}"; do
-    rollout_and_wait $EXAMPLE
+    rollout_and_wait "$EXAMPLE"
```
nearora-msft, line 87: Why is this change needed?

jmclong: If there's a space in $EXAMPLE it will split into multiple args. For example

```bash
#!/usr/bin/env bash

hello_world() {
  echo $1
}
VAR="Hello, world"
hello_world $VAR
hello_world "$VAR"
```
Prints:
```
Hello,
Hello, world!
```

I looked at the context, and it was clear that $EXAMPLE is meant to be a single argument, so I fixed it.

nearora-msft: Ah ok, thanks for explaining!

nearora-msft (approved):

PR #1294, opened 2022-04-21
[V2] bug: failed operation should not be requeued if operation queue is scheduled for deletion

sunpa93:
**What type of PR is this?**
/kind bug

* With current implementation, failed operation can end up getting requeued to a "to be deleted" operation queue if operation queue is cleaned up in the middle of the operation.
* This can lead to situations where an operation queue for a deleted volume continues rolling with failure and flooding logs and wasting resources.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/azvolume.go
+++ b/pkg/controller/azvolume.go
@@ -201,7 +201,7 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *diskv1b
 	}
 
 	// override volume operation queue to prevent any other replica operation from being executed
-	release := r.controllerSharedState.overrideAndClearOperationQueue(azVolume.Name)
+	release := r.controllerSharedState.closeOperationQueue(azVolume.Name)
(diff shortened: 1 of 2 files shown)


sunpa93:
@edreed

pkg/controller/azvolume.go
```diff
@@ -201,7 +201,7 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *diskv1b
 	}
 
 	// override volume operation queue to prevent any other replica operation from being executed
-	release := r.controllerSharedState.overrideAndClearOperationQueue(azVolume.Name)
+	release := r.controllerSharedState.cleanOperationQueue(azVolume.Name)
```
edreed, line 204: `closeOperationQueue` would better communicate intended use.

PR #1300, opened 2022-04-26
[V2] chore: DeleteVolume should wait for replicas to be fully deleted

sunpa93:
instead of requeuing

**What type of PR is this?**
/kind chore

* `AzVolume` controller originally relied on reconciler requeue to wait for replica `AzVolumeAttachment` to be fully deleted but this can lead to huge backoffs causing a significant delay in volume deletion.
* This PR leverages condition watcher / waiter component to have `AzVolume` controller wait for the replica AzVolumeAttachments to be deleted to avoid over issuance of requeue
* And replaces existing polling in replica controller with event triggered waiter package.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/common.go
+++ b/pkg/controller/common.go
@@ -1214,12 +1214,15 @@ func (s *scoreByNodeCapacity) score(ctx context.Context, nodeScores map[string]i
 			// if failed to get node's remaining capacity, remove the node from the candidate list and proceed
 			w.Logger().Errorf(err, "failed to get remaining capacity of node (%s)", nodeName)
 			delete(nodeScores, nodeName)
-		}
-		if remainingCapacity-len(s.volumes) <= 0 {
-			delete(nodeScores, nodeName)
+			continue
 		}
 
 		nodeScores[nodeName] = score + (nodeScoreLowCoefficient * remainingCapacity)
+
+		if remainingCapacity-len(s.volumes) < 0 {
+			delete(nodeScores, nodeName)
+		}
+
 		w.Logger().V(5).Infof("node (%s) can accept %d more attachments", nodeName, remainingCapacity)
 	}
 	return nodeScores, nil
@@ -1363,9 +1366,11 @@ func (c *SharedState) prioritizeNodes(ctx context.Context, pods []v1.Pod, volume
 	}
 
 	// normalize score
+	numFiltered := 0
 	for _, node := range nodes {
 		if _, exists := nodeScores[node.Name]; !exists {
 			nodeScores[node.Name] = -1
+			numFiltered++
 		}
 	}
 
@@ -1373,7 +1378,7 @@ func (c *SharedState) prioritizeNodes(ctx context.Context, pods []v1.Pod, volume
 		return nodeScores[nodes[i].Name] > nodeScores[nodes[j].Name]
 	})
 
-	return nodes
+	return nodes[:len(nodes)-numFiltered]
 }
 
 func (c *SharedState) filterAndSortNodes(ctx context.Context, nodes []v1.Node, pods []v1.Pod, volumes []string) ([]v1.Node, error) {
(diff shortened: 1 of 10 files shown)


pkg/watcher/conditionwatcher.go
```diff
@@ -61,18 +61,18 @@ type waitEntry struct {
...
 
-type conditionWatcher struct {
+type ConditionWatcher struct {
 	informerFactory azurediskInformers.SharedInformerFactory
 	waitMap         sync.Map // maps namespaced name to waitEntry
 	namespace       string
 }
 
-func newConditionWatcher(ctx context.Context, azDiskClient azDiskClientSet.Interface, informerFactory azurediskInformers.SharedInformerFactory, namespace string) *conditionWatcher {
+func NewConditionWatcher(ctx context.Context, azDiskClient azDiskClientSet.Interface, informerFactory azurediskInformers.SharedInformerFactory, namespace string) *ConditionWatcher {
```
edreed, line 70: Now that this is in a `watcher` package, it makes sense to rename this simply `New` to prevent stutter. That way we can simply say `watcher.New(...)`.

pkg/watcher/conditionwatcher.go
```diff
@@ -35,12 +35,12 @@ import (
 	azurediskInformers "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/client/informers/externalversions"
 )
 
-type objectType string
+type ObjectType string
```
edreed, line 38: `ObjectType` doesn't seem to be exposed outside of this package; does it really need to be a package public symbol? Same with its constants below.

sunpa93: We need object type when we initialize a new `ConditionWaiter`(e.g. https://github.com/sunpa93/azuredisk-csi-driver/blob/c26a3adf73fba8b2f81129eaa2a79b827495cf8b/pkg/provisioner/crdprovisioner.go#L244)

edreed: nm. I see that these are exposed outside of the package as a parameter to `NewConditionWaiter`.

pkg/controller/common.go
```diff
@@ -2177,3 +2181,16 @@ func (c *SharedState) getNodesForReplica(ctx context.Context, volumeName string,
 
 	return filtered, nil
 }
+
+func verifyDeleted(obj interface{}, objectDeleted bool) (bool, error) {
```
edreed, line 2185: Consider `verifyAzVolumeAttachmentDeleted` since this is specific to that object.

pkg/watcher/conditionwaiter.go
```diff
@@ -23,24 +23,24 @@ import (
...
-	case azVolumeType:
+	case AzVolumeType:
 		obj, err = w.watcher.informerFactory.Disk().V1beta1().AzVolumes().Lister().AzVolumes(namespace).Get(w.objName)
-	case azVolumeAttachmentType:
+	case AzVolumeAttachmentType:
 		obj, err = w.watcher.informerFactory.Disk().V1beta1().AzVolumeAttachments().Lister().AzVolumeAttachments(namespace).Get(w.objName)
-	case azDriverNodeType:
+	case AzDriverNodeType:
 		obj, err = w.watcher.informerFactory.Disk().V1beta1().AzDriverNodes().Lister().AzDriverNodes(namespace).Get(w.objName)
 	}
```
edreed, line 46: I believe we may have an issue here if the previous `Get` calls fail due to `errors.NotFound`. In the case where we are waiting for the delete of an object, this would cause the wait to fail when it should really succeed. I believe we should convert a "not found" error into a call of `conditionFunc` with `objectDeleted` set to `true`.

sunpa93: You are right. Good catch! 👍

sunpa93:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/19bcb28179f26c975d758a670028221a1df09407/pkg/controller/replica.go#L177-L188

Probably can replace this polling with conditionWaiter

PR #1305, opened 2022-04-27
[V2] chore: e2e test util fixes

sunpa93:
**What type of PR is this?**
/kind cleanup

* e2e test relies on storage class parameters for many of its functionalities and the sku name field is handled case sensitive in these test utils when it should be case insensitive.
* This is causing certain bugs as below:
A ZRS disk is created for multi-az case but with storage class parameter sku name field hard coded.
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/04aba1e20ce3598d77e359198e4579335eb27024/test/e2e/dynamic_provisioning_test.go#L1652
But because the test util expects `consts.SkuNameField`, which is in lowercase, as storageclass parameter key, we never end up setting the volume's binding mode to `VolumeBindingImmediate`
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/04aba1e20ce3598d77e359198e4579335eb27024/test/e2e/driver/azuredisk_driver.go#L62-L64
* And for code manageability, hard coded values were replaced with constant except for one test case where it was left as is to verify that test util handles storage class parameters case insensitively.
* And some tester and test util functions were amended as they were relying on some invalid assumptions.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

This is not a v2 specific problem. Once the change is approved for v2, the commit can be cherry picked for v1 as well.

**Release note**:
```
none
```

### DIFF
--- a/test/e2e/driver/azuredisk_driver.go
+++ b/test/e2e/driver/azuredisk_driver.go
@@ -59,10 +59,20 @@ func (d *azureDiskDriver) GetDynamicProvisionStorageClass(parameters map[string]
 	provisioner := d.driverName
 	generateName := fmt.Sprintf("%s-%s-dynamic-sc-", namespace, normalizeProvisioner(provisioner))
 	var allowedTopologies []v1.TopologySelectorTerm
-	if skuName, ok := parameters[consts.SkuNameField]; ok && strings.HasSuffix(skuName, "_ZRS") {
-		mode := storagev1.VolumeBindingImmediate
-		bindingMode = &mode
-	} else {
+	isZRS := false
+	for key, value := range parameters {
+		if strings.EqualFold(key, consts.SkuNameField) {
+			if strings.HasSuffix(value, "_ZRS") {
+				mode := storagev1.VolumeBindingImmediate
+				bindingMode = &mode
+				isZRS = true
+			} else if strings.EqualFold(os.Getenv("AZURE_CLOUD_NAME"), "AZURESTACKCLOUD") && !strings.EqualFold(value, "Standard_LRS") && !strings.EqualFold(value, "Premium_LRS") {
+				parameters[consts.SkuNameField] = "Standard_LRS"
+			}
+			break
+		}
+	}
+	if !isZRS {
 		if len(allowedTopologyValues) > 0 {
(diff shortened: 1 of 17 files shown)


andyzhangx:
in such case, it's a test case issue since both `skuname` and `skuName` is allowed (all storage class parameter names are case insensitive)

andyzhangx:
> @andyzhangx I understand that storage class parameters names are case insensitive but the e2e test utils that we rely on seem to be case sensitive. `GetDynamicStorageClass` in both v1 and v2 branch check whether a specific string value exists in the storageclass parameter map to take certain actions. I think these util functions should probably be amended so that they handle parameters case insensitive and using a const instead of hardcoded value for these parameter keys would help us maintain the code. What do you think?

@sunpa93 yes, that's test case issue, the test case should handle parameters case insensitive

test/e2e/driver/azuredisk_driver.go
```diff
@@ -59,10 +59,22 @@ func (d *azureDiskDriver) GetDynamicProvisionStorageClass(parameters map[string]
...
+	isZRS := false
+	for key, value := range parameters {
+		if strings.EqualFold(key, consts.SkuNameField) {
+			if strings.HasSuffix(value, "_ZRS") {
+				mode := storagev1.VolumeBindingImmediate
+				bindingMode = &mode
+				isZRS = true
+			}
+
+			if strings.EqualFold(os.Getenv("AZURE_CLOUD_NAME"), "AZURESTACKCLOUD") && !strings.EqualFold(value, "Standard_LRS") && !strings.EqualFold(value, "Premium_LRS") {
```
edreed, line 71: This can be `else if` because we know the value is not "Standard_LRS" or "Premium_LRS" if the first `if` clause was true.

sunpa93:
e2e-single-az-mainv2 test failure debrief:
1. Deployment created
2. PublishVolume call for node-0 and node-1 fails in first attempt due to surpassing 5 minutes cloud timeout.
3. PublishVolume succeeds on its second attempt but at this point, the test has already failed as `Deployment.Create` waits for 5 minutes for deployment to be complete.

PR #1306, opened 2022-04-28
[V2] test: add unit tests to az-analyze

hccheng72:
**What type of PR is this?**
/kind test

Add unit tests to az-analyze tool

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


cmd/az-analyze/cmd/azv_test.go
```diff
@@ -0,0 +1,194 @@
...
+    tests := []struct {
+        description string
+        verifyFunc func()
+    }{
+        {
+            description: "specified pod name with more than one pv",
+            verifyFunc: func() {
+                result := GetAzVolumesByPod(fakeClientsetK8s, fakeClientsetAzDisk, "test-pod-0", "default")
+                expect := []AzvResource {
+                    {
```
nearora-msft, line 119: We can create AzvResource outside of each test case, so that we can reuse them for all test cases.

e.g. azvResource1 :=  {
                        ResourceType: "test-pod-0",
                        Namespace:    consts.DefaultAzureDiskCrdNamespace,
                        Name:         "test-azVolume",
                        State:        v1beta1.VolumeCreated,
                    },

 azvResource2 := azvResource1 :=  {
                        ResourceType: "test-pod-0",
                        Namespace:    consts.DefaultAzureDiskCrdNamespace,
                        Name:         "test-azVolume-1",
                        State:        v1beta1.VolumeCreated,
                    },
 And then under expect for each test case, we can write []AzvResource{azvResource1 , azvResource2}

cmd/az-analyze/cmd/azv_test.go
```diff
@@ -0,0 +1,194 @@
...
+    fakeClientsetAzDisk := NewTestAzDiskClientset()
+
+    tests := []struct {
+        description string
+        verifyFunc func()
+    }{
+        {
+            description: "specified pod name with more than one pv",
+            verifyFunc: func() {
+                result := GetAzVolumesByPod(fakeClientsetK8s, fakeClientsetAzDisk, "test-pod-0", "default")
```
nearora-msft, line 117: We should use the constant for the default namespace.

cmd/az-analyze/cmd/common_test.go
```diff
@@ -0,0 +1,170 @@
+/*
```
nearora-msft, line 1: We should rename this file too test_util.go or something like that, common_test would imply that this file contains unit tests for the functions in common.go.

cmd/az-analyze/cmd/common_test.go
```diff
@@ -0,0 +1,170 @@
...
+			Namespace: consts.DefaultAzureDiskCrdNamespace,
+		},
+		Spec: v1beta1.AzVolumeSpec{
+			VolumeName: "test-azVolume-0",
+			Parameters: map[string]string{consts.PvcNameKey: "test-pvcClaimName-0"},
+		},
+		Status: v1beta1.AzVolumeStatus{
+			State: v1beta1.VolumeCreated,
+		},
+	}, &v1beta1.AzVolume{
```
nearora-msft, line 100: It would be cleaner to create slices for AzVolume and AzVolumeAttachment separately and then add those to this function. Otherwise, if we want to add more objects later, this definition will keep getting bigger.

cmd/az-analyze/cmd/common_test.go
```diff
@@ -0,0 +1,170 @@
...
+)
+
+func NewTestK8sClientset() *fake.Clientset {
+
+	return fake.NewSimpleClientset(&v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-pod-0",
+			Namespace: "default",
+		},
+		Spec: v1.PodSpec{
```
nearora-msft, line 37: Similar comment as below. It would be cleaner to add the slices for the objects separately.

cmd/az-analyze/cmd/common_test.go
```diff
@@ -0,0 +1,170 @@
...
+			Volumes: []v1.Volume{
+				{
+					Name: "volume-name-0",
+					VolumeSource: v1.VolumeSource{
+						PersistentVolumeClaim: &v1.PersistentVolumeClaimVolumeSource{
+							ClaimName: "test-pvcClaimName-0",
+						},
```
nearora-msft, line 44: We could also create constants for the pvc names as they are being used across the file.

cmd/az-analyze/cmd/azva_test.go
```diff
@@ -0,0 +1,345 @@
...
+func TestGetAllAzVolumeAttachements(t *testing.T) {
+	fakeClientsetK8s := NewTestK8sClientset()
+	fakeClientsetAzDisk := NewTestAzDiskClientset()
+
+	tests := []struct {
+		description string
+		verifyFunc  func()
+	}{
+		{
+			description: "specified namespace",
```
nearora-msft, line 153: The description can be more descriptive, like "test get all azvolumeattachments when a namespace is specified"

cmd/az-analyze/cmd/azva.go
```diff
@@ -116,20 +123,15 @@ type AzvaResourceAll struct {
 }
 
 // return azVolumeAttachements with all Pods/Nodes/Zones when no flags is provided
-func GetAllAzVolumeAttachements(namespace string) []AzvaResourceAll {
-	// access to config and Clientsets
-	config := getConfig()
-	clientsetK8s := getKubernetesClientset(config)
-	clientsetAzDisk := getAzDiskClientset(config)
-
+func GetAllAzVolumeAttachements(clientsetK8s kubernetes.Interface, clientsetAzDisk versioned.Interface, namespace string) []AzvaResourceAll {
```
nearora-msft, line 126: Ideally, we should've had a single resource with all the fields(nodes,zone,pod) instead of having 2 resources called AzvaResource and AzvaResourceAll. In case a user wants to get the azva by pod, node and zone would've been empty. Similarly, if node/zone is specified, pod,zone/node,pod would've been empty. But it might require some refactoring, so we can take it up later in case we've some time left at the end.

cmd/az-analyze/cmd/azva.go
```diff
@@ -116,20 +123,15 @@ type AzvaResourceAll struct {
...
 	result := make([]AzvaResourceAll, 0)
 
 	if namespace == "" {
 		namespace = "default"
 	}
 
 	// get pvc claim names of pod(s)
-	pvcClaimNameSet := make(map[string]string)
+	pvcClaimNameSet := make(map[string][]string)
```
nearora-msft, line 134: Some of the code for  GetAllAzVolumeattachments and GetAzVolumeAttachmentsByPod seems same. Would it be possible to take that part out in a function and resuse that function in both the GetAllAzVolumeattachments and  GetAzVolumeAttachmentsByPod .

cmd/az-analyze/cmd/azva_test.go
```diff
@@ -0,0 +1,345 @@
...
+		verifyFunc  func()
+	}{
+		{
+			description: "specified zone name with one node",
+			verifyFunc: func() {
+				result := GetAzVolumeAttachementsByZone(fakeClientsetK8s, fakeClientsetAzDisk, "eastus-0")
+				expect := []AzvaResource{azvaResource_zone1}
+
+				for i := 0; i < len(result); i++ {
+					require.Equal(t, result[i].ResourceType, expect[i].ResourceType)
```
nearora-msft, line 312: We could take this common part in a function and reuse it across the test functions.

cmd/az-analyze/cmd/azva.go
```diff
@@ -216,28 +215,26 @@ func GetAzVolumeAttachementsByPod(podName string, namespace string) []AzvaResour
...
+					Age:          metav1.Now().Sub(azVolumeAttachment.CreationTimestamp.Time),
+					RequestRole:  azVolumeAttachment.Spec.RequestedRole,
+					Role:         azVolumeAttachment.Status.Detail.Role,
+					State:        azVolumeAttachment.Status.State,
+				})
+			}
 		}
 	}
-
+	fmt.Println(result) // debug
```
nearora-msft, line 232: Do we need to print the result here?

hccheng72: No, just for debugging purposes. All of them will be deleted eventually.

cmd/az-analyze/cmd/azva_test.go
```diff
@@ -0,0 +1,317 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1beta1 "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/azuredisk/v1beta1"
+	consts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
+)
+
+var azvaResourceAll0 = AzvaResource{
```
nearora-msft, line 29: Now that we have removed AzvaResourceAll and are using AzvaResource for all the scenarios, do we really need to have different azvaResource by pod, zone, node, all etc? Can't we have a few AzvaResource objects that have the information for pods,nodes,zones? We can certainly verify the result based on what we are testing. Like if we are testing GetAzvaByZone, we would only match the zone. Same for pod and node.

nearora-msft: Looking more closely at it, the current implementation seems fine to me. Because the actual output will have the fields empty. I'll resolve this.

hccheng72: If the result is from GetAzvaByZone, its PodName and NodeName should be empty string. So don't we check that?

nearora-msft (approved):

cmd/az-analyze/cmd/azva_test.go
nearora-msft: Yeah, we can resolve this!

PR #1320, opened 2022-05-06
[WIP][V2] Logging Analysis Tool

hccheng72:
**What type of PR is this?**
/kind feature

An integrative tool that fetches and parses appropriate log lines to bring better insight of the V2 driver behavior.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/cmd/az-log/cmd/common.go
+++ b/cmd/az-log/cmd/common.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"path/filepath"
+	"time"
+
+	homedir "github.com/mitchellh/go-homedir"
+	"github.com/spf13/cobra"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	consts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
+)
+
+func GetFlags(cmd *cobra.Command) ([]string, []string, []string, string, bool, bool){
+	volumes, _ := cmd.Flags().GetStringSlice("volume")
+	nodes, _ := cmd.Flags().GetStringSlice("node")
+	requestIds, _ := cmd.Flags().GetStringSlice("request-id")
+	afterTime, _ := cmd.Flags().GetString("after-time")
+	isWatch, _ := cmd.Flags().GetBool("volume")
+	isPrevious, _ := cmd.Flags().GetBool("volume")
+
+	return volumes, nodes, requestIds, afterTime, isWatch, isPrevious
+}
+
+func GetLogsByAzDriverPod(podName string, container string, volumes []string, nodes []string, requestIds []string, afterTime string, isWatch bool, isPrevious bool) {
+	var kubeconfig string
+	if home, _ := homedir.Dir(); home != "" {
+		kubeconfig = filepath.Join(home, ".kube", "config")
+	}
+
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	if err != nil {
+		panic(err.Error())
+	}
+
+	clientsetK8s, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
+	}
+
+	t, _ := time.Parse(time.RFC3339, afterTime)
+	sinceTime := metav1.NewTime(t)
+
+	podLogOptions := v1.PodLogOptions{
+        Container: container,
+		Previous: isPrevious,
+		SinceTime: &sinceTime,
+    }
+
+	req := clientsetK8s.CoreV1().Pods(consts.ReleaseNamespace).GetLogs(podName, &podLogOptions)
+	podLogs, err := req.Stream(context.TODO())
+	if err != nil {
+		panic(err.Error())
+	}
+
+	defer podLogs.Close()
+
+	// buf := new(bytes.Buffer)
+    // _, err = io.Copy(buf, podLogs)
+    // if err != nil {
+    //     panic(err.Error())
+    // }
+    // str := buf.String()
+
+	buf := bufio.NewScanner(podLogs)
+	for buf.Scan() {
+        fmt.Println(buf.Text()) // TODO: filter volumes, nodes, requestIds
+    }
+}
+
--- a/cmd/az-log/cmd/controller.go
+++ b/cmd/az-log/cmd/controller.go
@@ -1,0 +1,65 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+// controllerCmd represents the controller command
+var controllerCmd = &cobra.Command{
+	Use:   "controller",
+	Short: "A brief description of your command",
+	Long: `A longer description that spans multiple lines and likely contains examples
+and usage of using your command. For example:
+
+Cobra is a CLI library for Go that empowers applications.
+This application is a tool to generate the needed files
+to quickly create a Cobra application.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		volumes, _ := cmd.Flags().GetStringSlice("volume")
+		nodes, _ := cmd.Flags().GetStringSlice("node")
+		requestIds, _ := cmd.Flags().GetStringSlice("request-id")
+		afterTime, _ := cmd.Flags().GetString("after-time")
+		isWatch, _ := cmd.Flags().GetBool("volume")
+		isPrevious, _ := cmd.Flags().GetBool("volume")
+		fmt.Println("controller called")
+	},
+}
+
+func init() {
+	getCmd.AddCommand(controllerCmd)
+
+	// Here you will define your flags and configuration settings.
+
+	// Cobra supports Persistent Flags which will work for this command
+	// and all subcommands, e.g.:
+	// controllerCmd.PersistentFlags().String("foo", "", "A help for foo")
+
+	// Cobra supports local flags which will only run when this command
+	// is called directly, e.g.:
+	// controllerCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
+}
+
+func getControllerPodName() string {
+	return ""
+}
+
+func GetLogsByConttroler(podName string, volumes []string, nodes []string, ) {
+
+}
--- a/cmd/az-log/cmd/controller.go
+++ b/cmd/az-log/cmd/controller.go
@@ -1,5 +1,5 @@
 /*
-Copyright © 2022 NAME HERE <EMAIL ADDRESS>
+Copyright 2022 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -13,11 +13,10 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
+
 package cmd
 
 import (
-	"fmt"
-
 	"github.com/spf13/cobra"
 )
 
@@ -32,34 +31,17 @@ Cobra is a CLI library for Go that empowers applications.
 This application is a tool to generate the needed files
 to quickly create a Cobra application.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		volumes, _ := cmd.Flags().GetStringSlice("volume")
-		nodes, _ := cmd.Flags().GetStringSlice("node")
-		requestIds, _ := cmd.Flags().GetStringSlice("request-id")
-		afterTime, _ := cmd.Flags().GetString("after-time")
-		isWatch, _ := cmd.Flags().GetBool("volume")
-		isPrevious, _ := cmd.Flags().GetBool("volume")
-		fmt.Println("controller called")
+		volumes, nodes, requestIds, afterTime, isWatch, isPrevious := GetFlags(cmd)
+
+		GetLogsByAzDriverPod("csi-azuredisk2-controller-6cb54cc5b5-6h6bt", "azuredisk", volumes, nodes, requestIds, "2019-08-22T12:00:00Z", true, false) // TODO: find constant
 	},
 }
 
 func init() {
 	getCmd.AddCommand(controllerCmd)
-
-	// Here you will define your flags and configuration settings.
-
-	// Cobra supports Persistent Flags which will work for this command
-	// and all subcommands, e.g.:
-	// controllerCmd.PersistentFlags().String("foo", "", "A help for foo")
-
-	// Cobra supports local flags which will only run when this command
-	// is called directly, e.g.:
-	// controllerCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
 }
 
 func getControllerPodName() string {
 	return ""
 }
 
-func GetLogsByConttroler(podName string, volumes []string, nodes []string, ) {
-
-}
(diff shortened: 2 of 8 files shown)


cmd/az-log/cmd/node.go
```diff
@@ -0,0 +1,49 @@
...
+This application is a tool to generate the needed files
+to quickly create a Cobra application.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			fmt.Println("Node is a required argument for the node command")
+		} else {
+			podName := args[0];
+			volumes, nodes, requestIds, afterTime, isWatch, isPrevious := GetFlags(cmd)
+
+			GetLogsByAzDriverPod("azuredisk", podName, volumes, nodes, requestIds, afterTime, isWatch, isPrevious)
```
nearora-msft, line 42: Shouldn't "azuredisk" bet the 2nd parameter ad podName the 1st?

nearora-msft: We should also make this container name "azuredisk" a constant

hccheng72: Thanks. Fixed the argument order and created a constant for "azuredisk".

cmd/az-log/cmd/node.go
```diff
@@ -0,0 +1,53 @@
...
+
+Cobra is a CLI library for Go that empowers applications.
+This application is a tool to generate the needed files
+to quickly create a Cobra application.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			fmt.Println("node is a required argument for \"node\" command")
+			os.Exit(0)
+		}
+		podName := args[0]
```
nearora-msft, line 41: Aren't we getting the node name in this argument?

cmd/az-log/cmd/controller.go
```diff
@@ -0,0 +1,76 @@
...
+Cobra is a CLI library for Go that empowers applications.
+This application is a tool to generate the needed files
+to quickly create a Cobra application.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		volumes, nodes, requestIds, since, sinceTime, isFollow, isPrevious := GetFlags(cmd)
+		config := getConfig()
+		clientsetK8s := getKubernetesClientset(config)
+
+		pod := getControllerPodName(clientsetK8s)
+		GetLogsByAzDriverPod(clientsetK8s, pod, AzureDiskContainer, volumes, nodes, requestIds, since, sinceTime, isFollow, isPrevious) // 17:57:01.170816 "2022-05-10T22:07:40Z"
```
nearora-msft, line 46: nit: we should remove this timestamp comment.

cmd/az-log/cmd/common.go
```diff
@@ -0,0 +1,241 @@
...
+
+func getKubernetesClientset(config *rest.Config) *kubernetes.Clientset {
+	clientsetK8s, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
+	}
+	return clientsetK8s
+}
+
+func GetLogsByAzDriverPod(clientsetK8s kubernetes.Interface, podName string, container string, volumes []string,
```
nearora-msft, line 85: Should we move this to pod.go instead to keep it consistent? Since we have the GetLogsByFile under the file.go?

cmd/az-log/cmd/node.go
hccheng72: My idea is the controller and node are K8s pod objects actually, so I called still it podName. Correct me if I'm wrong.

cmd/az-log/cmd/common.go
```diff
@@ -0,0 +1,241 @@
...
+		t, err := TimestampFormatValidation(sinceTime)
+		if err != nil {
+			fmt.Println(err.Error())
+			os.Exit(0)
+		}
+		timestamp := metav1.NewTime(t)
+		v1PodLogOptions.SinceTime = &timestamp
+	}
+
+	podLogOptions := make([]v1.PodLogOptions, 0)
```
nearora-msft, line 112: Is there a need to have multiple PodLogOption objects in a slice? Can we just use one instead, with all the options set?

cmd/az-log/cmd/common.go
```diff
@@ -0,0 +1,241 @@
...
+	for buf.Scan() {
+		log := buf.Text()
+
+		if !isAfterTime {
+			isAfterTime = LogTimeFilter(log, sinceTime)
+		}
+
+		if isAfterTime {
+			isPrint := true
+			if len(volumes) > 0 {
```
nearora-msft, line 186: Instead of having multiple conditions to check whether the volumes/nodes/requestIds are present in a line, can we have a single check separated by or/and conditions?

cmd/az-log/cmd/node.go
nearora-msft: Yeah, but isn't the requirement for the user to give the nodeName? The podname for azuredisk node pods is different form the name of the node.

cmd/az-log/cmd/common.go
hccheng72: When flag --previous is true, one PodLogOptions object is set field Previous = true to print logs from previous container and the other one is set Previous = false to reach logs from current container. That's why two objects with different field values are used.

cmd/az-log/cmd/common.go
```diff
@@ -0,0 +1,241 @@
...
+	if since != "" && sinceTime != "" {
+		fmt.Println("error: only one of --since/--since-time may be specified")
+		os.Exit(0)
+	}
+
+	return volumes, nodes, requestIds, since, sinceTime, isFollow, isPrevious
+}
+
+func getConfig() *rest.Config {
+	var kubeconfig string
```
nearora-msft, line 64: We should get the value of kubeconfig from the config file, like we did for az-analyze.

cmd/az-log/cmd/common.go
```diff
@@ -0,0 +1,241 @@
...
+		if !isAfterTime {
+			isAfterTime = LogTimeFilter(log, sinceTime)
+		}
+
+		if isAfterTime {
+			isPrint := true
+			if len(volumes) > 0 {
+				isPrint = false
+				for _, v := range volumes {
+					isPrint = strings.Contains(log, v)
```
nearora-msft, line 189: can be shortened to:
 if strings.Contains(log, v){
break
}

cmd/az-log/cmd/common.go
```diff
@@ -0,0 +1,241 @@
...
+					if isPrint {
+						break
+					}
+				}
+			}
+
+			if !isPrint {
+				fmt.Println("No logs are queried")
```
nearora-msft, line 197: The error message could be more descriptive, maybe something like: "No logs found for volumes: <insert-volume-name> "

cmd/az-log/cmd/common.go
nearora-msft: For that we could use different functions for each filter, like containsVolumes(volumes..) && containsNodes(nodes..) && containsRequestIds(requestIds..)

cmd/az-log/cmd/common.go
```diff
@@ -0,0 +1,241 @@
...
+					isPrint = strings.Contains(log, n)
+					if isPrint {
+						break
+					}
+				}
+			}
+
+			if !isPrint {
+				fmt.Println("No logs are queried")
+				os.Exit(0)
```
nearora-msft, line 213: Will this exit the process? Should we use continue instead, as we would still want to continue with the rest of the logs even though one of the log lines doesn't match.

cmd/az-log/cmd/common.go
hccheng72: yeah, will add this feature too

cmd/az-log/cmd/common.go
hccheng72: Does it make sense to just use one function like ContainsAny for volumes/nodes/requestids since they are all checking if a log contains a specific substring?

cmd/az-log/cmd/common.go
hccheng72: I found printing this message is an error (as well as os.Exit(0)). It should just mean that log isn't not queried rather than all logs. I fixed the code.

cmd/az-log/cmd/controller.go
```diff
@@ -0,0 +1,83 @@
...
+			}
+		}
+	},
+}
+
+func init() {
+	getCmd.AddCommand(controllerCmd)
+}
+
+func GetAzuredisk2Controller(clientsetK8s kubernetes.Interface) *v1.Pod {
```
nearora-msft, line 61: np: This can be renamed to GetLeaderControllerPod as that is more descriptive of what this function does.

cmd/az-log/cmd/controller.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		panic(err.Error())
+	}
+	holder := *lease.Spec.HolderIdentity
+	node := strings.Split(holder, "_")[0]
+
+	// Get pod name of the leader controller
+	pods, err := clientsetK8s.CoreV1().Pods(consts.ReleaseNamespace).List(context.Background(), metav1.ListOptions{
+		FieldSelector: "spec.nodeName=" + node,
+		LabelSelector: "app=csi-azuredisk2-controller",
```
nearora-msft, line 73: If we hardcode this, it will only run for the side by side deployment of azuredisk. We can add a parameter in the config file, where the user can mention whether they are running side by side deployment or not. Based on that we can add the label selector here.a

cmd/az-log/cmd/controller.go
```diff
@@ -0,0 +1,83 @@
...
+	Long:  `csi-azuredisk2-controller`,
+	Run: func(cmd *cobra.Command, args []string) {
+		volumes, nodes, requestIds, since, sinceTime, isFollow, isPrevious := GetFlags(cmd)
+		config := getConfig()
+		clientsetK8s := getKubernetesClientset(config)
+
+		pod := GetAzuredisk2Controller(clientsetK8s)
+		for {
+			currPodName := pod.Name
+			GetLogsByAzDriverPod(clientsetK8s, currPodName, AzureDiskContainer, volumes, nodes, requestIds, since, sinceTime, isFollow, isPrevious)
```
nearora-msft, line 45: Can we move the 1st call to GetLogsByAzDriverPod outside the for loop and run the for loop only if isFollow is true? It doesn't make sense to me that the user should have to wait for 20 seconds even if the follow mode is off.

cmd/az-log/cmd/file.go
```diff
@@ -0,0 +1,87 @@
...
+	Use:   "file",
+	Short: "A pod's logs file",
+	Long:  `A pod's logs file`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			fmt.Println("file name is a required argument for \"file\" command")
+			os.Exit(0)
+		}
+		filePath := args[0]
+		volumes, nodes, requestIds, since, sinceTime, _, _ := GetFlags(cmd)
```
nearora-msft, line 40: Can we show an error saying that this option isn't supported if the user tries to use follow or previous mode with the file logs.

cmd/az-log/cmd/node.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+		}
+	},
+}
+
+func init() {
+	getCmd.AddCommand(nodeCmd)
+}
+
+func GetAzuredisk2Node(clientsetK8s kubernetes.Interface, node string) *v1.Pod {
```
nearora-msft, line 68: np: Similarly this can be renamed to GetAzurediskPodFromNode.

cmd/az-log/cmd/node.go
```diff
@@ -0,0 +1,82 @@
...
+	// Get pod name of the azuredisk2-node running on given node
+	pods, err := clientsetK8s.CoreV1().Pods(consts.ReleaseNamespace).List(context.Background(), metav1.ListOptions{
+		FieldSelector: "spec.nodeName=" + node,
+		LabelSelector: "app=csi-azuredisk2-node",
```
nearora-msft, line 72: Similarly, we should use the label here based on whether we are running side by side deployment or not.

cmd/az-log/cmd/node.go
```diff
@@ -0,0 +1,82 @@
...
+	},
+}
+
+func init() {
+	getCmd.AddCommand(nodeCmd)
+}
+
+func GetAzuredisk2Node(clientsetK8s kubernetes.Interface, node string) *v1.Pod {
+	// Get pod name of the azuredisk2-node running on given node
+	pods, err := clientsetK8s.CoreV1().Pods(consts.ReleaseNamespace).List(context.Background(), metav1.ListOptions{
```
nearora-msft, line 70: This release namespace can also be other than "kube-system", so instead of hardcoding it, we should provide it as a parameter in the config file. If none is given, we can default to kube-system.

cmd/az-log/cmd/controller.go
hccheng72: Yeah, good point!

cmd/az-log/cmd/common.go
```diff
@@ -0,0 +1,174 @@
...
+
+func GetReleaseNamespace() string {
+	releaseNamespace := viper.GetString("releaseNamespace")
+	if releaseNamespace == "" {
+		releaseNamespace = consts.ReleaseNamespace
+	}
+	return releaseNamespace
+}
+
+func IsV2InstalledSideBySide() string {
```
nearora-msft, line 168: Since it doesn't return a boolean value, it would make sense to rename this function to GetDriverInstallationType or something like that.

cmd/az-log/cmd/node.go
```diff
@@ -0,0 +1,85 @@
...
+	},
+}
+
+func init() {
+	getCmd.AddCommand(nodeCmd)
+}
+
+func GetAzurediskPodFromNode(clientsetK8s kubernetes.Interface, node string) *v1.Pod {
+	// Get pod name of the azuredisk2-node running on given node
+	pods, err := clientsetK8s.CoreV1().Pods(consts.ReleaseNamespace).List(context.Background(), metav1.ListOptions{
```
nearora-msft, line 73: Did you forget to use the GetReleaseNamespace function here?

cmd/az-log/config/az-log.yaml
```diff
@@ -0,0 +1,4 @@
+kubeconfig: "" # default is "$HOME/.kube/config"
+driverNamespace: "" # default is "azure-disk-csi"
```
nearora-msft, line 2: Are we using the driverNamespace anywhere in the code?

hccheng72: Yes, it is used for searching for lease of controller plugin.

cmd/az-log/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Source Options
+|Command|Description|
+|---|---|
+|az-log get controller |Fetch and output logs from the leader controller plugin.|
+|az-log get node \<node-name\> |Fetch and output logs from the node plugin on the given node.|
+|az-log get file \<file-name\> |Fetch and output logs from a *.log file.|
+|az-log get pod \<pod-name\>/\<container-name\> |Fetch and output logs from a specific container and plugin running for the driver.|
+
+### Retrieval Options
```
nearora-msft, line 28: We can mention that these options will not work with the file command.

Makefile
```diff
@@ -373,3 +373,17 @@ pod-failover-test-containers:
 .PHONY: upgrade-test
 upgrade-test:
 	go test -v -timeout=0 ${GOTAGS} ./test/upgrade
+
+.PHONY: install-az-analyze
+install-az-analyze:
+	go install ./cmd/az-analyze
+	mkdir -p /var/tmp/azuredisk-csi-driver/cmd/az-analyze/config
+	rm -f /var/tmp/azuredisk-csi-driver/cmd/az-analyze/config/az-analyze.yaml
+	ln -s $(PWD)/cmd/az-analyze/config/az-analyze.yaml /var/tmp/azuredisk-csi-driver/cmd/az-analyze/config/az-analyze.yaml
```
nearora-msft, line 382: Will go install always install the tool under the var/tmp/.. folder? Can we give any other path, possibly that uses $GOBIN? So that the driver gets installed at $GOBIN and we can also create the link to the config file at that location.

cmd/az-log/cmd/root.go
```diff
@@ -0,0 +1,82 @@
...
+
+// initConfig reads in config file and ENV variables if set.
+func initConfig() {
+	if cfgFile != "" {
+		// Use config file from the flag.
+		viper.SetConfigFile(cfgFile)
+	} else {
+		// Search config in home directory with name "az-log" (without extension).
+		viper.SetConfigType("yaml")
+		viper.AddConfigPath("/var/tmp/azuredisk-csi-driver/cmd/az-log/config")
```
nearora-msft, line 71: Do we need to hardcode this path? I believe we should still keep the relative path as what we had before.

Makefile
hccheng72: The tool will be installed at $GOPATH/bin. Only the link to the config file is under var/tmp/.. folder.

cmd/az-log/cmd/root.go
hccheng72: If a relative path is used, the config file can't be found when running the tool in different directories.

cmd/az-log/README.md
hccheng72: Yeah, thanks for the reminder.

cmd/az-log/cmd/root.go
edreed: I believe we should set the config name to `az-analyze.config` and provide multiple search paths in this order: `/etc/az-analyze`, `$HOME/.config/az-analyze` and `.`. This would provide for global, per-user and current working directory based configurations and is pretty typical of apps for config management.

Makefile
nearora-msft: When we make changes in accordance to what Ed suggested below to support multiple search paths based on global, per-user and current working directory, we wouldn't need to copy this file when we run "make install-az-log". We could just mention in the README files that these config paths are supported.

hccheng72: Then since we don't need to copy the file, should I remove the make command "make install-az-log"? Users can install by just running "go install".

nearora-msft: I think we should still keep the make command, to keep it consistent with how we build and install everything else in the driver.

edreed:

PR #1330, opened 2022-05-13
[V2] bug: azdiskscheduler volume access mode check for prioritize instead of filter

sunpa93:
**What type of PR is this?**
/kind bug

* AzDiskScheduler was filtering results based on volume access mode checks and this was leading to early exits.
* It should not filter based on volume access mode checks but instead prioritize.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azdiskschedulerextender/azdiskschedulerextender.go
+++ b/pkg/azdiskschedulerextender/azdiskschedulerextender.go
@@ -168,7 +168,7 @@ func filter(context context.Context, schedulerExtenderArgs schedulerapi.Extender
 		}
 	}
 
-	diskRequestedByPod, diskRequestedByPodCount := getDisksRequestedByPod(ns, requestedVolumes, v1.ReadOnlyMany, v1.ReadWriteMany)
+	diskRequestedByPod, diskRequestedByPodCount := getDisksRequestedByPodExcludeAccessModes(ns, requestedVolumes)
 	if diskRequestedByPodCount == 0 {
 		return formatFilterResult(schedulerExtenderArgs.Nodes.Items, schedulerExtenderArgs.NodeNames, failedNodes, ""), nil
 	}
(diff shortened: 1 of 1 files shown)


pkg/azdiskschedulerextender/azdiskschedulerextender.go
```diff
@@ -243,7 +243,7 @@ func prioritize(context context.Context, schedulerExtenderArgs schedulerapi.Exte
 		go getAzVolumeAttachments(context, volumesChan)
 
 		// create a lookup map of all the volumes the pod needs
-		volumesPodNeeds, _ := getDisksRequestedByPod(ns, requestedVolumes)
+		volumesPodNeeds, _ := getDisksRequestedByPod(ns, requestedVolumes, v1.ReadOnlyMany, v1.ReadWriteMany)
```
landreasyan, line 246: Sorry for being neat picky here, I know I only asked for this change, but looking at it again I think this is still a really easy mistake to make in the future. To keep the change in this PR minimal, maybe we should rename the function to `getDisksForPodExcludeAccessMode`

pkg/azdiskschedulerextender/azdiskschedulerextender.go
```diff
@@ -437,7 +437,7 @@ func setNodeScoresToZero(nodes []v1.Node) (priorityList schedulerapi.HostPriorit
 	return
 }
 
-func getDisksRequestedByPod(ns string, requestedVolumes []v1.Volume, filterOutAccessModes ...v1.PersistentVolumeAccessMode) (disksRequestedByPod map[string]struct{}, requestedDiskCount int) {
+func getDisksRequestedByPodExcludeAccessModes(ns string, requestedVolumes []v1.Volume, filterOutAccessModes ...v1.PersistentVolumeAccessMode) (disksRequestedByPod map[string]struct{}, requestedDiskCount int) {
```
landreasyan, line 440: References? 😁

edreed:
Test failures will be resolved with #1340

PR #1344, opened 2022-05-18
chore: resize volume test should use tester's storageclass parameters

sunpa93:
**What type of PR is this?**
/kind chore

* The storageclass parameters set in dynamic provisioning test cases are not being properly propagated to resize volume tester.
* That said, the test was always using parameters returned `driver.GetParameters()`, which causes the tester to use `StandardSSD_LRS` sku all the time.
* And dynamic resizing doesn't seem to be supported for `STANDARD_LRS`, hence the dynamic provisioning test case `should create a block volume on demand and dynamically resize it without detaching` were amended to use `StandardSSD_LRS` instead of `Standard_LRS`

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/test/e2e/dynamic_provisioning_test.go
+++ b/test/e2e/dynamic_provisioning_test.go
@@ -827,7 +827,7 @@ func (t *dynamicProvisioningTestSuite) defineTests(isMultiZone bool) {
 			Volume:                 volume,
 			Pod:                    pod,
 			ResizeOffline:          false,
-			StorageClassParameters: map[string]string{"skuName": "Standard_LRS"},
+			StorageClassParameters: map[string]string{"skuName": "StandardSSD_LRS"},
 		}
 		test.Run(cs, ns)
 	})
@@ -868,7 +868,7 @@ func (t *dynamicProvisioningTestSuite) defineTests(isMultiZone bool) {
 			Volume:                 volume,
 			Pod:                    pod,
 			ResizeOffline:          false,
-			StorageClassParameters: map[string]string{"skuName": "Standard_LRS"},
+			StorageClassParameters: map[string]string{"skuName": "StandardSSD_LRS"},
 		}
 		test.Run(cs, ns)
 	})
--- a/test/e2e/testsuites/dynamically_provisioned_resize_volume_tester.go
+++ b/test/e2e/testsuites/dynamically_provisioned_resize_volume_tester.go
@@ -52,7 +52,7 @@ type DynamicallyProvisionedResizeVolumeTest struct {
 }
 
 func (t *DynamicallyProvisionedResizeVolumeTest) Run(client clientset.Interface, namespace *v1.Namespace) {
-	tStatefulSet, cleanup := t.Pod.SetupStatefulset(client, namespace, t.CSIDriver, driver.GetParameters())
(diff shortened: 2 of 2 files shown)


andyzhangx:
thanks for the PR, is there any doc about `dynamic resizing doesn't seem to be supported for STANDARD_LRS`? it's not mentioned here: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/expand-disks#expand-an-azure-managed-disk

sunpa93:
@andyzhangx It is not explicitly stated in the document so I guess it was an overstatement to say that dynamic resizing seems to be not supported with `Standard_LRS` but the dynamic resizing behavior for `Standard_LRS` sku seems to be very flaky. The `LiveResizeStorageClient` errors in most circumstances with the message that the disk cannot be resized while it is attached to a running VM. Tested with both v1 and v2 driver, dynamic resizing consistently fails with this LiveResizeStorageClient error in most cases.

andyzhangx:
> @andyzhangx It is not explicitly stated in the document so I guess it was an overstatement to say that dynamic resizing seems to be not supported with `Standard_LRS` but the dynamic resizing behavior for `Standard_LRS` sku seems to be very flaky. The `LiveResizeStorageClient` errors in most circumstances with the message that the disk cannot be resized while it is attached to a running VM. Tested with both v1 and v2 driver, dynamic resizing consistently fails with this LiveResizeStorageClient error in most cases.

@sunpa93 thanks, I tried on standard_lrs disk, actually it works, not sure what's going wrong here, also the resize e2e tests are still failing with this PR.

sunpa93:
@andyzhangx Thanks for the followup. Yes, dynamic resizing had also worked for me with `Standard_LRS` but more often it had failed. 

And I see that volume resizing test case is failing but it is strange that the test case that now fails is not dynamic volume resizing test case but offline volume resizing test case where a volume is detached from the node before it is resized. And the only difference for this test case compared to the previous would be that the sku names now get properly propagated to the tester.

So prior to this change, the offline volume resizing test case would have also run on `StandardSSD_LRS` because it would have used the default sku
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/8b97465f21c3c379ca3e22dc109a691243fe0dea/test/e2e/testsuites/dynamically_provisioned_resize_volume_tester.go#L55
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/8b97465f21c3c379ca3e22dc109a691243fe0dea/test/e2e/driver/azuredisk_driver.go#L130-L134

But now, it runs on `Standard_LRS` as defined in the test case (unless the test runs on one of ZRS supported regions)
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/33bf8eaa70a7464e07e29b49f738cdfa384985db/test/e2e/dynamic_provisioning_test.go#L784-L791
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/33bf8eaa70a7464e07e29b49f738cdfa384985db/test/e2e/testsuites/dynamically_provisioned_resize_volume_tester.go#L55

And the offline volume resizing has been consistently passing for v2 test, which makes it more confusing.

PR #1360, opened 2022-06-01
helm: use correct nodeSelector for k8s 1.25+

jackfrancis:
**What type of PR is this?**

/kind failing-test
/kind api-change

This PR sets the correct nodeSelector in the helm chart so that it works with k8s 1.25+

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
+++ b/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
@@ -26,7 +26,11 @@ spec:
       nodeSelector:
         kubernetes.io/os: linux
         {{- if .Values.controller.runOnMaster}}
+            {{- if ge .Capabilities.KubeVersion.Minor "25" }}
+        node-role.kubernetes.io/control-plane: ""
+            {{- else }}
         node-role.kubernetes.io/master: ""
+            {{- end }}
         {{- end}}
       priorityClassName: system-cluster-critical
 {{- with .Values.controller.tolerations }}


charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
```diff
@@ -26,7 +26,11 @@ spec:
       nodeSelector:
         kubernetes.io/os: linux
         {{- if .Values.controller.runOnMaster}}
```
marosset, line 28: Should we add a new `runOnControlPlane` setting and have this line check if either is set and the apply the correct node-selector?

That is more work now but probably better in the long run.

marosset: fair enough

andyzhangx: I think we could remove `controller.runOnMaster` chart, replace with `controller.runOnControlPlane` directly

andyzhangx:
this PR should have fixed the test failure on 1.25+: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1362

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
jackfrancis: Won't this break back-compat for existing users? (or has a new version of the chart w/ this configuration never been released?)

andyzhangx: that's ok for chart config of new versions, add a new `controller.runOnControlPlane` could be safer, and after a few versions, we could remove `controller.runOnMaste`

jackfrancis: how about we do that as a follow-up change? in the meantime looks like we can merge this to fix broken tests

andyzhangx: the windows test failure should already be fixed by https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1362, no worry, I will provide a fix using `controller.runOnControlPlane` later

andyzhangx: PR(https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1366) would address the issue.

PR #1361, opened 2022-06-01
[V2] feat: update volume attachment with publish context upon attachment completion

sunpa93:
**What type of PR is this?**
/kind bug

* There was an issue found when downgrading v2 driver to v1 driver with missing lun value in `VolumeAttachment` object.
* This PR addresses this issue by updating the `VolumeAttachment` object with publish context when attachment completes.
* 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azureconstants/azure_constants.go
```diff
@@ -58,11 +58,15 @@ const (
...
 	PvcNameKey                    = "csi.storage.k8s.io/pvc/name"
+	PvcNameLabel                  = "csi.storage.k8s.io/pvc"
 	PvcNamespaceKey               = "csi.storage.k8s.io/pvc/namespace"
 	PvcNamespaceTag               = "kubernetes.io-created-for-pvc-namespace"
+	PvcNamespaceLabel             = "csi.storage.k8s.io/pvc-namespace"
 	PvcNameTag                    = "kubernetes.io-created-for-pvc-name"
 	PvNameKey                     = "csi.storage.k8s.io/pv/name"
 	PvNameTag                     = "kubernetes.io-created-for-pv-name"
+	PvNameLabel                   = "csi.storage.k8s.io/pv"
+	VolumeAttachmentKey           = "csi.storage.k8s.io/volumeattachment/name"
```
edreed, line 69: Is this defined by CSI? If not, we should use our own annotation namespace, e.g. `disk.csi.azure.com/volumeattachment/name`.

pkg/controller/attach_detach.go
```diff
@@ -251,6 +253,19 @@ func (r *ReconcileAttachDetach) triggerAttach(ctx context.Context, azVolumeAttac
 				_, uerr := updateState(azv, diskv1beta1.Attached, forceUpdate)
 				return uerr
 			}
+
+			if azVolumeAttachment.Spec.RequestedRole == diskv1beta1.PrimaryRole {
+				vaName, err := r.controllerSharedState.waitForVolumeAttachmentName(cloudCtx, azVolumeAttachment)
+				if err == nil {
+					vaUpdateFunc := func(obj interface{}) error {
+						va := obj.(*storagev1.VolumeAttachment)
+						va.Status.AttachmentMetadata = response
```
edreed, line 262: It would be safer to simply add/replace an existing LUN entry rather than replace the entire map. Event though it's unlikely, there's no guarantee that I am aware of to prevent CSI from storing other values there in addition to that returned by ControllerPublishVolume.

sunpa93:
Will revert `go.mod` to `github.com/edreed/cloud-provider-azure` once https://github.com/edreed/cloud-provider-azure/pull/97 is merged. Temporarily updated the file with forked repo to test the changes while the PR is pending.

pkg/azureutils/azure_disk_utils.go
```diff
@@ -127,6 +127,23 @@ const (
 	UpdateAll       CRIUpdateMode = 0b11
 )
 
+type CloudAttachResult struct {
+	publishContext map[string]string
+	err            error
+}
+
+func SetCloudAttachResult(publishContext map[string]string, err error) CloudAttachResult {
```
edreed, line 135: `NewCloudAttachResult`

pkg/provisioner/cloudprovisioner.go
```diff
@@ -329,36 +329,39 @@ func (c *CloudProvisioner) ListVolumes(
 	return c.listVolumesInNodeResourceGroup(ctx, start, int(maxEntries))
 }
 
+// PublishVolume calls AttachDisk asynchronously and returns early lun assignment value and a channel for the async attach results.
 func (c *CloudProvisioner) PublishVolume(
 	ctx context.Context,
 	volumeID string,
 	nodeID string,
-	volumeContext map[string]string) (map[string]string, error) {
+	volumeContext map[string]string) (chan azureutils.CloudAttachResult, map[string]string, error) {
```
edreed, line 337: Rather than return the `PublishVolume` result either through the channel or by direct return (forcing the caller to handle both cases), it would be cleaner and simpler to *always* return results through the channel.

pkg/provisioner/cloudprovisioner.go
```diff
@@ -329,36 +329,39 @@ func (c *CloudProvisioner) ListVolumes(
...
 	var err error
+	var attachResultCh chan azureutils.CloudAttachResult
```
edreed, line 339: `CloudAttachResult` should just be part of the provisioner package.

pkg/provisioner/cloudprovisioner.go
```diff
@@ -377,20 +380,31 @@ func (c *CloudProvisioner) PublishVolume(
...
+		go func() {
+			ctx, w := workflow.New(ctx)
+			defer func() { w.Finish(err) }()
+
+			resultLun, resultErr := c.cloud.AttachDisk(ctx, asyncAttach, diskName, volumeID, nodeName, cachingMode, disk)
+			if resultErr != nil {
+				w.Logger().Errorf(err, "attach volume %q to instance %q failed", volumeID, nodeName)
+			}
+			w.Logger().V(2).Infof("attach operation successful: volume %q attached to node %q.", volumeID, nodeName)
+			attachResultCh <- azureutils.SetCloudAttachResult(map[string]string{azureconstants.LUN: strconv.Itoa(int(resultLun))}, resultErr)
```
edreed, line 399: Do we really need to return the LUN as part of the `CloudAttachResult`? It should be the same as the one we already received from the `lunCh` and returned to the caller.

sunpa93: Yes, that sounds right. I will address this along with an issue I spotted with blocking lun channel.

pkg/provisioner/cloudprovisioner.go
sunpa93: Shouldn't we keep the type definition in a separate file for interfaces?

edreed: The result is specific to the `CloudProvisioner`. I don't believe we have a use case for it otherwise and it's OK to have type dependencies on the `provisioner` package event when interfaces are defined in another package.

sunpa93:
/retest
`pull-azuredisk-csi-driver-e2e-capz-mainv2` failing due to `ConflictingUserInput` error, known to occur when attempting to make requests to attach a disk to multiple VMs simultaneously. Will be addressed in future PRs.

PR #1365, opened 2022-06-02
[V2] fix az-analyze tool and add docs

hccheng72:
**What type of PR is this?**
/kind bug
/kind documentation

Fix an error on format of object's age, edit tool description, and add a README file.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


cmd/az-analyze/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Features
+|Command|Description|
+|---|---|
+|az-analyze get azv|Show all azVolumes with a pod column.
+|--pod \<pod-name\>|Show all azVolumes used by a given pod|
+|--namespace \<pod-namespace\>|Specify the namespace of the pod. If it's not specified, default namesapce is used.|
+
+|Command|Description|
+|---|---|
+|az-analyze get azva|Show all azVolumesAttachments with pod, node, and zone as columns|
```
nearora-msft, line 26: It would be more clear if we separate out the command and it' parameters in different sections. It is a bit confusing to have just the parameters under the command column in the table.

cmd/az-analyze/cmd/root.go
```diff
@@ -69,7 +67,7 @@ func initConfig() {
 	} else {
 		// Search config in home directory with name ".az-analyze" (without extension).
 		viper.SetConfigType("yaml")
-		viper.AddConfigPath("./config")
+		viper.AddConfigPath("/var/tmp/azuredisk-csi-driver/cmd/az-analyze/config")
```
nearora-msft, line 70: We should use the relative path, like we were using before.

nearora-msft:
It'd be good to add the Makefile changes for az-analyze in this PR instead of the other one, so that they can be merged independent of each other.

cmd/az-analyze/cmd/root.go
edreed: I believe we should set the config name to `az-analyze.config` and provide multiple search paths in this order: `/etc/az-analyze`, `$HOME/.config/az-analyze` and `.`. This would provide for global, per-user and current working directory based configurations and is pretty typical of apps for config management.

edreed:
Test failure is unrelated to this change.

PR #1385, opened 2022-06-17
[V2] fix: ensure AzDriverNode recreated if deleted

### LINKED ISSUE #1356: [v2] v2 node agent doesn't re-create the azdrivernodes CRI in the heartbeat report path
**What happened**:
If the azdrivernodes CRI for a node gets deleted accidently, the heartbeat report for the node will stop, effectively taking it out of rotation from placement by extender.

CSI-Node-Agent should re-create the azdrivernodes CRI in the heart beat report path.

**How to reproduce it**:

1. Deploy the v2 driver.
2. Delete one of the azdrivernodes CRIs
kubectl delete azdrivernodes.disk.csi.azure.com aks-nodepool1-51977605-vmss000000 -n kube-system
3. Monitor to see if the CRI gets re-created
kubectl get -A azdrivernodes.disk.csi.azure.com

**Environment**:
- CSI Driver version: v2.0.0-beta.3

edreed:
**What type of PR is this?**

/kind bug

Ensures the `AzDriverNode` object for a node is recreated if deleted.

Fixes #1356 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azuredisk_v2.go
+++ b/pkg/azuredisk/azuredisk_v2.go
@@ -23,12 +23,14 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"math/rand"
 	"os"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/go-logr/logr"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -36,6 +38,7 @@ import (
 	apiRuntime "k8s.io/apimachinery/pkg/runtime"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/klog/v2"
 	"k8s.io/klog/v2/klogr"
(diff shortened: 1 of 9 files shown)


pkg/azuredisk/azuredisk_v2.go
```diff
@@ -434,91 +455,146 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
...
+	err := d.updateOrCreateAzDriverNode(innerCtx, azDriverNodeHealthy)
+
+	w.Finish(err)
+
+	return err
+}
+
+// updateOrCreateAzDriverNode updates the AzDriverNode status, creating a new object if one does not exist
+func (d *DriverV2) updateOrCreateAzDriverNode(ctx context.Context, status azDriverNodeStatus) error {
+	logger := logr.FromContextOrDiscard(ctx)
```
sunpa93, line 552: Is the intent here to use a plain logger for the heartbeat update loop and context logger for when registering `AzDriverNode`?

sunpa93 (approved):
/lgtm

pkg/azuredisk/azuredisk_v2.go
edreed: In both cases, this will the workflow logger. Long term, I'd like us to move to a model where we always get the logger from the context rather than assume klog.

sunpa93: It seems like the context fed in this loop doesn't contain worklow though. Am I missing something?
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/4e6c6da101faf1ec382552ef3ee57b427b925b3a/pkg/azuredisk/azuredisk_v2.go#L522-L524

sunpa93: I agree that we should move to getting logger from context long term.

edreed: `updateAzDriverNodeHeartbeat` creates a workflow.

sunpa93: It does but I don't think the context in `updateAzDriverNodeHeartbeat` is being passed onto `runAzDriverNodeHeartbeatLoop` though.

edreed: We sync'ed up offline - yes, the main heartbeat loop will use the plain logger for now. Longer term, we should add `logr`-based loggers to the context passed here and use it.

sunpa93:
The changes are from rebasing.
/lgtm

PR #1389, opened 2022-06-22
[V2] feat: Skipping prioritize/(filter) extender processing for non-v2 disks

### LINKED ISSUE #1097: [V2] data disk count is invalid in shared disk mode
**What happened**:
1) create a node pool with 2 nodes (DS2_v2 size), each node could attach 8 data disks at most
2) create 8 pods with one disk mount each (with `maxShares: 2`)
3) create the 9th pod with one disk mount, disk would be scheduled to one node, stuck in `ContainerCreating` state forever, and error would be like following:

```
Events:
  Type     Reason              Age               From                     Message
  ----     ------              ----              ----                     -------
  Warning  FailedScheduling    47s               default-scheduler        0/2 nodes are available: 2 pod has unbound immediate PersistentVolumeClaims.
  Warning  FailedScheduling    47s               default-scheduler        0/2 nodes are available: 2 pod has unbound immediate PersistentVolumeClaims.
  Normal   Scheduled           40s               default-scheduler        Successfully assigned default/statefulset-azuredisk-share-8 to aks-agentpool-70325192-vmss000001
  Warning  FailedAttachVolume  7s (x7 over 39s)  attachdetach-controller  AttachVolume.Attach failed for volume "pvc-54c55671-22d4-4d8c-ba69-834634e65573" : rpc error: code = FailedPrecondition desc = Retriable: false, RetryAfter: 0s, HTTPStatusCode: 409, RawError: {
  "error": {
    "code": "OperationNotAllowed",
    "message": "The maximum number of data disks allowed to be attached to a VM of this size is 8.",
    "target": "dataDisks"
  }
}
```

Root cause: in shared disk mode, current maximum data disk count calculation is invalid any more, need to adjust

It's ok that with with `maxShares: 2`, we could only attach 8 data disks, while the 9th pod should be in `Pending` state instead of  scheduled to one node which obviously could not attach any disk any more.

**How to reproduce it**:

**Environment**:
- CSI Driver version: V2
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

landreasyan:
…V2 driver disks.

**What type of PR is this?**
Added on optimization in the extender to skip processing on prioritize and filter calls for non-v2 driver volumes. Currently, only prioritize is skipped until we fix #1097 in the controller but the functionality is added for both calls. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azdiskschedulerextender/azdiskschedulerextender.go
+++ b/pkg/azdiskschedulerextender/azdiskschedulerextender.go
@@ -58,6 +58,7 @@ var (
 	kubeClientset              kubernetes.Interface
 	kubeExtensionClientset     azdisk.Interface
 	criNamespace               string
+	azDiskVolumes              sync.Map
 	pvcToPvMap                 sync.Map
 	pvToDiskNameMap            sync.Map
 )
@@ -79,6 +80,7 @@ func init() {
 func initSchedulerExtender(ctx context.Context) {
 	pvcToPvMap = sync.Map{}
 	pvToDiskNameMap = sync.Map{}
+	azDiskVolumes = sync.Map{}
 	var err error
 	kubeExtensionClientset, err = getKubernetesExtensionClientsets()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


andyzhangx:
@landreasyan is https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/1097 fixed or not?

landreasyan:
> @landreasyan is #1097 fixed or not?

@andyzhangx #1907 is fixed but we are changing the approach we took in the controller to prefer primary attachments over replicas which will eliminate the need of having this fix in the scheduler extender. Removing the fix for #1907 from the extender will make it completely non-blocking and in line with the expectations of the autoscaler.

PR #1401, opened 2022-06-26
[V2] fix: fix retry of failed attach

edreed:
**What type of PR is this?**

/kind bug

/kind regression

An attach operation is only started by the `attach_detach` controller if the `AzVolumeAttachment.Status.Detail` field is `nil`. This PR set this field to `nil` when we need to retry an attach operation. This PR also updates the workflow annotations when the retry is triggered via `CrdProvisioner.WatchForAttach` or `CrdProvisioner.waitForLun`.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


sunpa93 (approved):
Good catch!
/lgtm

PR #1406, opened 2022-07-02
[V2] feat: allow detaches of replica attachment for primary attachment

sunpa93:
request

**What type of PR is this?**
/kind feature

* This PR includes a temporary change to allow detach of replica attachment for primary attachment request.
* It targets to handle other relevant data race issues.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azureutils/azure_disk_utils.go
+++ b/pkg/azureutils/azure_disk_utils.go
@@ -84,6 +84,25 @@ const (
 	diskNameGenerateMaxLength = 76
 )
 
+var AttachmentRoles = map[AttachmentRoleMode]string{
+	PrimaryOnly: string(azdiskv1beta2.PrimaryRole),
+	ReplicaOnly: string(azdiskv1beta2.ReplicaRole),
+}
+
+type AttachmentRoleMode int
+
+const (
+	PrimaryOnly AttachmentRoleMode = iota
+	ReplicaOnly
+	AllRoles
+)
+
+type LabelPair struct {
+	Key      string
+	Operator selection.Operator
+	Entry    string
+}
+
 type ClientOperationMode int
 
 type ManagedDiskParameters struct {
@@ -186,7 +205,7 @@ func GetFStype(attributes map[string]string) string {
 	return ""
 }
 
-func GetNodeMaxDiskCount(labels map[string]string) (int, error) {
+func GetNodeMaxDiskCountWithLabels(labels map[string]string) (int, error) {
 	if labels == nil {
 		return 0, fmt.Errorf("labels for the node are not provided")
 	}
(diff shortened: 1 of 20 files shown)


pkg/provisioner/crdprovisioner.go
```diff
@@ -46,22 +53,143 @@ import (
...
+	}
+	kubeClient, err := kubeClientset.NewForConfig(kubeConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, consts.DefaultInformerResync)
+	startWatchingKube(context.Background(), kubeInformerFactory)
+
+	azInformerFactory := azdiskinformers.NewSharedInformerFactory(azdiskClient, consts.DefaultInformerResync)
```
landreasyan, line 182: We need to wait for cache sync for `azInformerFactory` as well.

pkg/provisioner/crdprovisioner.go
```diff
@@ -46,22 +53,143 @@ import (
...
 }
 
 const (
 	// TODO: Figure out good interval and timeout values, and make them configurable.
 	interval = time.Duration(1) * time.Second
 )
 
-func NewCrdProvisioner(azdiskClient azdisk.Interface, objNamespace string) (*CrdProvisioner, error) {
-	informerFactory := azdiskinformers.NewSharedInformerFactory(azdiskClient, consts.DefaultInformerResync)
+type CachedReader struct {
```
landreasyan, line 67: Should we move the `CacheReader` definition and methods to a separate file or to utils?

pkg/azureutils/azure_disk_utils.go
```diff
@@ -84,6 +84,25 @@ const (
...
 )
 
+var AttachmentRoles = map[AttachmentRoleMode]string{
+	PrimaryOnly: string(azdiskv1beta2.PrimaryRole),
+	ReplicaOnly: string(azdiskv1beta2.ReplicaRole),
+}
+
+type AttachmentRoleMode int
+
+const (
```
landreasyan, line 94: Would be good to combine this with the rest of the constants at the beginning of the file.

sunpa93: I believe defining these constant after its type definition makes it more readable.

pkg/provisioner/crdprovisioner.go
sunpa93: `CachedReader` usage and its definition is going to vary based on what objects the instance needs to watch; this definition at the moment implements `client.Reader` interface in a `crdProvisioner` specific way. So it would be more suitable, I think, to leave it in the crdProvisioner file for now.

PR #1408, opened 2022-07-06
[V2] [WIP] feat: add rate limiter latency metric to find throttling

chloeegong:
**What type of PR is this?**

- The `RateLimiterLatency` metric creates a new `HistogramVec` Prometheus metric to measure throttling in the driver
- Uses the metric library in client-go to enable and register the new metric

**Which issue(s) this PR fixes:**

Fixes #

**Requirements:**

- [ ]  uses [conventional commit messages](https://www.conventionalcommits.org/)

- [ ]  includes documentation

- [ ]  adds unit tests

- [ ]  tested upgrade from previous version

**Special notes for your reviewer:

Release note:**


edreed:
I assume this is a temporary change for review and test coverage?

landreasyan:
> I assume this is a temporary change for review and test coverage? In general, we must not modify the `vendor` directory directly since this is managed by `go mod` and this change will cause the `verify` tests to fail.

Yes, these changes will be moved to our code in the next commit.

PR #1410, opened 2022-07-08
[V2] feat: add rate limiter latency metric to find throttling

chloeegong:
**What type of PR is this?**

- The RateLimiterLatency metric uses a HistogramVec imported from the metrics package to measure throttling in the driver

- Uses the metric library in client-go to enable and register the new metric in the `azuredisk_v2.go` file 
Which issue(s) this PR fixes:

Fixes #

**Requirements:**

- [ ]  uses [conventional commit messages](https://www.conventionalcommits.org/)
- [ ]  includes documentation
- [ ]  adds unit tests
- [ ]  tested upgrade from previous version

**Special notes for your reviewer:

Release note:**

### DIFF
--- a/pkg/azuredisk/azuredisk_v2.go
+++ b/pkg/azuredisk/azuredisk_v2.go
@@ -24,6 +24,7 @@ import (
 	"flag"
 	"fmt"
 	"math/rand"
+	"net/url"
 	"os"
 	"strings"
 	"sync"
@@ -46,7 +47,10 @@ import (
 
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
+	clientmetrics "k8s.io/client-go/tools/metrics"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
 	azdiskv1beta2 "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/azuredisk/v1beta2"
 	azdisk "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/client/clientset/versioned"
 	azdiskinformers "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/client/informers/externalversions"
@@ -89,6 +93,27 @@ const (
 	azDriverNodeHealthy      azDriverNodeStatus = "Driver node healthy."
 )
 
+// LatencyAdapter implements LatencyMetric.
+type LatencyAdapter struct {
+	metric *metrics.HistogramVec
+}
+
+// Observe increments the request latency metric for the given verb/URL.
+func (l *LatencyAdapter) Observe(_ context.Context, verb string, u url.URL, latency time.Duration) {
+	if latency > consts.LongThrottleLatency {
+		l.metric.WithLabelValues(verb, u.String()).Observe(latency.Seconds())
+	}
+}
+
+// RateLimiterLatency reports the rate limiter latency in seconds per verb/URL.
+var rateLimiterLatency = metrics.NewHistogramVec(
+	&metrics.HistogramOpts{
+		Subsystem: consts.RestClientSubsystem,
+		Name:      consts.LatencyKey,
+		Help:      "Rate limiter latency in seconds. Broken down by verb and URL.",
+		Buckets:   []float64{1, 5, 10, 15, 30, 45, 60},
+	}, []string{"verb", "url"})
+
 // DriverV2 implements all interfaces of CSI drivers
 type DriverV2 struct {
 	DriverCore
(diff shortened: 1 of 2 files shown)


pkg/azuredisk/azuredisk_v2.go
```diff
@@ -112,6 +140,10 @@ type DriverV2 struct {
 	kubeClientQPS                     int
 }
 
+func init() {
+	clientmetrics.RateLimiterLatency = &LatencyAdapter{metric: rateLimiterLatency}
```
edreed, line 144: We should call the `[Register](https://pkg.go.dev/k8s.io/client-go@v0.24.2/tools/metrics#Register)` function to override the default metrics reporter. It can only be called once, it doesn't appear that anything does so currently. I expected this is the supported way for users to provide metrics loggers.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -89,6 +92,31 @@ const (
...
+	l.metric.WithLabelValues(verb, u.String()).Observe(latency.Seconds())
+}
+
+// RateLimiterLatency reports the rate limiter latency in seconds per verb/URL.
+var rateLimiterLatency = metrics.NewHistogramVec(
+	&metrics.HistogramOpts{
+		Subsystem: RestClientSubsystem,
+		Name:      LatencyKey,
+		Help:      "Rate limiter latency in seconds. Broken down by verb and URL.",
+		Buckets:   []float64{0.1, .25, .5, 1},
```
edreed, line 117: We've seen latencies greatly exceed 1s in the past. Perhaps add 5, 10, 15, 30, 45, 60?

pkg/azuredisk/azuredisk_v2.go
landreasyan: `Register` is already being called in controller-runtime in an [`init()`](https://github.com/kubernetes-sigs/controller-runtime/blob/ffd9ec8768b7d2e82aee6b0c76d481a5bdda4075/pkg/metrics/client_go_adapter.go#L139) function (this package is imported by the manager package) before we even get to our code to update any of the options, so calling it again here will have no effect. The only way we can change the configuration and enable the rate limiter in our code is by updating the package variables in client-go.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -89,6 +92,31 @@ const (
 	azDriverNodeHealthy      azDriverNodeStatus = "Driver node healthy."
 )
 
+// Metrics subsystem and all of the keys used by the rest client.
+const (
+	RestClientSubsystem = "rest_client"
+	LatencyKey          = "throttling_request_latency_seconds" // edits: prefix or postfix
```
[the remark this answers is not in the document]
landreasyan: Yes, I think we should.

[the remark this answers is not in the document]
landreasyan: Also, since we are using an abstraction over the rest client in our code, I am thinking that maybe we should rename `rest_client` to `kube_client`? And maybe `throttling_request_latency_seconds` to `rate_limiter_hrottling_request_latency_seconds`?

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -89,6 +92,31 @@ const (
 	azDriverNodeHealthy      azDriverNodeStatus = "Driver node healthy."
 )
 
+// Metrics subsystem and all of the keys used by the rest client.
+const (
+	RestClientSubsystem = "kube_client"
```
landreasyan, line 97: You have these constants defined in two files. We don't need these here if we are defining them in `azure_constants.go`. The naming foe these ones is probably better tho.

sunpa93 (approved):
/lgtm

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -89,6 +93,25 @@ const (
...
 )
 
+// LatencyAdapter implements LatencyMetric.
+type LatencyAdapter struct {
+	metric *metrics.HistogramVec
+}
+
+// Observe increments the request latency metric for the given verb/URL.
+func (l *LatencyAdapter) Observe(_ context.Context, verb string, u url.URL, latency time.Duration) {
+	l.metric.WithLabelValues(verb, u.String()).Observe(latency.Seconds())
```
sunpa93, line 103: We should check if `latency < longThrottleLatency`, where `longThrottleLatency = 50 * time.Millisecond`, and only observe the rate limiter latency that throttled.

sunpa93 (changes requested):
We should only observe throttled latency for our throttling test purpose.

pkg/azuredisk/azuredisk_v2.go
landreasyan: I think the check would be sign should be reversed here. We want to observe the requests that are above 50ms, so when `latency > longThrottleLatency`.

landreasyan: btw, I just checked the code, and all the other calls get throttled too, it is just under 50ms.

edreed:
The verify tests will fail until #1434 is merged.

PR #1445, opened 2022-08-02
fix: invalid ultra disk default iops issue

### LINKED ISSUE #1442: Requesting `iopsPerGB` or `allowAutoIOPSPerGBIncrease ` for ultradisk parameter.

if your storage class is like below (with ultradisk as the sku name), the default provisioned IOPS for such disk is 500
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
allowVolumeExpansion: true
metadata:
  name: ultra-disk
parameters:
  cachingMode: None
  skuname: UltraSSD_LRS
provisioner: disk.csi.azure.com
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer

```

when you provision a PVC with disksize greater than 500GB it results in the following error.

```json
"statusMessage": "{\"error\":{\"code\":\"BadRequest\",\"message\":\"The value 500 of parameter disk.diskIOPSReadWrite is not valid. The supported range for a UltraSSD_LRS disk of size 512 GB is between 512 and 153600.\"}}",
```

Our pods are in pending and this is affecting our customers from provisioning Ultradisk over 500GB of disk capacity. 

We need something like `iopsPerGB ` as explained [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/parameters.md) in aws-ebs's driver or a flag like this `allowAutoIOPSPerGBIncrease` that sets a meaningful default IOPS so that Volume mount wont fail. 

can this flag also be mutable to existing StorageClass so that we dont have to create a new storage to support this feature? 

**Additional context**

Note: Support request ID	2207310050000344 and affecting our production system right now.

andyzhangx:
**What type of PR is this?**
/kind bug

fix: invalid ultra disk default iops issue

Fixes #1442

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

this PR tries to fix the following issue:
```
  "error": {
    "code": "BadRequest",
    "message": "The value 500 of parameter disk.diskIOPSReadWrite is not valid. The supported range for a UltraSSD_LRS disk of size 512 GB is between 512 and 153600."
  }
```

**Release note**:
```
fix: invalid ultra disk default iops issue
```

### DIFF
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -405,3 +405,35 @@ func getNodeInfoFromLabels(ctx context.Context, nodeName string, kubeClient clie
 	}
 	return node.Labels[consts.WellKnownTopologyKey], node.Labels[consts.InstanceTypeKey], nil
 }
+
+// getDefaultDiskIOPSReadWrite according to requestGiB
+//
+//	ref: https://docs.microsoft.com/en-us/azure/virtual-machines/disks-types#ultra-disk-iops
+func getDefaultDiskIOPSReadWrite(requestGiB int) int {
+	iops := azurecloudconsts.DefaultDiskIOPSReadWrite
+	if requestGiB > iops {
+		iops = requestGiB
+	}
+	if iops > 160000 {
+		iops = 160000
+	}
+	return iops
+}
+
+// getDefaultDiskMBPSReadWrite according to requestGiB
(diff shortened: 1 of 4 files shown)


dhilipkumars (approved):

ZeroMagic (approved):
/lgtm

PR #1446, opened 2022-08-02
[V2] fix: Volume persistent-storage not csi error log

### LINKED ISSUE #1416: [V2] Fix "Volume persistent-storage not csi" misclassified error log line
**What happened**:
On pod creation reconciliation, we check the [claimToVolumeMap ](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/a9975e9ef8dfeab8c0656510bd8711d3d3a01c49/pkg/controller/common.go#L585) for AzDiskDriver volumes, but the pv is loaded to the map on its [update reconciliation](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/a9975e9ef8dfeab8c0656510bd8711d3d3a01c49/pkg/controller/pv.go#L205) when it is bound which happens after pod is added. As a result for all volumes of the pod we end up returning a confusing error of "Volume persistent-storage not csi". I believe the reconciliation on pod update later fixes this misclassification but we need to change the workflow/error to remove the confusing error log. 

Here is where the log is printed: 

**How to reproduce it**:

**Environment**:
- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

shlokshah-dev:
**What type of PR is this?**
/kind cleanup

[V2] Fix "Volume persistent-storage not csi" misclassified error log line

Fixes #1416

**Requirements**:

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/common.go
+++ b/pkg/controller/common.go
@@ -565,11 +565,11 @@ func (c *SharedState) addPod(ctx context.Context, pod *v1.Pod, updateOption upda
 		}
 		namespacedClaimName := getQualifiedName(pod.Namespace, volume.PersistentVolumeClaim.ClaimName)
 		if _, ok := c.claimToVolumeMap.Load(namespacedClaimName); !ok {
-			// Log message if the Pod status is not Running
+			// Log message if the Pod status is Running
 			if pod.Status.Phase == v1.PodRunning {
 				w.Logger().V(5).Infof("Skipping Pod %s. Volume %s not csi. Driver: %+v", pod.Name, volume.Name, volume.CSI)
-				continue
 			}
+			continue
 		}
 		w.Logger().V(5).Infof("Pod %s. Volume %v is csi.", pod.Name, volume)
 		claimSet.add(namespacedClaimName)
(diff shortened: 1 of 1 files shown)


landreasyan (approved):

pkg/controller/common.go
```diff
@@ -565,8 +565,11 @@ func (c *SharedState) addPod(ctx context.Context, pod *v1.Pod, updateOption upda
 		}
 		namespacedClaimName := getQualifiedName(pod.Namespace, volume.PersistentVolumeClaim.ClaimName)
 		if _, ok := c.claimToVolumeMap.Load(namespacedClaimName); !ok {
-			w.Logger().V(5).Infof("Skipping Pod %s. Volume %s not csi. Driver: %+v", pod.Name, volume.Name, volume.CSI)
-			continue
+			// Log message if the Pod status is not Running
```
sunpa93, line 568: Log message if the Pod status _is Running_?

pkg/controller/common.go
```diff
@@ -565,8 +565,11 @@ func (c *SharedState) addPod(ctx context.Context, pod *v1.Pod, updateOption upda
...
+			if pod.Status.Phase == v1.PodRunning {
+				w.Logger().V(5).Infof("Skipping Pod %s. Volume %s not csi. Driver: %+v", pod.Name, volume.Name, volume.CSI)
+				continue
```
sunpa93, line 571: We would need this `continue` to be outside of `if pod.Status.Phase == v1.PodRunning`.

pkg/controller/common.go
shlokshah-dev: Updated!

pkg/controller/common.go
shlokshah-dev: Updated

PR #1448, opened 2022-08-03
[V2] chore: add early return logic for scheduler filter / prioritize and update prioritize scoring logic

sunpa93:
**What type of PR is this?**
/kind feature

* This change allows for AzDiskSchedulerExtender to early return on filter when no azuredisk provisioned disk is requested by pod. This fix is dependent on #1406.
* This change updates node score logic to account for node remaining capacity.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azdiskschedulerextender/azdiskschedulerextender.go
```diff
@@ -150,9 +150,9 @@ func filter(context context.Context, schedulerExtenderArgs schedulerapi.Extender
 	ns := schedulerExtenderArgs.Pod.Namespace
 
 	// get requested disks created by Azure Disk V2 driver and return early if none are found
-	diskRequestedByPod, totalRequestedDiskCount := getQualifyingDisksRequestedByPod(ns, requestedVolumes)
-	// TODO: change to len(diskRequestedByPod) when we fix #1097 in the controller
-	if len(requestedVolumes) == 0 {
+	diskRequestedByPod, _ := getQualifyingDisksRequestedByPod(ns, requestedVolumes)
+	// when there is no v2 disks requested by pod, don't filter and return
```
edreed, line 154: `when there are no v2 disks requested by the pod, skip filtering and return all nodes.`

pkg/azdiskschedulerextender/azdiskschedulerextender.go
```diff
@@ -183,38 +186,17 @@ func filter(context context.Context, schedulerExtenderArgs schedulerapi.Extender
...
+				failedNodes[node.Name] = fmt.Sprintf("AzDriverNode for %s is not ready.", node.Name)
 			}
 		} else {
-			failedNodes[node.Name] = fmt.Sprintf("AzDriverNode for %s is not ready.", node.Name)
+			// if AzDriverNode cannot be found for the node, either 1) AzDriverNode cache is stale or 2) the node plugin has not started and needs to be scheduled.
+			// so add the node to filtered node list.
+			storeFilteredNodeNames = append(storeFilteredNodeNames, node.Name)
 		}
 	}
 	filteredNodeNames = &storeFilteredNodeNames
```
landreasyan, line 202: Should we add the check for empty ` filteredNodeNames`?

sunpa93: Please refer to the comment in the group chat. I think as long as the extender can handle the `AzDriverNode` missing case, we should maintain the behavior of returning an empty filtered node list if all nodes in cluster are unhealthy. This should retrigger scheduling until healthy nodes become available. What do you think?

pkg/azdiskschedulerextender/azdiskschedulerextender.go
```diff
@@ -432,6 +420,8 @@ func setNodeScoresToZero(nodes []v1.Node) (priorityList schedulerapi.HostPriorit
 	return
 }
 
+// diskRequestedByPod == total number of v2 disks requested by pod
+// totalRequestedDiskCount = total number of pvc disks requested by pod
 func getQualifyingDisksRequestedByPod(ns string, requestedVolumes []v1.Volume, filterOutAccessModes ...v1.PersistentVolumeAccessMode) (disksRequestedByPod map[string]struct{}, totalRequestedDiskCount int) {
```
landreasyan, line 425: is `totalRequestedDiskCount` used anywhere now? I think we can probably remove it.

pkg/azdiskschedulerextender/azdiskschedulerextender.go
```diff
@@ -354,21 +335,28 @@ func getKubernetesClientset() (*kubernetes.Clientset, error) {
...
 
-	score := int64(replicaCount*100) - int64(attachingReplicaCount*25) - int64((totalVolumeAttachmentCount-replicaCount)*10)
+	var remainingNodeCapacity int
+	if nodeMaxCapacity, err := azureutils.GetNodeMaxDiskCountWithLabels(node.Labels); err != nil {
+		remainingNodeCapacity = 0
+	} else {
+		remainingNodeCapacity = nodeMaxCapacity - allAttachmentsCount
+	}
+
+	score := int64(createdAttachmentsForPodCount*100) + (100 - int64(attachingAttachmentsForPodCount*10)) + int64(remainingNodeCapacity)
```
landreasyan, line 359: Is there a reason we are updating the scoring logic?

sunpa93: It's a small update to account for node's remaining disk count.

landreasyan: `createdAttachmentsForPodCount`  with the current logic, already included `attachingAttachmentsForPodCount` so basically our formula now is something like: 
`(x+y)*100 +100-y*10+z = 100x+90y+z+100

landreasyan: So, nodes that have more attachments in `Attaching` or `Failed` state will score higher with this logic.

pkg/azdiskschedulerextender/azdiskschedulerextender.go
```diff
@@ -183,38 +186,17 @@ func filter(context context.Context, schedulerExtenderArgs schedulerapi.Extender
...
 				filteredNodes = append(filteredNodes, node)
 				storeFilteredNodeNames = append(storeFilteredNodeNames, node.Name)
+			} else {
+				failedNodes[node.Name] = fmt.Sprintf("AzDriverNode for %s is not ready.", node.Name)
 			}
 		} else {
-			failedNodes[node.Name] = fmt.Sprintf("AzDriverNode for %s is not ready.", node.Name)
+			// if AzDriverNode cannot be found for the node, either 1) AzDriverNode cache is stale or 2) the node plugin has not started and needs to be scheduled.
+			// so add the node to filtered node list.
+			storeFilteredNodeNames = append(storeFilteredNodeNames, node.Name)
```
landreasyan, line 199: This only works as long as the assumption that all nodes in the cluster will have AzDriverNode created for them at one point. Because otherwise we could schedule a pod on a node without the driver and it will fail provisioning disks. This is ok for now because the strategy of the extender has really shifted to only work for failover, but this is something we should watch out for in the future in case our approach changes or there are some node affinity rules that come into play.

pkg/azdiskschedulerextender/azdiskschedulerextender.go
landreasyan: We could also count only `Attached` volumes in `createdAttachmentsForPodCount`

pkg/azdiskschedulerextender/main.go
```diff
@@ -42,6 +43,7 @@ func init() {
 var (
 	metricsAddress              = flag.String("metrics-address", "0.0.0.0:29606", "export the metrics")
 	azDiskSchedulerExtenderPort = flag.String("port", "8889", "port used by az scheduler extender")
+	driverName                  = flag.String("drivername", consts.DefaultDriverName, "name of the driver")
```
landreasyan, line 46: We should make sure this is added to our scheduler config in aks

pkg/azdiskschedulerextender/azdiskschedulerextender.go
sunpa93: >So, nodes that have more attachments in Attaching state will score higher with this logic.

It's `(100 - (attaching * 10)`, so we are counting number of attached attachments here and prioritizing nodes with more requested disks attached.

pkg/azdiskschedulerextender/main.go
sunpa93: It is for now the same as the default driver name so we can modify the chart when necessary.

pkg/azdiskschedulerextender/azdiskschedulerextender.go
sunpa93: It is required for our driver to create `AzDriverNode` object for all nodes in cluster in order for our controller to function and a lot of logic in the controller, including `node controller` and `node availability controller` rely on this premise. We deploy our node controller as a Daemonset; this will ensure that all nodes in cluster have its own node controller running, which will also create an AzDriverNode object for its host node.

pkg/azdiskschedulerextender/azdiskschedulerextender.go
landreasyan: what I am saying is attaching is counted twice and when you simplify the formula, you get what I have.

pkg/azdiskschedulerextender/main.go
landreasyan: This will break with side-by-side setup, so we do need to modify it

pkg/azdiskschedulerextender/azdiskschedulerextender.go
sunpa93: > what I am saying is attaching is counted twice and when you simplify the formula, you get what I have.

Yes, but you see that in the equation `x = number of requested attachments that's attached` and `y = number of requested attachments that's not yet attached`.

So, we give higher coefficient for attached attachments, and a slightly lower coefficient for attaching / pending attachments.

pkg/azdiskschedulerextender/main.go
sunpa93: I thought you were referring to the scheduler extender chart in AKS rp.

But yes, we will need to update the helm chart in the main_v2 repo.

pkg/azdiskschedulerextender/azdiskschedulerextender.go
landreasyan: x includes y is what I am saying. line 273. So you end up with `100x+90y+z+100` this is not what we want

landreasyan: Actually, let me rephrase that - it is close to what we want but wouldn't it be better to have a bigger difference 100 vs 90 and I don't see the point of keeping the 100 constant

pkg/azdiskschedulerextender/main.go
landreasyan: Yep, I was initially and then I remembered about the side-by-side setup :D I think we should pass this value explicitly in both places because it will break the scheduler extender's logic if the driver gets deployed under a different name

pkg/azdiskschedulerextender/azdiskschedulerextender.go
landreasyan: I think we can do something like this: 
`score := int64(createdAttachmentsForPodCount*100) - int64(attachingAttachmentsForPodCount*50) + int64(remainingNodeCapacity)`
This will always be positive and attached volumes will be more strongly prioritized

pkg/azdiskschedulerextender/azdiskschedulerextender.go
landreasyan: The recent update achieves what I was hoping.

pkg/azdiskschedulerextender/azdiskschedulerextender.go
sunpa93: The subtraction from 100 was so that we can keep the score in a `xyz` fashion where each digit is kind of reserved for each priority criteria.  `x`, which is in the 100th place, to record score for mostly highly prioritized requirement, `y`, in the 10th place, for the second most highly prioritized requirement, and so on. I understand that this will not always be the case because `x`, `y`, `z` will not always be less than 10 but it does keep our priorities straight.

sunpa93: > I think we can do something like this: `score := int64(createdAttachmentsForPodCount*100) - int64(attachingAttachmentsForPodCount*50) + int64(remainingNodeCapacity)` This will always be positive and attached volumes will be more strongly prioritized

This is then equating 1 attachment in attached state with 2 attachments in attaching state. Say there are 2 nodes, A and B. A has 2 attachments: 0 attached and 2 attaching. B has 1 attachment: 1 attached and 0 attaching. We end up giving same scores for the two nodes and worse, if A has less remaining node capacity than B, we will end up giving higher score to B, which is not what we want.

sunpa93: Given that we have a clear ordering of preference in prioritization requirements, I think we should have our score should reflect on this gradation between each requirement properly.

pkg/azdiskschedulerextender/main.go
sunpa93: Sounds good.

pkg/azdiskschedulerextender/azdiskschedulerextender.go
landreasyan: In your example it will be: 3*100-2*50 =200 vs 2*100-2*50=0, we will not be really equalizing the scores but yeah to avoid them being multiples, we can pick another number instead of 50.

landreasyan: And I still think we should keep the baseline at 0 not 100.

sunpa93: What I am saying is let's reserve each digit for individual requirement. 100s for score for number of _created_ attachments, 10s for score number of _attached_ attachments, and 1s for score for number of remaining node capacity.

sunpa93: We could make this clearer by using `10 * number of attached attachments` instead of `100 - 10 * number of attaching attachments`

sunpa93: > In your example it will be: `3*100-2*50 =200` vs `2*100-2*50=0`, we will not be really equalizing the scores but yeah to avoid them being multiples, we can pick another number instead of 50.

In the example, node A would have `2 * 100 - 2 * 50 = 100` and B would have `1 * 100 - 0 * 50 = 100`. So these 2 nodes would end up with the same score.

landreasyan: line 273 you count all volumes regardless of status in createdAttachmentsForPodCount, line 275 those with Attaching state get counted **again** in  nodeNameToNumAttachingAttachmentsForPodsMap.

landreasyan: `createdAttachmentsForPodCount` already includes `nodeNameToNumAttachingAttachmentsForPodsMap`

sunpa93: Yes, I am aware as I commented above.

> > what I am saying is attaching is counted twice and when you simplify the formula, you get what I have.
> 
> Yes, but you see that in the equation `x = number of requested attachments that's attached` and `y = number of requested attachments that's not yet attached`.
> 
> So, we give higher coefficient for attached attachments, and a slightly lower coefficient for attaching / pending attachments.

sunpa93: What I am saying is we can make the prioritization more discrete by keeping them in separate digits. We prioritize nodes based on the following order: 1) number of requested attachments created on the node, 2) number of requested attachments attached on the node, and 3) number of remaining node capacity.

And we should **always** prioritize node with more requested attachments over nodes with less created attachments even if the other node has its attachments in _attached_ state. And we can keep this clear ordering by putting those 3 scores in separate digits.

landreasyan: :D

landreasyan: Yes, that is what I am also saying and the right formula for calculating that score is: 
`core := int64(createdAttachmentsForPodCount*100) - int64(attachingAttachmentsForPodCount*50) + int64(remainingNodeCapacity)` and not what you have because `createdAttachmentsForPodCount=x+y` and `attachingAttachmentsForPodCoun=y`

sunpa93: `score := int64(createdAttachmentsForPodCount*100) - int64(attachingAttachmentsForPodCount*50) + int64(remainingNodeCapacity)` will not guarantee this ordering as described above. We equate node A with 2 attachments (0 attached, 2 attaching) and B with 1 attachment (1 attached), as they will both end up with `100`.

landreasyan: And with the score you suggest the 2 attaching will score higher than 1 attached 180 vs 100.

sunpa93: And whatever coefficient you give for `attachingAttachmentsForPodCount`, it's not going to be any different. Say we give `30` instead of `50`.

Node A: `4 attachments (0 attached, 4 attaching)` will have `4 * 100 - 4 * 30= 180`
Node B: `2 attachments (2 attached, 0 attaching)` will have `2 * 100 - 0 * 30 = 200`

We end up giving higher score to B when we should be giving higher score to A.

landreasyan: and like I said, you can make 50 to 25 where 1 attachment will be equivalent to to 4 attaching, which is what we had before

sunpa93: I think `100 - attachingAttachmentsForPodCount * 10` threw you off. To keep things less confusing and more consistent, why don't we use `attachedAttachmentsForPodCount` instead of `attachingAttachmentsForPodCount`. And make the scoring formula more straightforward by replacing `100 - attachingAttachmentsForPodCount * 10` with `10 * attachedAttachmentsForPodCount`?

i.e. `score := 100 * createdAttachmentsForPodCount + 10 * attachedAttachmentsForPodCount + remainingNodeCapacity`

landreasyan: Ok, so why do we want to give higher score to A? This is something new

sunpa93: We want to give higher score to A because if we schedule pod on B, we have higher risk of having to detach and reattach volume in case the volume's `maxShares` is saturated.

sunpa93: Always scheduling pod to node with most requested attachments minimizes this risk.

landreasyan: I don't think anything was throwing me off and this is almost exactly what I am trying to get to `100 - attachingAttachmentsForPodCount * 10` doesn't have the intend that you wanted because it was giving a higher coefficient to attaching volumes than what we wanted. If you don't want to penalize for volumes being in attaching state but that is counted as advantage that's a different discussion and you can keep it as addition instead of the penalty like our previous logic was.

landreasyan: I guess what I also don't understand is why you want to prioritize attaching volumes more than the attached ones. This workflow is exercised on failover and in most cases, we would have attached volumes on some node while there might be another node trying to attach them to new node because the initial node with primaries failed.

sunpa93: I am not trying to prioritize attaching volumes over attached ones. That's why I suggested
`score := 100 * createdAttachmentsForPodCount + 10 * attachedAttachmentsForPodCount + remainingNodeCapacity`

sunpa93: With the addition of logic in `CrdProvisioner` to detach replicas to host primaries, we are now more likely to break the replica grouping. And that is why we need to enclose with scheduleextender logic to minimize the risk of scheduling pod onto a node where we have to detach and reattach more volumes.

landreasyan: OK, yes this is exactly what I am trying to say as well, except with attaching volumes.

However, with this formula we have 110*attached+100*attaching and with the formula initially you had 100*attached+90*attaching and I was wondering if we should keep this ratio bigger that is why I suggested 50 but replace 50 with 10 and you get the same thing as your formula but simpler.  `score := 100 * createdAttachmentsForPodCount - 10 * attaching + remainingNodeCapacity` you still get `100*attached + 90*attaching+remainingNodeCapacity`. This is all I am trying to get at.

sunpa93: Yeah, I think we are in agreement. With `110*attached+100*attaching`, we will never end up scheduling pod to a node with less requested attachment and when the two nodes have the same number of requested attachments, this will allow us to differentiate between the two because we give higher coefficient to `attached`. 

When we penalize, no matter what coefficient we give for penalty, we could theoretically end up scheduling pod on node with less attachments, and that's all I am trying to get at :D

landreasyan: Yes, you are right with the new logic penalizing for being in `attaching` state is no longer for our advantage

landreasyan: And hopefully this is more readable to avoid such lengthy discussions in the future :D

sunpa93: Will do. I will add comment to describe above as well.

sunpa93: There wasn't much logic ourside of this function to update anyway. I have updated `int64(attachedAttachmentsForPodCount * 10)` instead for consistency and clarity.

PR #1457, opened 2022-08-09
[V2] feat: Making scheduler extender node cache capable

landreasyan:
**What type of PR is this?**
Current configuration of scheduler extender was setting `nodeCacheCapable=false` which was resulting in the scheduler passing unnecessary node metadata to the extender. This PR sets `nodeCacheCapable=true` and updates the extender's implementation to not rely on NodeList value passed from the scheduler. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azdiskschedulerextender/main.go
```diff
@@ -116,13 +116,6 @@ func handleFilterRequest(response http.ResponseWriter, request *http.Request) {
 		return
 	}
 
-	allNodes := args.Nodes.Items
-	if len(allNodes) == 0 {
```
edreed, line 120: Should we make an equivalent check of `args.NodeNames`?

charts/latest-v2/azuredisk-csi-driver/templates/csi-azuredisk-scheduler-extender.yaml
```diff
@@ -24,7 +24,7 @@ data:
       filterVerb: "filter"
       prioritizeVerb: "prioritize"
       weight: 1
-      nodeCacheCapable: false
```
edreed, line 27: We need to update [deploy/latest-v2/csi-azuredisk-scheduler-extender.yaml](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/main_v2/deploy/latest-v2/csi-azuredisk-scheduler-extender.yaml#L26) as well.

pkg/azdiskschedulerextender/main.go
landreasyan: I am not sure we need this check in main and respond with BadRequest, we were not doing this for prioritize. I think if nodes are empty, we should just return an empty list?

edreed: I'm not sure what the expected behavior is if an empty list is passed by kube-scheduler here. Given that we don't generally fail operations on unexpected input and simply return what was given to us, returning an empty list in this case seems reasonable.

abhisheksinghbaghel (approved):

PR #1463, opened 2022-08-11
[V2] chore: Create v2.0.0-beta.6 release

landreasyan:
**What type of PR is this?**
Creating a new release with scheduler extender performance optimizations. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


charts/index.yaml
```diff
@@ -12,7 +21,7 @@ entries:
     version: v2.0.0-beta.6
```
edreed, line 21: We'll need to update the `latest-v2` chart to beta.7 as well.

sunpa93 (approved):
We need to update the `latest_v2` charts as well but let's keep it for a separate PR given that we are short in time and a new commit will require new test runs which will take longer to complete?

EDIT:
If helm's going to be confused of `appVersions` with `same version`, let's make a new commit.

charts/index.yaml
landreasyan: Oh, that's right!

sunpa93 (changes requested):
`latest-v2/azuredisk-csi-driver-v2.0.0-beta.6.tgz` needs to be removed.

charts/index.yaml
```diff
@@ -3,7 +3,25 @@ entries:
...
+    appVersion: v2.0.0-beta.6
+    created: "2022-08-11T14:09:09.585067103-07:00"
+    description: Azure disk Container Storage Interface (CSI) Storage Plugin
+    digest: 0e7280d2bcd752668c4439078ffdfc4567ac97af97d8baa07c322b99c34fb741
+    name: azuredisk-csi-driver
+    urls:
+    - https://raw.githubusercontent.com/kubernetes-sigs/azuredisk-csi-driver/main_v2/charts/v2.0.0-beta.6/azuredisk-csi-driver-v2.0.0-beta.6.tgz
+    version: v2.0.0-beta.6
+  - apiVersion: v1
+    appVersion: latest-v2
```
sunpa93, line 23: Probably will need to run `helm repo index` once more?

edreed, line 23: There's a duplicate entry for `latest-v2`. Re-index now that the beta.6 tgz has been deleted?

landreasyan: yep 🤦‍♀️

PR #1486, opened 2022-08-24
[V2] chore: recreate pod-failover CRD

connorbradshaw10:
**What type of PR is this?**
/kind cleanup

- Switches podfailure CRD to be auto-generated
- Clears up verbage in pod-failover README
- Adds kube-config flags and defines level of inheritance

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


hack/update-codegen.sh
```diff
@@ -24,3 +24,10 @@ bash "${CODEGEN_PKG}"/generate-groups.sh "deepcopy,client,informer,lister" \
   sigs.k8s.io/azuredisk-csi-driver/pkg/apis \
   azuredisk:v1beta1,v1beta2 \
   --go-header-file "${SCRIPT_ROOT}"/hack/boilerplate/boilerplate.generatego.txt
+
```
edreed, line 27: Let's put this in a separate script since the azuredisk CRDs and test CRDs won't need to be updated at the same time, e.g. `update-test-crds.sh`.

edreed:
This is a cleanup from a previous PR, so `chore:` would be the more appropriate commit prefix.

test/podFailover/apis/podfailure/v1beta1/register.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright The Kubernetes Authors.
+Copyright 2017 The Kubernetes Authors.
```
edreed, line 2: Copyrights for new code should use the current year.

test/podFailover/apis/podfailure/v1beta1/types.go
```diff
@@ -14,39 +14,45 @@ See the License for the specific language governing permissions and
...
-	Items []PodFailure `json:"items"`
-}
-
 type PodFailureSpec struct {
 	PodName string `json:"podName"`
 }
 
 type PodFailureStatus struct {
+	// +optional
 	FailureType string `json:"failureType"`
```
edreed, line 46: Optional fields, should also be tagged with `omitempty`

test/podFailover/controller/cmd/main.go
```diff
@@ -35,9 +36,12 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
-	podfailure "sigs.k8s.io/azuredisk-csi-driver/test/podFailover/api"
+	podfailure "sigs.k8s.io/azuredisk-csi-driver/test/podFailover/apis/client/clientset/versioned"
+	podfailurev1beta1 "sigs.k8s.io/azuredisk-csi-driver/test/podFailover/apis/podfailure/v1beta1"
 )
 
+var kubeconfig = flag.String("kube-config", "", "kube config path for cluster")
```
edreed, line 43: The standard parameter name for Kubernetes apps is `kubeconfig`. We should use it here and elsewhere.

test/podFailover/controller/cmd/main.go
```diff
@@ -251,18 +265,18 @@ func getFailureType(pod v1.Pod) (string, error) {
...
-	podfailureobj, err := clientset.PodFailures(namespace).Get(ctx, podName, metav1.GetOptions{})
+func getCrdAndReportFailure(ctx context.Context, podName string, namespace string, failureType string, clientset *podfailure.Clientset) error {
+	podfailureobj, err := clientset.PodfailureV1beta1().PodFailures(namespace).Get(ctx, podName, metav1.GetOptions{})
 	if err != nil {
 		return fmt.Errorf("failed to get crd for pod %q: %v", podName, err)
 	}
 
-	updatedPodFailure := &podfailure.PodFailure{}
+	updatedPodFailure := &podfailurev1beta1.PodFailure{}
 	podfailureobj.DeepCopyInto(updatedPodFailure)
```
edreed, line 275: You could simply use `DepyCopy`, e.g. `updatedPodFailure := podfailureobj.DeepCopy()`, and avoid the need for importing the `podfailurev1beta1` package. It would reduce the churn if/when we update the API version.

test/podFailover/workload/cmd/main.go
```diff
@@ -67,15 +68,18 @@ func main() {
 	flag.Parse()
 	var config *rest.Config
 	var err error
```
edreed, line 70: It doesn't seem like we need to declare this variable outside the scope of the `if` statement.

connorbradshaw10: the scope of config creations is weird within different flows so the defined config at the top makes sure all versions of config are instantiated to the same variable

edreed: I was referring to the `err` variable. This is non-blocking.

PR #1500, opened 2022-09-07
[V2] test: add and edit unit tests to replica.go

hccheng72:
**What type of PR is this?**
/kind test

Add and edit unit tests to `pkg/controller/replica.go`

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/replica_test.go
+++ b/pkg/controller/replica_test.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
+
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -68,7 +69,6 @@ func TestReplicaReconcile(t *testing.T) {
 				replicaAttachment := testReplicaAzVolumeAttachment
 				now := metav1.Time{Time: metav1.Now().Add(-1000)}
 				replicaAttachment.DeletionTimestamp = &now
-				replicaAttachment.Status.State = azdiskv1beta2.Detaching
 
 				newVolume := testAzVolume0.DeepCopy()
 				newVolume.Status.Detail = &azdiskv1beta2.AzVolumeStatusDetail{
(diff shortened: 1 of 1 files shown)


pkg/controller/replica_test.go
```diff
@@ -68,7 +69,7 @@ func TestReplicaReconcile(t *testing.T) {
 				replicaAttachment := testReplicaAzVolumeAttachment
 				now := metav1.Time{Time: metav1.Now().Add(-1000)}
 				replicaAttachment.DeletionTimestamp = &now
-				replicaAttachment.Status.State = azdiskv1beta2.Detaching
+				replicaAttachment.Status.State = azdiskv1beta2.DetachmentFailed
```
sunpa93, line 72: Let's remove this line for this test case, attachment's state should not have any impact on this test case and will only confuse the readers going forward.

Instead, consider adding a new test case "[Success] Should delete replica attachment if attachment fails and requeue" to ensure that the reconciler will delete replica attachment in `AttachmentFailed` state and return reconcile result, where `requeue == true`

sunpa93: >Instead, consider adding a new test case "[Success] Should delete replica attachment if attachment fails and requeue" to ensure that the reconciler will delete replica attachment in AttachmentFailed state and return reconcile result, where requeue == true

Oh, you have already added "[Success] Should delete a failed-attachment replica." Let's go ahead and remove this line in that case 👍

pkg/controller/replica_test.go
```diff
@@ -114,6 +115,80 @@ func TestReplicaReconcile(t *testing.T) {
...
+					replicaAttachment)
+
+				mockClients(controller.cachedClient.(*mockclient.MockClient), controller.azClient, controller.kubeClient)
+				return controller
+			},
+			verifyFunc: func(t *testing.T, controller *ReconcileReplica, result reconcile.Result, err error) {
+				require.NoError(t, err)
+				require.False(t, result.Requeue)
+
+				_, err = controller.azClient.DiskV1beta2().AzVolumeAttachments(testNamespace).Get(context.TODO(), testPersistentVolume0Name+"-"+testNode1Name+"-attachment", metav1.GetOptions{})
```
sunpa93, line 188: Use `testReplicaAzVolumeAttachment.Name()` instead.

pkg/controller/replica_test.go
hccheng72: I edited the attachment's state to `DetachmentFailed` for testing [this case](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/fde668a0d77effb4af845a115572180f85a65f9f/pkg/controller/replica.go#L95). And we can verify if the state has been updated to `ForceDetachPending` right before [delete the attachment manually in the verifyFun](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/fde668a0d77effb4af845a115572180f85a65f9f/pkg/controller/replica_test.go#L96). Not sure if it's needed?

sunpa93: For this test case, we want to check if a new replica attachment is created when one is deleted.

hccheng72: Got it. I've updated the code.

sunpa93 (approved):
/lgtm

PR #1501, opened 2022-09-07
[V2] test: add and edit unit tests to pod.go

hccheng72:
**What type of PR is this?**
/kind test

Add a unit test to `pkg/controller/pod.go`
Add the other node to existing tests such that they can test if replicas are created in the same node.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/pod_test.go
+++ b/pkg/controller/pod_test.go
@@ -133,6 +133,7 @@ func TestPodReconcile(t *testing.T) {
 					&testPersistentVolume1,
 					&testNode0,
 					&testNode1,
+					&testNode2,
 					newPod)
 
 				mockClients(controller.cachedClient.(*mockclient.MockClient), controller.azClient, controller.kubeClient)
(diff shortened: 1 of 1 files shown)


pkg/controller/pod_test.go
```diff
@@ -221,6 +224,48 @@ func TestPodReconcile(t *testing.T) {
...
+				}
+
+				controller := NewTestPodController(
+					mockCtl,
+					testNamespace,
+					newVolume,
+					newAttachment,
+					&testNode0,
+					&testNode1,
+					&testPersistentVolume0)
```
sunpa93, line 246: It would be good to actually register a pod object, run `addPod`, and delete the registered pod to make sure that the next reconcile removes entry added from `addPod`.

sunpa93 (approved):
/lgtm

PR #1505, opened 2022-09-12
[V2] [WIP] test: add e2e tests for conflicting UUIDs issue and supporting for se…

alice-zheyan-yu:
Add e2e tests for volume snapshot with conflicting UUID issue

**What type of PR is this?**
/kind test

When a disk is partitioned, each partition gets a unique identifier (UUID) and the OS uses this to uniquely identify the partition. Problems arise when two partitions with the same UUID are mounted on the same node. This isn’t likely with newly created managed disks. However, it can happen when you create a disk, take a snapshot of it and the create a new disk using the snapshot as its data source.
Recently, the Disk Resource Provider (DiskRP) team has made a change to managed disk creation from snapshot that ensures the copy gets a new partition UUID. It’s currently in limited flighting and out test subscription (Xstore Container Storage) has the feature enabled. This PR modifies the test implementation to allow testing of this scenario and others. Specifically, it supports an option to schedule the overwrite pod and the snapshot pot on the same or on different nodes. This PR won’t be able to check this in upstream until the feature has been made public. Will need another PR to clean up all the TODOs.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,6 +138,7 @@ ifdef TEST_WINDOWS
 		--set controller.logLevel=6 \
 		--set schedulerExtender.replicas=1 \
 		--set node.logLevel=6 \
+		--set snapshot.enabled=true \
 		--set cloud=$(CLOUD)
 else
 	helm $(HELM_COMMAND) azuredisk-csi-driver charts/${CHART_VERSION}/azuredisk-csi-driver --namespace kube-system --wait --timeout=15m -v=5 --debug \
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -138,6 +138,7 @@ ifdef TEST_WINDOWS
 		--set controller.logLevel=6 \
 		--set schedulerExtender.replicas=1 \
 		--set node.logLevel=6 \
+		--set snapshot.enabled=true \
```
landreasyan, line 141: I am assuming this was for local testing. This shouldn't be committed.

landreasyan (approved):

PR #1508, opened 2022-09-12
[V2] fix: adding scheduler to pod-failover workload pods

connorbradshaw10:
**What type of PR is this?**

/kind bug

Adds scheduler extender to pod failover workload pods so v2 with replica works as intended
Fixes same node failovers
Changes SKU of pod failover template to make sure the max disks is enough

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/test/e2e/manifest/pod-failover.json
+++ b/test/e2e/manifest/pod-failover.json
@@ -30,13 +30,13 @@
         "masterProfile": {
             "count": 1,
             "dnsPrefix": "{dnsPrefix}",
-            "vmSize": "Standard_D4s_v3"
+            "vmSize": "Standard_D8s_v3"
         },
         "agentPoolProfiles": [
             {
                 "name": "agentpool1",
                 "count": 3,
-                "vmSize": "Standard_D4s_v3",
+                "vmSize": "Standard_D8s_v3",
                 "availabilityProfile": "AvailabilitySet",
                 "distro": "aks-ubuntu-18.04"
             }
--- a/test/podFailover/workload/scenarios/azDiskV2-Replicas.yaml
+++ b/test/podFailover/workload/scenarios/azDiskV2-Replicas.yaml
@@ -10,6 +10,9 @@ storageClass:
     allowVolumeExpansion: true
   provisioner: disk.csi.azure.com
 
+# use only with standalone installations of azdisk or it will break the pods
+schedulerName: csi-azuredisk-scheduler-extender
+
 namespace: azdiskv2-pod-failover-1pod1pvc
 
 workloadType: 1pod1pvc1replica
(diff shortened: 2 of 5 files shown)


test/podFailover/workload/scenarios/azDiskV2-Replicas.yaml
```diff
@@ -10,6 +10,8 @@ storageClass:
     allowVolumeExpansion: true
   provisioner: disk.csi.azure.com
 
+schedulerName: csi-azuredisk-scheduler-extender
```
edreed, line 13: This default will only work with Standalone or SxS deployments of the driver. When using the V2 driver included in AKS, the scheduler extender is part of the default scheduling configuration and a scheduler name should not be provided.

connorbradshaw10: yes, this is intended to be added to fix the pipelines which use helm to install V2. are you saying that if schedulerName is added to an AKS installed AzDisk it will break the pod?

edreed: The pod won't be scheduled because the requested scheduler profile does not exist.

connorbradshaw10: ok. will make a note of that in the yaml example

test/e2e/manifest/pod-failover.json
```diff
@@ -30,13 +30,13 @@
...
             "dnsPrefix": "{dnsPrefix}",
-            "vmSize": "Standard_D4s_v3"
+            "vmSize": "Standard_D8s_v3"
         },
         "agentPoolProfiles": [
             {
                 "name": "agentpool1",
                 "count": 3,
-                "vmSize": "Standard_D4s_v3",
+                "vmSize": "Standard_D8s_v3",
```
nearora-msft, line 39: Why is this change made?

nearora-msft: Nvm, I just read the description.

test/podFailover/workload/scenarios/sameNodeFailover.yaml
```diff
@@ -14,7 +14,7 @@ namespace: samenode-pod-failover-1pod1pvc
 
 podCount: 3
 
-pvcCount: 3
+pvcCount: 1
```
nearora-msft, line 17: Is it pvcPerPod? If that is the case, we could change the name to pvcPerPodCount in order to avoid confusion.

nearora-msft:
lgtm other than a minor comment

test/e2e/manifest/pod-failover.json
connorbradshaw10: this is made so I can support up to 16 disks on a node. the pipelines run 1pod1pvc 1pod3pvc 3pod3pvc concurrently, so there are a maximum of 14 disk being attached to a node. doing this to avoid mount failures

edreed:
Changes do not impact E2E tests.

PR #1515, opened 2022-09-14
[V2] Adding options to configure attach/detach rate limiter on azure client

landreasyan:
**What type of PR is this?**
Adding options to configure attach/detach rate limiter on azure client. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/charts/latest-v2/azuredisk-csi-driver/values.yaml
+++ b/charts/latest-v2/azuredisk-csi-driver/values.yaml
@@ -74,6 +74,9 @@ controller:
   provisionerWorkerThreads: 40
   attacherWorkerThreads: 500
   vmssCacheTTLInSeconds: -1
+  enableAzureClientAttachDetachRateLimiter: true
+  azureClientAttachDetachRateLimiterQPS: 1.3
+  azureClientAttachDetachRateLimiterBucket: 80
   logLevel: 5
   tolerations:
     - key: "node-role.kubernetes.io/master"
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -47,30 +47,6 @@ import (
 	azure "sigs.k8s.io/cloud-provider-azure/pkg/provider"
 )
 
-// DriverOptions defines driver parameters specified in driver deployment
-type DriverOptions struct {
-	NodeID                     string
-	DriverName                 string
-	VolumeAttachLimit          int64
-	EnablePerfOptimization     bool
-	CloudConfigSecretName      string
-	CloudConfigSecretNamespace string
-	CustomUserAgent            string
-	UserAgentSuffix            string
-	UseCSIProxyGAInterface     bool
-	EnableDiskOnlineResize     bool
-	AllowEmptyCloudConfig      bool
-	EnableAsyncAttach          bool
-	EnableListVolumes          bool
-	EnableListSnapshots        bool
-	SupportZone                bool
-	GetNodeInfoFromLabels      bool
-	EnableDiskCapacityCheck    bool
-	VMSSCacheTTLInSeconds      int64
-	VMType                     string
-	RestClientQPS              int
-}
-
 // CSIDriver defines the interface for a CSI driver.
 type CSIDriver interface {
 	csi.ControllerServer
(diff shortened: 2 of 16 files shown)


charts/latest-v2/azuredisk-csi-driver/values.yaml
```diff
@@ -74,6 +74,9 @@ controller:
   provisionerWorkerThreads: 40
   attacherWorkerThreads: 500
   vmssCacheTTLInSeconds: -1
+  enableAzureClientAttachDetachRateLimiter: false
```
sunpa93, line 77: Are we going to disable attach detach rate limiter by default?

charts/latest-v2/azuredisk-csi-driver/values.yaml
```diff
@@ -74,6 +74,9 @@ controller:
...
+  azureClientAttachDetachRateLimiterQPS: 2.6
```
sunpa93, line 78: Does this mean our testing region allow 468 queries per 3 minutes?

charts/latest-v2/azuredisk-csi-driver/values.yaml
landreasyan: This is for standalone runs in Aurora. I'll check if there is a way to configure this in Aurora instead of here.

charts/latest-v2/azuredisk-csi-driver/values.yaml
landreasyan: That is the increased limit in some regions but not the one we are currently testing

landreasyan: Yes, and will keep these values to defaults.

pkg/azuredisk/azuredisk.go
```diff
@@ -90,11 +93,14 @@ type hostUtil interface {
...
-	deviceHelper optimization.Interface
-	nodeInfo     *optimization.NodeInfo
-	ioHandler    azureutils.IOHandler
-	hostUtil     hostUtil
+	ready                                    chan struct{}
+	deviceHelper                             optimization.Interface
+	nodeInfo                                 *optimization.NodeInfo
+	ioHandler                                azureutils.IOHandler
+	hostUtil                                 hostUtil
+	enableAzureClientAttachDetachRateLimiter bool
```
edreed, line 101: Please do not add any configuration related fields to `DriverCore`. The V2 driver uses the `AzDiskDriverConfiguration` directly making these values redundant. We don't build the V1 driver out of this branch, so you can either add the fields to the `Driver` struct or simply pass default values for the V1 driver.

pkg/azurediskplugin/main.go
```diff
@@ -44,39 +44,42 @@ var (
...
+	enableListVolumes                        = flag.Bool("enable-list-volumes", consts.DefaultEnableListVolumes, "boolean flag to enable ListVolumes on controller")
+	enableListSnapshots                      = flag.Bool("enable-list-snapshots", consts.DefaultEnableListSnapshots, "boolean flag to enable ListSnapshots on controller")
+	enableDiskCapacityCheck                  = flag.Bool("enable-disk-capacity-check", consts.DefaultEnableDiskCapacityCheck, "boolean flag to enable volume capacity check in CreateVolume")
+	kubeClientQPS                            = flag.Int("kube-client-qps", consts.DefaultKubeClientQPS, "QPS for the rest client. Defaults to 15.")
+	vmssCacheTTLInSeconds                    = flag.Int64("vmss-cache-ttl-seconds", consts.DefaultVMSSCacheTTLInSeconds, "vmss cache TTL in seconds (600 by default)")
+	enableAzureClientAttachDetachRateLimiter = flag.Bool("enable-attach-detach-rate-limiter", consts.DefaultEnableAzureClientAttachDetachRateLimiter, "azure client rate limiter for attach/detach operations (true by default)")
```
edreed, line 70: The command-line parameters have been deprecated, though we still use them in the AKS RP. I am OK leaving these in for now, but we should add them to the list of deprecated parameters [here](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/cee1ff90d97c3776fd140b1491b4d04aad9a4dea/pkg/azureconstants/azure_constants.go#L207).

edreed: I see now that you have already added them.

pkg/azureutils/azure_disk_utils.go
```diff
@@ -525,7 +525,14 @@ func CreateValidDiskName(volumeName string, usedForLabel bool) string {
 }
 
 // GetCloudProviderFromClient get Azure Cloud Provider
-func GetCloudProviderFromClient(kubeClient *clientset.Clientset, secretName, secretNamespace, userAgent string, allowEmptyCloudConfig bool) (*azure.Cloud, error) {
+func GetCloudProviderFromClient(
+	kubeClient *clientset.Clientset,
+	secretName, secretNamespace,
+	userAgent string,
+	allowEmptyCloudConfig bool,
+	enableAzureClientAttachDetachRateLimiter bool,
```
edreed, line 533: We should consider passing `CloudConfiguration` here rather than continuing to add new parameters.

pkg/provisioner/cloudprovisioner.go
```diff
@@ -102,8 +105,19 @@ func NewCloudProvisioner(
 	enableOnlineDiskResize bool,
 	allowEmptyCloudConfig bool,
 	enableAsyncAttach bool,
+	enableAzureClientAttachDetachRateLimiter bool,
```
edreed, line 108: Consider passing `CloudConfiguration` rather than adding new parameters.

edreed (approved):

PR #1523, opened 2022-09-20
[V2] test: add a failover scale test

hccheng72:
**What type of PR is this?**
/kind test

- Removing unused scale tests.
- Adding a command line flag for the number of volumes to a pod.
- Adding a failover test at scale:
1. Creating the maximum replicas on the first half nodes by cordoning the second half nodes.
2. After all pods are running, cordoning the first half nodes, uncordoning the second half nodes, and wait for all replica attachments are attached.
3. Deleting all the running pods and measuring the failover latency.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/test/scale/azdisk_scale_test.go
+++ b/test/scale/azdisk_scale_test.go
@@ -17,10 +17,13 @@ limitations under the License.
 package scale
 
 import (
+	"fmt"
+
 	"github.com/onsi/ginkgo"
 	v1 "k8s.io/api/core/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
+	azdisk "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/client/clientset/versioned"
 	consts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 	testconsts "sigs.k8s.io/azuredisk-csi-driver/test/const"
 	"sigs.k8s.io/azuredisk-csi-driver/test/e2e/driver"
@@ -55,7 +58,7 @@ func scaleTests(isMultiZone bool) {
 	ginkgo.It("Scale test scheduling and starting multiple pods with a persistent volume.", func() {
 		testutil.SkipIfUsingInTreeVolumePlugin()
 		volumes := []resources.VolumeDetails{}
-		for j := 1; j <= 1; j++ {
+		for j := 1; j <= *volMountedToPod; j++ {
 			volume := resources.VolumeDetails{
 				FSType:    "ext4",
 				ClaimSize: "256Gi",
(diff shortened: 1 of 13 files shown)


test/scale/azdisk_scale_test.go
```diff
@@ -151,12 +118,10 @@ func scaleTests(isMultiZone bool) {
 		test.Run(cs, ns, schedulerName)
 	})
 
-	ginkgo.It("Scale test scheduling and starting multiple pods with multiple persistent volume requests with replicas and reschedule on deletion.", func() {
+	ginkgo.It("Scale test scheduling and starting multiple pods with one persistent volume request and reschedule on deletion.", func() {
```
landreasyan, line 121: np: let's not specify `one` as we can now configure that number with a flag.

test/scale/scale_tester.go
```diff
@@ -178,94 +186,115 @@ scaleInTest:
...
+	totalNumberOfReplicaAtts := totalNumberOfPods * (maxShares - 1)
 
-		go func(ss *resources.TestStatefulset) {
-			defer wg.Done()
-			defer ginkgo.GinkgoRecover()
+waitForReplicaAtts:
+	for {
+		select {
+		case <-timeout:
+			break waitForReplicaAtts
```
landreasyan, line 256: Let's add a log line to print the number of replicas when the test times out to give us some idea of how much progress it was able to make

hccheng72: Thanks for the reminder. I also add that if `numberOfAttachedReplicaAtts != totalNumberOfReplicaAtts` will fail the test.

edreed:
Holding for additional changes. Will clear on next push.

/hold

test/scale/scale_tester.go
```diff
@@ -178,94 +186,117 @@ scaleInTest:
...
+	// start scale out test
+	startScaleOut := time.Now()
 	tStorageClass, storageCleanup := t.Pod.CreateStorageClass(client, namespace, t.CSIDriver, t.StorageClassParameters)
 	defer storageCleanup()
+	statefulSet, cleanupStatefulSet := t.Pod.SetupStatefulset(client, namespace, t.CSIDriver, schedulerName, t.Replicas, &tStorageClass, map[string]string{"group": "azuredisk-scale-tester"})
+	defer cleanupStatefulSet(45 * time.Minute)
+
+	statefulSet.CreateWithoutWaiting()
+	ticker := time.NewTicker(time.Minute)
+	numberOfRunningPods1 := podutil.WaitForPodsWithMatchingLabelToRun(*scaleTestTimeout, ticker, totalNumberOfPods, client, namespace)
```
landreasyan, line 222: You could initialize the `ticker` in the `WaitForPodsWithMatchingLabelToRun` instead of passing it to the function.

test/scale/scale_tester.go
```diff
@@ -178,94 +186,117 @@ scaleInTest:
...
+	testPod := resources.TestPod{
+		Client: client,
+	}
 
-	ginkgo.By("Initiating statefulset setup")
+	// uncordon the first half nodes
+	for _, node := range nodes[:midNode] {
+		node := node
+		framework.Logf("Uncordoning %s.", node)
+		testPod.SetNodeUnschedulable(node, false)
```
landreasyan, line 203: Instead of having `SetNodeUnschedulable` a menthod of `testPod` let's make it a function in `test/resources/utils.go` and pass the `client` as a parameter.

test/scale/scale_tester.go
```diff
@@ -178,94 +186,117 @@ scaleInTest:
...
+
+	// cordon the second half nodes)
+	for _, node := range nodes[midNode:] {
+		node := node
+		framework.Logf("Cordoning %s.", node)
+		testPod.SetNodeUnschedulable(node, true)
+	}
 
+	// start scale out test
+	startScaleOut := time.Now()
```
landreasyan, line 214: This is wrong in the other scale test as well, but I think we should start measuring the scale out right before `statefulSet.CreateWithoutWaiting()`

test/scale/scale_tester.go
```diff
@@ -178,94 +186,117 @@ scaleInTest:
...
+	framework.ExpectNoError(err)
+	labelSelector := labels.NewSelector().Add(*roleReq)
 
-			statefulset.Create()
-		}(statefulSets[i])
-	}
+	ticker.Reset(time.Minute)
+	tickerCount := 0
+	timeout := time.After(time.Duration(*scaleTestTimeout) * time.Minute)
+	startCreateReplicaAtts := time.Now()
```
landreasyan, line 246: We should be measuring the replica creation right after we uncordon the second group of nodes, around line 238

test/scale/scale_tester.go
```diff
@@ -178,94 +186,117 @@ scaleInTest:
...
-			framework.ExpectNoError(err)
-			time.Sleep(30 * time.Second)
-			err = ss.WaitForPodReadyOrFail()
-			framework.ExpectNoError(err)
-		}(statefulSets[i])
+	// start failover test
+	ticker.Reset(time.Minute)
+	startFailoverScale := time.Now()
+	numberOfRunningPods2 := podutil.WaitForPodsWithMatchingLabelToRun(*scaleTestTimeout, ticker, totalNumberOfPods, client, namespace)
+	failoverScaleCompleted := time.Since(startFailoverScale).Minutes()
```
landreasyan, line 285: We should start measuring the time right after delete on line 280.

test/scale/scale_tester.go
hccheng72: Yeah, it makes more sense to move the function to `node_utils.go`. I'll get rid of the existing one.

test/scale/scale_tester.go
hccheng72: Make sense, will change the other sale test together.

Issue #1492, opened 2022-08-30
[V2] azuredisk_csi_driver_controller_publish_volume latency metrics is incorrect in v2 driver


andyzhangx:
**What happened**:
In `ControllerPublishVolume` func, disk attach is async call, so `azuredisk_csi_driver_controller_publish_volume` latency metrics is incorrect in v2 driver now, need to get actuall disk attach latency, same issue for `azuredisk_csi_driver_controller_unpublish_volume` latency metrics

```
azure_metrics.go:114] "Observed Request Latency" latency_seconds=0.000512805 request="azuredisk_csi_driver_controller_publish_volume" resource_group="kubetest-rfl35lck" subscription_id="0e46bd28-a80f-4d3a-8200-d9eb8d80cb2e" source="disk.csi.azure.com" volumeid="/subscriptions/0e46bd28-a80f-4d3a-8200-d9eb8d80cb2e/resourceGroups/kubetest-rfl35lck/providers/Microsoft.Compute/disks/pvc-065ce6f7-649d-4bd2-af11-c15b48e8e556" node="k8s-agentpool-18545342-vmss000001" result_code="succeeded"
```

cc @edreed 

**What you expected to happen**:

**How to reproduce it**:

**Anything else we need to know?**:

**Environment**:

- CSI Driver version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

landreasyan:
The issue is fixed by #1531. New `attach_volume_latency` is added to track the latency of the actual attach operation.

PR #1531, opened 2022-09-28
[V2] feat: added attach_volume_latency metric for controllerserver

shlokshah-dev:
**What type of PR is this?**
/kind test

Added `attach_volume_latency` metric for the controllerserver

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/controllerserver_v2.go
+++ b/pkg/azuredisk/controllerserver_v2.go
@@ -299,17 +299,7 @@ func (d *DriverV2) ControllerUnpublishVolume(ctx context.Context, req *csi.Contr
 		mc.ObserveOperationWithResult(isOperationSucceeded, consts.VolumeID, diskURI, consts.Node, string(nodeName))
 	}()
 
-	// attachMc := metrics.NewMetricContext(d.cloudProvisioner.GetMetricPrefix(), "detach_volume_latency", d.cloudProvisioner.GetCloud().ResourceGroup, d.cloudProvisioner.GetCloud().SubscriptionID, d.Name)
-	// isDetachSuccessful := false
-
 	err := d.crdProvisioner.UnpublishVolume(ctx, diskURI, nodeID, req.GetSecrets(), consts.DemoteOrDetach)
-
-	// go func() {
-	// 	if err := d.crdProvisioner.WaitForDetach(ctx, diskURI, nodeID); err == nil {
-	// 		isDetachSuccessful = true
-	// 		mc.ObserveOperationWithResult(isDetachSuccessful, consts.VolumeID, diskURI, consts.Node, string(nodeName))
-	// 	}
-	// }
 
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


pkg/azuredisk/controllerserver_v2.go
```diff
@@ -246,8 +247,26 @@ func (d *DriverV2) ControllerPublishVolume(ctx context.Context, req *csi.Control
 
 	volumeCapability := generateAzVolumeCapability(volCap)
 
+	attachMc := metrics.NewMetricContext(d.cloudProvisioner.GetMetricPrefix(), "attach_volume_latency", d.cloudProvisioner.GetCloud().ResourceGroup, d.cloudProvisioner.GetCloud().SubscriptionID, d.Name)
+	isAttachSuccessful := false
+
 	response, err := d.crdProvisioner.PublishVolume(ctx, diskURI, nodeID, &volumeCapability, req.GetReadonly(), req.GetSecrets(), req.GetVolumeContext())
 
+	// Go Routine to calculate the attach_volume_latency
+	if err == nil {
```
edreed, line 256: The dispatch of the goroutine can be moved after the error check and return statement below. This would avoid the redundant check of `err` and the possibility that someone inserts code between the goroutine dispatch and real error return check.

edreed:
@shlokshah-dev Can you please squash these commits into 1? Something like, `git rebase -i  HEAD^^^^^^^`. The merge commit may give you some issues, though. For the future, it's generally better to rebase on main_v2 than merge to pull in updates and resolve conflicts..

pkg/azuredisk/controllerserver_v2.go
shlokshah-dev: Done.

shlokshah-dev:
@edreed The different commits have been squashed to 1 commit as suggested.

pkg/azuredisk/controllerserver_v2.go
```diff
@@ -282,8 +299,18 @@ func (d *DriverV2) ControllerUnpublishVolume(ctx context.Context, req *csi.Contr
 		mc.ObserveOperationWithResult(isOperationSucceeded, consts.VolumeID, diskURI, consts.Node, string(nodeName))
 	}()
 
+	// attachMc := metrics.NewMetricContext(d.cloudProvisioner.GetMetricPrefix(), "detach_volume_latency", d.cloudProvisioner.GetCloud().ResourceGroup, d.cloudProvisioner.GetCloud().SubscriptionID, d.Name)
```
sunpa93, line 302: If we aren't going to be using `detach_volume_latency`, let's get rid of the commented lines. If we plan to use it but are deferring for futuer PR, let's add `TODO` with a short description in the comment area.

shlokshah-dev: I think it would be best for now to remove the commented lines of code. As discussed, the existing `controller_unpublish_volume` metric covers the detach metric latency as well.

sunpa93 (approved):
/lgtm

landreasyan:
This PR fixes #1492.

PR #1532, opened 2022-09-28
[V2] bug: WaitForAttach call should also be able to detach replicas for primary

sunpa93:
**What type of PR is this?**
/kind bug

* Currently, only CrdProvisioner.PublishVolume is capable of detaching replicas to make resources available for primary attachment.
* However, due to async attach behavior and possible races between CrdProvisioner and the CRD controllers, CrdProvisioner.WaitForAttach also needs to be able to detach replicas for primary.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/provisioner/crdprovisioner.go
+++ b/pkg/provisioner/crdprovisioner.go
@@ -23,6 +23,7 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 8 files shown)


edreed:
Hold pending cut of the next release.

/hold

pkg/provisioner/crdprovisioner.go
```diff
@@ -637,9 +570,10 @@ func (c *CrdProvisioner) PublishVolume(
...
-	for i := 0; requiredUnpublishCount > 0 && i < len(unpublishOrder); i++ {
-		if err = c.detachVolume(ctx, unpublishOrder[i]); err != nil {
-			err = status.Errorf(codes.Internal, "failed to make request to unpublish volume (%s) from node (%s): %v", unpublishOrder[i].Spec.VolumeName, unpublishOrder[i].Spec.NodeName, err)
+	for _, replicaToDetach := range replicasToDetach {
+		w.Logger().V(5).Infof("Detaching replica attachment (%s) to make resource for primary attachment (%s)", replicaToDetach.Name, attachmentName)
+		if err = c.detachVolume(ctx, replicaToDetach); err != nil {
+			err = status.Errorf(codes.Internal, "failed to make request to unpublish volume (%s) from node (%s): %v", replicaToDetach.Spec.VolumeName, replicaToDetach.Spec.NodeName, err)
 			return nil, err
 		}
 		requiredUnpublishCount--
```
hccheng72, line 579: Is variable `requiredUnpublishCount` still used in this function?

edreed:
@sunpa93 Is this change ready to go?

pkg/provisioner/crdprovisioner.go
```diff
@@ -641,12 +573,12 @@ func (c *CrdProvisioner) PublishVolume(
...
 	}
 
-	for i := 0; requiredUnpublishCount > 0 && i < len(unpublishOrder); i++ {
-		if err = c.detachVolume(ctx, unpublishOrder[i]); err != nil {
-			err = status.Errorf(codes.Internal, "failed to make request to unpublish volume (%s) from node (%s): %v", unpublishOrder[i].Spec.VolumeName, unpublishOrder[i].Spec.NodeName, err)
+	for _, replicaToDetach := range replicasToDetach {
+		w.Logger().V(5).Infof("Detaching replica attachment (%s) to make resource for primary attachment (%s)", replicaToDetach.Name, attachmentName)
+		if err = c.detachVolume(ctx, replicaToDetach); err != nil {
+			err = status.Errorf(codes.Internal, "failed to make request to unpublish volume (%s) from node (%s): %v", replicaToDetach.Spec.VolumeName, replicaToDetach.Spec.NodeName, err)
 			return nil, err
```
edreed, line 580: Shouldn't we try to detach the next replica instead of just failing? The `detachVolume` function synchronously detaches the volume. Can there ever be more than one replica being detached? (`getReplicasToDetach` returns an array, so I assume yes.) If so, it would be good to do the detaches in parallel and wait for completion of all. This would reduce publish latency and increase the likelihood of the detaches happening in a single batch since they should be on the same node.

pkg/provisioner/crdprovisioner.go
```diff
@@ -737,11 +669,54 @@ func (c *CrdProvisioner) waitForLun(ctx context.Context, volumeID, nodeID string
...
+
+	var replicasToDetach []*azdiskv1beta2.AzVolumeAttachment
+	if replicasToDetach, err = c.getReplicasToDetach(ctx, volumeName, nodeID); err != nil {
+		// log errors and continue
+		w.Logger().Errorf(err, "Failed to get replica attachments to detach.")
+	}
+
+	for _, replicaToDetach := range replicasToDetach {
+		w.Logger().V(5).Infof("Detaching replica attachment (%s) to make resource for primary attachment (%s)", replicaToDetach.Name, azureutils.GetAzVolumeAttachmentName(volumeName, nodeID))
+		if err = c.detachVolume(ctx, replicaToDetach); err != nil {
```
edreed, line 714: The `detachVolume` function calls `WaitForAttach`. Since the former is called in order to detach replicas to make room for primaries, we'll end up recursively trying to detach replicas in this case won't we?

pkg/provisioner/crdprovisioner.go
```diff
@@ -1124,6 +1099,81 @@ func isAzVolumeSpecSameAsRequestParams(defaultAzVolume *azdiskv1beta2.AzVolume,
 		reflect.DeepEqual(defaultAccReq, accessibilityReq))
 }
 
+func (c *CrdProvisioner) getReplicasToDetach(ctx context.Context, volumeName, nodeID string) (unpublishOrder []*azdiskv1beta2.AzVolumeAttachment, err error) {
+	filterVAToDetach := func(attachments []azdiskv1beta2.AzVolumeAttachment) (numAttachment int, unpublishOrder []*azdiskv1beta2.AzVolumeAttachment) {
```
edreed, line 1103: Consider making this a standalone function rather than an inline function.

pkg/provisioner/crdprovisioner.go
```diff
@@ -1200,6 +1250,17 @@ func waitForAttachVolumeFunc(obj interface{}, objectDeleted bool) (bool, error)
 	return false, nil
 }
 
+func waitForAttachVolumeCompleteFunc(obj interface{}, objectDeleted bool) (bool, error) {
```
landreasyan, line 1253: np: since this method is specific to Primary attachments, only returns true for success and doesn't do the waiting maybe the name could be something like `isPrimaryAttachmentSuccessful` or `isPrimaryAttachSuccessful`.

pkg/provisioner/crdprovisioner.go
```diff
@@ -737,11 +669,54 @@ func (c *CrdProvisioner) waitForLun(ctx context.Context, volumeID, nodeID string
...
+	defer waiter.Close()
+
+	_, err = waiter.Wait(ctx)
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
 func (c *CrdProvisioner) WaitForAttach(ctx context.Context, volumeID, nodeID string) (*azdiskv1beta2.AzVolumeAttachment, error) {
```
landreasyan, line 695: this method no longer only waits for attach and needs to be renamed.

pkg/provisioner/crdprovisioner.go
```diff
@@ -737,11 +669,54 @@ func (c *CrdProvisioner) waitForLun(ctx context.Context, volumeID, nodeID string
...
 	defer func() { w.Finish(err) }()
 
+	volumeName, err := azureutils.GetDiskName(volumeID)
+	if err != nil {
+		return nil, err
+	}
+	volumeName = strings.ToLower(volumeName)
+
+	var replicasToDetach []*azdiskv1beta2.AzVolumeAttachment
+	if replicasToDetach, err = c.getReplicasToDetach(ctx, volumeName, nodeID); err != nil {
```
landreasyan, line 707: Hmm... Do we want to do all this work every time we wait for lun or attachment to complete?

hccheng72:
Can we move creating attachmentObj (line 563) right after attachmentObj isn't found? So if the creation fails, the function can return early and save `getReplicasToDetach` process.

sunpa93:
> Can we move [creating attachmentObj](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/354043cc23ab44baa242a9af406933f058ebd0db/pkg/provisioner/crdprovisioner.go#L563) right after attachmentObj isn't found? So if the creation fails, the function can return early and save `getReplicasToDetach` process.

Current `AzVolumeAttachment` CRI creation workflow has two possible scenarios: 1) when there are replicas that need to be detached and 2) when there isn't. When no replica needs to be detached, the workflow is rather simple, we create an `AzVolumeAttachment` CRI with `VolumeAttachmentRequestAnnotation`, and wait for the disk lun value to be assigned. However, when there are replicas that need to be detached (i.e. maxShares saturated, or max disk attached saturated), we need to issue a Create call for `AzVolumeAttachment` CRI without a `VolumeAttachRequestAnnotation` to prevent race between the attach operation of the new primary and the detach operation of existing replicas. This is why we, at the moment, need to check for `replicasToDetach` prior to creating the CRI.

As for:

> Same idea for:
> 
> https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/354043cc23ab44baa242a9af406933f058ebd0db/pkg/provisioner/crdprovisioner.go#L587

Yes, I think moving this to before calling for `getReplicasToDetach` makes sense and be beneficial.

For:
> https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/354043cc23ab44baa242a9af406933f058ebd0db/pkg/provisioner/crdprovisioner.go#L623

The error handling we need may exactly be the detachment of the replicas, so it would make sense for us to call `getReplicasToDetach` prior to checking the error.

edreed:

PR #1542, opened 2022-10-07
[V2] feat: add preferred affinity logic and unit tests for prioritize and filter logic

### LINKED ISSUE #1423: [V2] shared disk conflict with NodeAffinity and PodAffinity
**What happened**:
1. V2 driver scheduler extender would set different score on node if shared disk has already been attached to the node, thus it would has conflict with preferred node affinity, preferred pod affinity at least: https://kubernetes.io/docs/tasks/configure-pod-container/assign-pods-nodes-using-node-affinity/#schedule-a-pod-using-preferred-node-affinity

e.g. if shared disk is already attached to one node, and that node is not preferred node affinity, will it scheduled or not scheduled on to that node? what's the node score in such case?

2. what about the required node affinity case? e.g.
e.g. if shared disk is already attached to one node, and that node is not in the node affinity, will it scheduled or not scheduled on to that node? what's the node score in such case? I suppose in that case, pod should NOT be scheduled to that node

**How to reproduce it**:

**Environment**:
- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sunpa93:
**What type of PR is this?**
/kind feature

* This PR adds preferred affinity score plugins to replica attachment node selection logic along with fixing / refactoring existing plugin codes.

Fixes #1423

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


pkg/controller/common.go
```diff
@@ -1010,3 +1099,32 @@ func verifyObjectFailedOrDeleted(obj interface{}, objectDeleted bool) (bool, err
 	}
 	return false, nil
 }
+
+func WatchObject(mgr manager.Manager, objKind source.Kind) error {
```
edreed, line 1103: It would be good to document the use case for this function.

pkg/controller/common.go
```diff
@@ -259,138 +268,56 @@ func (p *interPodAffinityFilter) filter(ctx context.Context, nodes []v1.Node) ([
...
+	// Logging purpose
+	evictedNodes := make([]string, len(nodes)-len(filteredNodes))
+	i := 0
+	for _, node := range nodes {
+		if !qualifyingNodes.has(node.Name) {
+			evictedNodes[i] = node.Name
+			i++
+		}
+	}
+	w.Logger().V(5).Infof("nodes (%+v) filtered out by %s", evictedNodes, p.name())
```
edreed, line 319: I wonder if we should use `V(10)` here since this would emit quick a lot of data. Using a quick search of the vendor sources, it seems this verbosity level is for emitting this level of detail.

pkg/controller/common.go
```diff
@@ -419,95 +346,49 @@ func (p *interPodAntiAffinityFilter) filter(ctx context.Context, nodes []v1.Node
...
+	// Logging purpose
+	evictedNodes := make([]string, len(nodes)-len(filteredNodes))
+	i := 0
+	for _, node := range nodes {
+		if qualifyingNodes.has(node.Name) {
+			evictedNodes[i] = node.Name
+			i++
 		}
 	}
+	w.Logger().V(5).Infof("nodes (%+v) filtered out by %s", evictedNodes, p.name())
```
edreed, line 390: `V(10)`?

pkg/controller/common.go
```diff
@@ -741,30 +624,44 @@ func (s *scoreByNodeCapacity) name() string {
 	return "score by node capacity"
 }
 
-func (s *scoreByNodeCapacity) setup(pods []v1.Pod, volumes []string, state *SharedState) {
+func (s *scoreByNodeCapacity) priority() float64 {
+	return 1
+}
+
+func (s *scoreByNodeCapacity) setup(nodes []v1.Node, pods []v1.Pod, volumes []string, state *SharedState) {
+	s.nodes = nodes
```
edreed, line 632: It seems like we're only using the node name. Could we simply store the names instead of a full copy of the pods to reduce the memory usage?

pkg/controller/common.go
```diff
@@ -741,30 +624,44 @@ func (s *scoreByNodeCapacity) name() string {
...
-		nodeScores[nodeName] = score + (nodeScoreLowCoefficient * remainingCapacity)
+		nodeScores[node.Name] += int((float64(remainingCapacity) / float64(maxCapacity)) * math.Pow(10, s.priority()))
 
 		if remainingCapacity-len(s.volumes) < 0 {
-			delete(nodeScores, nodeName)
+			delete(nodeScores, node.Name)
 		}
 
-		w.Logger().V(5).Infof("node (%s) can accept %d more attachments", nodeName, remainingCapacity)
+		w.Logger().V(5).Infof("node (%s) can accept %d more attachments", node.Name, remainingCapacity)
```
edreed, line 664: `V(10)`?

pkg/controller/common.go
```diff
@@ -778,34 +675,226 @@ func (s *scoreByReplicaCount) name() string {
...
 	defer w.Finish(nil)
+
+	var requestedReplicaCount int
+
+	nodeReplicaCounts := map[string]int{}
+
 	for _, volume := range s.volumes {
+		azVolume, err := azureutils.GetAzVolume(ctx, s.state.cachedClient, nil, volume, s.state.objectNamespace, true)
+		if err != nil {
+			w.Logger().V(5).Errorf(err, "failed to get AzVolume (%s)", volume)
```
edreed, line 698: Errors should probably be logged at the base level rather than as verbose since they would have externally visible effects.

pkg/controller/shared_state.go
```diff
@@ -649,7 +649,7 @@ func (c *SharedState) filterNodes(ctx context.Context, nodes []v1.Node, pods []v
 			for i, filteredNode := range filteredNodes {
 				nodeStrs[i] = filteredNode.Name
 			}
-			w.Logger().V(2).Infof("Filtered node list from filter plugin (%s): %+v", filterPlugin.name(), nodeStrs)
+			w.Logger().Infof("Filtered node list from filter plugin (%s): %+v", filterPlugin.name(), nodeStrs)
```
edreed, line 652: `V(10)`?

pkg/controller/shared_state.go
```diff
@@ -668,16 +668,24 @@ func (c *SharedState) prioritizeNodes(ctx context.Context, pods []v1.Pod, volume
...
 			w.Logger().Errorf(err, "failed to score nodes by node scorer (%s)", nodeScorerPlugin.name())
 		} else {
 			// update node scores if scorer plugin returned success
 			nodeScores = updatedNodeScores
 		}
+		var nodeScoreResult string
+		for nodeName, score := range nodeScores {
+			nodeScoreResult += fmt.Sprintf("<%s: %d> ", nodeName, score)
+		}
+		w.Logger().Infof("node score after node score plugin (%s): %s", nodeScorerPlugin.name(), nodeScoreResult)
```
edreed, line 688: `V(10)`?

pkg/controller/common.go
sunpa93: Pod objects are used in other score plugins as `scoreByInterPodAffinity`
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/6120a76ee821b08db119deab10b8d155d4f415ff/pkg/controller/common.go#L759-L778
And node objects are used in `scoreByPodNodeAffinity`
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/6120a76ee821b08db119deab10b8d155d4f415ff/pkg/controller/common.go#L888-L893

PR #1544, opened 2022-10-09
[V2] chore: merge master as of 2022-10-07

edreed:
**What type of PR is this?**

/kind cleanup

Merges the master branch into main_v2 with changes as of 2022-10-07.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


test/resources/deployment.go
```diff
@@ -188,28 +188,20 @@ func (t *TestDeployment) PollForStringInPodsExec(command []string, expectedStrin
 func (t *TestDeployment) DeletePodAndWait() {
 	ch := make(chan error, len(t.Pods))
 	for _, pod := range t.Pods {
-		e2elog.Logf("Deleting pod %q in namespace %q", pod.Name, t.Namespace.Name)
 		go func(client clientset.Interface, ns, podName string) {
-			err := client.CoreV1().Pods(ns).Delete(context.TODO(), podName, metav1.DeleteOptions{})
+			err := e2epod.DeletePodWithWaitByName(client, podName, ns)
```
edreed, line 192: @landreasyan , @sunpa93 Was there a reason not to use `DeletePodWithWaitByName` here? Using `WaitForPodNoLongerRunningInNamespace` failed because sometimes the pod would be deleted before the final state could be observed. The test would fail in this case.

sunpa93: I don't think there is any reason not to use `DeletePodWithWaitByName` here. The method also specifies that it is `Resilient to the pod not existing` case.

PR #1550, opened 2022-10-14
[V2] bug: fix in conversion for scheduler extender latency log

landreasyan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azdiskschedulerextender/azdiskschedulerextender_metrics.go
+++ b/pkg/azdiskschedulerextender/azdiskschedulerextender_metrics.go
@@ -37,7 +37,7 @@ var (
 			Namespace:      scheduler,
 			Name:           "extender_filter_nodes_duration_seconds",
 			Help:           "Latency for handling filter nodes request.",
-			Buckets:        metrics.ExponentialBuckets(0.0001, 2, 12),
+			Buckets:        metrics.ExponentialBuckets(0.00001, 2, 12),
 			StabilityLevel: metrics.ALPHA,
 		},
 		[]string{"filter"})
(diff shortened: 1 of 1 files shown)


pkg/azdiskschedulerextender/azdiskschedulerextender_metrics.go
```diff
@@ -50,7 +50,7 @@ var (
 			Buckets:        metrics.ExponentialBuckets(0.0001, 2, 12),
```
edreed, line 50: Given the latencies we typically see are <20us, I wonder if we should adjust the buckets to have better granularity at these times, i.e. start at 0.00001.

pkg/azdiskschedulerextender/azdiskschedulerextender_metrics.go
```diff
@@ -84,6 +84,7 @@ func RegisterMetrics(schedulerMetrics ...metrics.Registerable) {
 
 func DurationSince(functionName string, start time.Time) float64 {
 	requestDuration := time.Since(start).Seconds()
-	klog.V(2).Infof("Call to %v took: %f ms\n", functionName, requestDuration/1000)
+	keysAndValues := []interface{}{"latency_seconds", requestDuration, "request", functionName}
+	klog.V(2).InfoS("Observed Request Latency", keysAndValues...)
```
edreed, line 88: You can also pass the key values directly instead of constructing an array first, e.g.
```go
klog.V(2).InfoS("Observed Request Latency", "latency_seconds", requestDuration, "request", functionName)
```

landreasyan: Yeah, I know but I thought this made the grouping clearer and the lines shorter. Should I update it?

edreed: I don't have a strong opinion one way or the other. Line length can be mitigated by breaking the statement across multiple lines.

PR #1553, opened 2022-10-18
[V2] feat: backload API DELETE call to deload api server calls upon detachment

sunpa93:
**What type of PR is this?**
/kind feature

* This backloads API DELETE calls on AzVolumeAttachment objects to mitigate API server throttling.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/attach_detach.go
+++ b/pkg/controller/attach_detach.go
@@ -108,8 +108,13 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
 		return reconcileReturnOnSuccess(azVolumeAttachment.Name, r.retryInfo)
 	}
 
-	// detachment request
+	// deletion request
 	if objectDeletionRequested(azVolumeAttachment) {
+		if err := r.removeFinalizer(ctx, azVolumeAttachment); err != nil {
+			return reconcileReturnOnError(ctx, azVolumeAttachment, "delete", err, r.retryInfo)
+		}
+		// detachment request
+	} else if volumeDetachRequested(azVolumeAttachment) {
 		if err := r.triggerDetach(ctx, azVolumeAttachment); err != nil {
 			return reconcileReturnOnError(ctx, azVolumeAttachment, "detach", err, r.retryInfo)
 		}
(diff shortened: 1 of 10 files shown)


pkg/controller/replica.go
```diff
@@ -67,11 +67,9 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
 
 	if azVolumeAttachment.Spec.RequestedRole == azdiskv1beta2.PrimaryRole {
 		// Deletion Event
```
edreed, line 69: `// Detach Event`

pkg/controller/shared_state.go
```diff
@@ -1003,7 +1003,7 @@ func (c *SharedState) cleanUpAzVolumeAttachmentByNode(ctx context.Context, azDri
 
 func (c *SharedState) cleanUpAzVolumeAttachments(ctx context.Context, attachments []azdiskv1beta2.AzVolumeAttachment, cleanUp cleanUpMode, caller operationRequester) error {
 	var err error
-	w, _ := workflow.GetWorkflowFromContext(ctx)
+	_, _ = workflow.GetWorkflowFromContext(ctx)
```
edreed, line 1006: Do we even need to do this at all bow?

sunpa93: No, :D

pkg/provisioner/crdprovisioner.go
```diff
@@ -546,7 +546,7 @@ func (c *CrdProvisioner) PublishVolume(
 	filterVAToDetach := func(attachments []azdiskv1beta2.AzVolumeAttachment) (numAttachment int, unpublishOrder []*azdiskv1beta2.AzVolumeAttachment) {
 		for i := range attachments {
 			// only count if deletionTimestamp not set
-			if attachments[i].GetDeletionTimestamp().IsZero() {
+			if attachments[i].GetDeletionTimestamp().IsZero() && !azureutils.MapContains(attachments[i].Status.Annotations, consts.VolumeDetachRequestAnnotation) {
```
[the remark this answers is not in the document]
sunpa93: `volumeDetachRequested` should only check for `VolumeDetachRequestAnnotation` because we now would be removing finalizer from the object when `deletionTimestamp` is set.

pkg/provisioner/crdprovisioner.go
```diff
@@ -648,7 +648,7 @@ func (c *CrdProvisioner) PublishVolume(
 	var updateFunc func(obj client.Object) error
 	updateMode := azureutils.UpdateCRIStatus
 	// if attachment is scheduled for deletion, new attachment should only be created after the deletion
-	if attachmentObj.DeletionTimestamp != nil {
+	if attachmentObj.DeletionTimestamp != nil || azureutils.MapContains(attachmentObj.Status.Annotations, consts.VolumeDetachRequestAnnotation) {
```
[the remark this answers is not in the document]
sunpa93: This clause is to check whether there already exists an `AzVolumeAttachment` and if there is, whether that `AzVolumeAttachment` is requested for detach or deletion at the moment. So, I don't think we would need a check for whether `DeletionTimestamp.IsZero` or not.

pkg/controller/attach_detach.go
```diff
@@ -108,8 +108,14 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
...
 
-	// detachment request
+	// deletion request
 	if objectDeletionRequested(azVolumeAttachment) {
+		if err := r.removeFinalizer(ctx, azVolumeAttachment); err != nil {
+			return reconcileReturnOnError(ctx, azVolumeAttachment, "delete", err, r.retryInfo)
+		}
+	}
+	// detachment request
+	if volumeDetachRequested(azVolumeAttachment) {
```
edreed, line 118: This should be an `else if`. Since the object deletion happens *after* the detach completes, we don't need to process a second detach for the `AzVolumeAttachment`.

PR #1554, opened 2022-10-18
[V2] test: make maxShares configurable to scale test

hccheng72:
**What type of PR is this?**
/kind test

Making maxShares configurable to the existing scale test such that we can opt to run a scale test with replica attachments.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


test/scale/azdisk_scale_test.go
```diff
@@ -55,7 +55,7 @@ func scaleTests(isMultiZone bool) {
 		testDriver = driver.InitAzureDiskDriver()
 	})
 
-	ginkgo.It("Scale test scheduling and starting multiple pods with a persistent volume.", func() {
+	ginkgo.It("Scale test scheduling and starting multiple pods with persistent volume(s).", func() {
```
alice-zheyan-yu, line 58: Should we also update the scale test name in Makefile: https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/main_v2/Makefile#L354 ?

hccheng72: Yes, thanks for pointing out.

test/scale/scale_tester.go
```diff
@@ -117,68 +134,48 @@ scaleOutTest:
...
-			} else if tickerCount == 5 {
-				fiveMinPodsDeleted = totalNumberOfPods - numberOfRemainingPods
-			} else if tickerCount == 10 {
-				tenMinPodsDeleted = totalNumberOfPods - numberOfRemainingPods
-			}
-		}
+	numberOfRemainingReplicaAtts := 0
+	scaleInAttsCompleted := float64(0)
+	if testWithReplicas {
+		time.Sleep(controller.DefaultTimeUntilGarbageCollection)
```
landreasyan, line 149: Would it maybe be better to call this sleep in a goroutine after all pods are deleted and then wait for its completion here? That way we will avoid sleeping for more than necessary because I think `WaitForAllVolumeAttachmentsToDetach` might take 2-3 mins or so.

test/scale/scale_tester.go
```diff
@@ -117,68 +134,48 @@ scaleOutTest:
...
-		select {
-		case <-timeout:
-			break scaleInTest
-		case <-ticker.C:
-			tickerCount = tickerCount + 1
-			numberOfRemainingPods = podutil.CountAllPodsWithMatchingLabel(client, namespace, map[string]string{"group": "azuredisk-scale-tester"}, string(v1.PodRunning))
-			if numberOfRemainingPods <= 0 {
-				break scaleInTest
-			}
+	numberOfRemainingVolumeAtts := volutil.WaitForAllVolumeAttachmentsToDetach(*scaleTestTimeout, 30*time.Second, totalNumberOfPods**volMountedToPod, client)
```
landreasyan, line 143: Since now with the replicas we won't actually detach, should we rename this to  `WaitForAllVolumeAttachmentsToDelete`?

test/scale/scale_tester.go
hccheng72: Thanks for catching that. I run `WaitForAllVolumeAttachmentsToDetach` in a goroutine instead to avoid it running longer than sleeping time and affecting accuracy of replicas detaching measurement.

test/scale/scale_tester.go
```diff
@@ -117,68 +134,56 @@ scaleOutTest:
...
-			}
-		}
+	var wg sync.WaitGroup
+	numberOfRemainingVolumeAtts := numberOfRunningPods * *volMountedToPod
+	scaleInVACompleted := float64(0)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		numberOfRemainingVolumeAtts = volutil.WaitForAllVolumeAttachmentsToDelete(*scaleTestTimeout, 30*time.Second, client)
+		scaleInVACompleted = time.Since(startScaleIn).Minutes()
```
landreasyan, line 150: np: we don't need to change this in this PR. Instead of waiting for done signal, should we return the timestamp in a channel and wait for that value? That way we wouldn't update a var from the outer scope in the goroutine.

landreasyan: Actually, the easier thing would be to call `time.Sleep(controller.DefaultTimeUntilGarbageCollection)` in the goroutine and wait for it before `WaitForAllReplicaAttachmentsToDetach`. We then do `WaitForAllVolumeAttachmentsToDelete` in the main thread while waiting for the goroutine.

landreasyan: As Huichan pointed out, the second solution won't be ideal because we have no guarantee that WaitForAllReplicaAttachmentsToDetach is always under 5 min.

hccheng72: As discussed with Lana, we use channel to return an integer and a timestamp inside a struct.

test/utils/pod/pod_utils.go
```diff
@@ -124,10 +126,33 @@ func WaitForPodsWithMatchingLabelToRun(testTimeout int, tickerDuration time.Dura
...
+	timeout := time.After(time.Duration(testTimeout) * time.Minute)
+
+	for {
+		select {
+		case <-timeout:
+			return
+		case <-ticker.C:
+			tickerCount++
+			numberOfRunningPods = CountAllPodsWithMatchingLabel(cs, ns, map[string]string{"group": "azuredisk-scale-tester"}, string(v1.PodRunning))
+			e2elog.Logf("%.1f min: %d pods are running", float64(tickerCount)*tickerDuration.Seconds()/60, numberOfRunningPods)
```
landreasyan, line 152: np: can we get the duration in minutes here directly?

hccheng72: Oh yes!

PR #1557, opened 2022-10-21
[V2] fix: replicas aren't getting cleaned up during driver uninstall

sunpa93:
**What type of PR is this?**

* Replica AzVolumeAttachments need to be detached and deleted during driver uninstall.
* This PR is necessary to unblock other tasks as 1) updating driver uninstall que and 2) adding allocation map.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/azvolume.go
+++ b/pkg/controller/azvolume.go
@@ -177,9 +177,8 @@ func (r *ReconcileAzVolume) triggerCreate(ctx context.Context, azVolume *azdiskv
 		if createErr != nil {
 			updateFunc = func(obj client.Object) error {
 				azv := obj.(*azdiskv1beta2.AzVolume)
-				azv = r.updateError(azv, createErr)
 				azv = r.deleteFinalizer(azv, map[string]bool{consts.AzVolumeFinalizer: true})
-				_, derr := r.updateState(azv, azdiskv1beta2.VolumeCreationFailed, forceUpdate)
+				_, derr := r.reportError(azv, azdiskv1beta2.VolumeCreationFailed, createErr)
 				return derr
 			}
 			updateMode = azureutils.UpdateAll
@@ -211,15 +210,6 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
 	ctx, w := workflow.New(ctx)
 	defer func() { w.Finish(err) }()
 
-	// Determine if this is a controller server requested deletion or driver clean up
-	volumeDeleteRequested := volumeDeleteRequested(azVolume)
-	preProvisionCleanupRequested := isPreProvisionCleanupRequested(azVolume)
-
-	mode := deleteCRIOnly
-	if volumeDeleteRequested || preProvisionCleanupRequested {
-		mode = detachAndDeleteCRI
-	}
-
 	// override volume operation queue to prevent any other replica operation from being executed
 	release := r.closeOperationQueue(azVolume.Name)
 	defer func() {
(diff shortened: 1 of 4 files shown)


pkg/controller/azvolume.go
```diff
@@ -247,71 +246,67 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
...
-				return derr
-			}
+		var updateFunc azureutils.UpdateCRIFunc
+		var err error
+		updateMode := azureutils.UpdateCRI
 
-			var updateFunc azureutils.UpdateCRIFunc
-			var err error
-			updateMode := azureutils.UpdateCRIStatus
+		reportError := func(obj interface{}, err error) error {
```
landreasyan, line 267: Updating AzVolume CRI with error with new error and then updating the state seems like a common operation. Why are we defining a new function for within a method rather than creating a separate function we can use more consistently.

pkg/controller/common.go
```diff
@@ -90,10 +90,10 @@ const (
 	pod              operationRequester = "pod-controller"
 )
 
-type cleanUpMode int
+type attachmentCleanUpMode int
```
landreasyan, line 93: Should we rename these in this PR instead?

pkg/controller/azvolume.go
```diff
@@ -215,9 +215,10 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
 	volumeDeleteRequested := volumeDeleteRequested(azVolume)
 	preProvisionCleanupRequested := isPreProvisionCleanupRequested(azVolume)
 
-	mode := deleteCRIOnly
+	// primary attachments should be detached only if volume is being deleted or pv was deleted.
+	mode := cleanUpAttachmentForUninstall
 	if volumeDeleteRequested || preProvisionCleanupRequested {
-		mode = detachAndDeleteCRI
+		mode = cleanUpAttachment
```
edreed, line 221: This if clause is now redundant to the one on line 232. This assignment can now be moved within the latter if clause.

sunpa93: True

pkg/controller/common.go
sunpa93: Sure

pkg/controller/azvolume.go
```diff
@@ -250,68 +245,59 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
...
+		deleteCtx, deleteCancel := context.WithTimeout(goCtx, cloudTimeout)
+		defer deleteCancel()
+
+		// Delete all AzVolumeAttachment objects bound to the deleted AzVolume
+		var attachments []azdiskv1beta2.AzVolumeAttachment
+		attachments, err = r.cleanUpAzVolumeAttachmentByVolume(deleteCtx, azVolume.Name, azvolume, azureutils.AllRoles, mode)
+		if err == nil {
+			var wg sync.WaitGroup
+			errors := make([]error, len(attachments))
+			numErrors := uint32(0)
```
landreasyan, line 274: I think it would be better to communicate with channels here. `errors` itself could be a buffered channel. We would then send all the errors thorough that channel and instead of wg.Wait () we can have 
```
go func() {
		wg.Wait()
		close(errors )
	}()
```

then we will loop over errors to check for error: 
```
for error := range errors {
		// update status
	}
```

pkg/controller/common.go
```diff
@@ -90,11 +90,11 @@ const (
...
 
 const (
```
landreasyan, line 95: np: I think it would be good to document the usage here.

pkg/controller/azvolume.go
sunpa93: Yeah, I see the benefit of using a buffered channel here. Let's merge this one and I will iterate on this in a separate PR as Huichan's change has dependency on this change.

pkg/controller/azvolume.go
```diff
@@ -250,68 +245,59 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
...
-			defer func() { goWorkflow.Finish(deleteErr) }()
-			waitCh <- goSignal{}
-
-			goCtx := goWorkflow.SaveToContext(context.Background())
-
-			deleteCtx, deleteCancel := context.WithTimeout(goCtx, cloudTimeout)
-			defer deleteCancel()
+	waitCh := make(chan goSignal)
+	//nolint:contextcheck // call is asynchronous; context is not inherited by design
+	go func() {
```
landreasyan, line 252: Do we need to do this in a goroutine?

sunpa93: I don't see a reason for why we shouldn't. Otherwise, we will have a reconciler worker blocked waiting for the attachments to be deleted.

landreasyan: I am not sure I understand what the waitChannel is used here for.

sunpa93: This is to ensure that the caller exits after goroutine child workflow is created. Otherwise, we end up in a situation where parent workflow completes before child workflow is created.

pkg/controller/azvolume.go
landreasyan: Sure!

pkg/controller/azvolume.go
landreasyan: Oh, I see it now!

PR #1562, opened 2022-10-27
[V2][WIP] Add node disk availability map

alice-zheyan-yu:
**What type of PR is this?**
A draft PR. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/controller/attach_detach.go
```diff
@@ -110,17 +111,52 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
...
+			r.logger.V(2).Info("aliceyu? delete: node: " + azVolumeAttachment.Spec.NodeName + " count: " + strconv.Itoa(remainingCapacity.(int)) + " vol: " + azVolumeAttachment.Spec.VolumeName + " role: " + string(azVolumeAttachment.Spec.RequestedRole))
+			r.SharedState.nodeDiskAvailabilityMap.Store(azVolumeAttachment.Spec.NodeName, remainingCapacity.(int)+1)
+		}
 		if err := r.removeFinalizer(ctx, azVolumeAttachment); err != nil {
+			// aliceyu? delete failed, -1
+			remainingCapacity, nodeExists := r.SharedState.nodeDiskAvailabilityMap.Load(azVolumeAttachment.Spec.NodeName)
+			if nodeExists {
+				r.logger.V(2).Info("aliceyu? delete: node: " + azVolumeAttachment.Spec.NodeName + " count: " + strconv.Itoa(remainingCapacity.(int)) + " vol: " + azVolumeAttachment.Spec.VolumeName + " role: " + string(azVolumeAttachment.Spec.RequestedRole))
+				r.SharedState.nodeDiskAvailabilityMap.Store(azVolumeAttachment.Spec.NodeName, remainingCapacity.(int)-1)
+			}
```
hccheng72, line 128: could we update the remaining capacity of the node only after removing finalizer successfully?

alice-zheyan-yu: I couldn't see much difference between my current logic and yours. Do you have some specific reasons for that?

hccheng72: Oh yeah, it should make no difference. Just to save one step to minus 1 back from the capacity if removing finalizer fails.

edreed:
Please rebase on main_v2 rather than merge updates to keep the commit history clean.

pkg/controller/attach_detach.go
```diff
@@ -110,7 +110,18 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
 
 	// deletion request
 	if objectDeletionRequested(azVolumeAttachment) {
+		r.SharedState.nodeDiskAvailabilityMapLock.Lock()
```
edreed, line 113: You can avoid the need for a global lock by using [`*atomic.Int32`](https://pkg.go.dev/sync/atomic#Int32) as the value in the `nodeDiskAvailabilityMap`. You need to store a pointer to the `atomic.Int32` in the `sync.Map` since you need to call methods on a shared value in order to update the count.

You can use [`atomic.Int32.Add(1)`](https://pkg.go.dev/sync/atomic@go1.19.3#Int32.Add) to increment the count. Decrementing the count is more complex since you need to use [`atomic.Int32.Load()`](https://pkg.go.dev/sync/atomic@go1.19.3#Int32.Load) to get the current value and then [`atomic.Int32.CompareAndSwap(oldValue, newValue int)`](https://pkg.go.dev/sync/atomic@go1.19.3#Int32.CompareAndSwap) to store the updated value if the current value is not zero. This needs to be done in a loop since another go routine may race to update the value, e.g.

```go
currentCapacity := 0
remainingCapacity, nodeExists := r.SharedState.nodeDiskAvailabilityMap.Load(azVolumeAttachment.Spec.NodeName)
if nodeExists && remainingCapacity != nil {
    for {
        currentCapacity = remainingCapacity.(*atomic.Int32).Load()
        if currentCapacity == 0 || remainingCapacity.(*atomic.Int32).CompareAndSwap(currentCapacity, currentCapacity - 1) {
            break
    }
}
```

pkg/controller/azdrivernode.go
```diff
@@ -83,6 +85,9 @@ func (r *ReconcileAzDriverNode) Reconcile(ctx context.Context, request reconcile
 
 // run an update on existing azdrivernode objects to store them under new version if necessary
 func (r *ReconcileAzDriverNode) Recover(ctx context.Context) error {
+	r.SharedState.nodeDiskAvailabilityMap.Range(func(k, v interface{}) bool {
+		return true
+	})
```
edreed, line 90: What's the purpose of this? It appears to be a no-op since the range function does nothing to the elements of the map.

pkg/controller/azdrivernode.go
```diff
@@ -124,9 +135,14 @@ func NewAzDriverNodeController(mgr manager.Manager, controllerSharedState *Share
...
 	}
 
-	// Predicate to only reconcile deleted nodes
+	// Predicate to only reconcile deleted nodes.
 	p := predicate.Funcs{
+		// For create node event, add the new node to nodeDiskAvailabilityMap
 		CreateFunc: func(e event.CreateEvent) bool {
+			diskCount, err := azureutils.GetNodeMaxDiskCountWithLabels(e.Object.GetLabels())
+			if err == nil {
+				reconciler.nodeDiskAvailabilityMap.Store(e.Object.GetName(), diskCount)
```
edreed, line 144: Predicates should not have side-effects. The Create predicate should simply return true and the Reconcile function should add the corresponding entry when it observes a node for which there is no current entry.

pkg/controller/shared_state.go
```diff
@@ -77,6 +77,8 @@ type SharedState struct {
 	crdClient                     crdClientset.Interface
 	conditionWatcher              *watcher.ConditionWatcher
 	azureDiskCSITranslator        csitranslator.InTreePlugin
+	nodeDiskAvailabilityMap       sync.Map
```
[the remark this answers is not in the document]
edreed: If `atomic.Int32` is used as I recommend, then we will not need to lock or separate struct.

pkg/controller/attach_detach.go
```diff
@@ -120,7 +131,20 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
...
+		defer r.SharedState.nodeDiskAvailabilityMapLock.Unlock()
+		// azVolumeAttachment state is nil which means it's a brand new attachment request instead of a retry
+		// the node's capacity of disk attachment should be decreased by 1
+		remainingCapacity, nodeExists := r.SharedState.nodeDiskAvailabilityMap.Load(azVolumeAttachment.Spec.NodeName)
+		if nodeExists && remainingCapacity != nil && remainingCapacity.(int) > 0 {
+			r.SharedState.nodeDiskAvailabilityMap.Store(azVolumeAttachment.Spec.NodeName, remainingCapacity.(int)-1)
+		} else {
+			err := status.Errorf(codes.Aborted, "node (%s) is unavailable for disk attachment", azVolumeAttachment.Spec.NodeName)
+			return reconcileReturnOnError(ctx, azVolumeAttachment, "attach", err, r.retryInfo)
+		}
```
edreed, line 144: I believe it would be better to perform the allocation in `triggerAttach` rather than directly in the reconciler.

pkg/controller/attach_detach.go
edreed: I agree with @hccheng72 - it would simplify the code.

alice-zheyan-yu: Done

pkg/controller/shared_state.go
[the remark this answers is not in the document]
alice-zheyan-yu: Done

alice-zheyan-yu:
> 

Done

pkg/controller/attach_detach.go
alice-zheyan-yu: Done

pkg/controller/azdrivernode.go
alice-zheyan-yu: Done

pkg/controller/azdrivernode.go
alice-zheyan-yu: Deleted

pkg/controller/attach_detach.go
alice-zheyan-yu: Done

alice-zheyan-yu:
Close this draft PR and will create a new PR for it.

PR #1570, opened 2022-10-28
[WIP][V2] feat: watch CRD deletion to track driver uninstall

hccheng72:
**What type of PR is this?**
/kind feature

This PR makes the decision on whether the cloud resource needs to be deleted / detached or not using CRD deletionTimestamp.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/controller/attach_detach.go
```diff
@@ -312,59 +314,59 @@ func (r *ReconcileAttachDetach) triggerDetach(ctx context.Context, azVolumeAttac
...
-		updateFunc := func(obj client.Object) error {
-			azv := obj.(*azdiskv1beta2.AzVolumeAttachment)
-			// Update state to detaching
-			_, derr := updateState(azv, azdiskv1beta2.Detaching, normalUpdate)
-			return derr
+	updateFunc := func(obj client.Object) error {
+		azv := obj.(*azdiskv1beta2.AzVolumeAttachment)
+		// Update state to detaching
+		updateMode := normalUpdate
+		if r.isDriverUninstall() {
```
sunpa93, line 327: We use state transition as a way to prevent multiple goroutines working on same tasks. If we use the `force-update` mode here, we will end up with multiple goroutines requesting detach for the same node and volume pair.

pkg/controller/shared_state.go
```diff
@@ -79,25 +84,52 @@ type SharedState struct {
...
+		eventRecorder:          eventRecorder,
+		cachedClient:           cachedClient,
+		crdClient:              crdClient,
+		azClient:               azClient,
+		kubeClient:             kubeClient,
+		conditionWatcher:       watcher.New(ctx, azClient, azdiskinformers.NewSharedInformerFactory(azClient, consts.DefaultInformerResync), objectNamespace),
 		azureDiskCSITranslator: csitranslator.NewAzureDiskCSITranslator(),
 	}
+
+	if crdClient != nil {
```
sunpa93, line 101: We can rely on `WatchObject` function to do this instead of creating a new shared informer factory for CRD object; it's a recent addition from this commit (https://github.com/kubernetes-sigs/azuredisk-csi-driver/commit/66e996a2b40de6ffdd250a3492f2d9dbf79a2c69#diff-9af1228cdbda651b5dd579508719c1815e2297e0a4ed5326d82e196eb7eb6972). Using this, we can simplify things by setting up a watch with controller runtime package and rely on cached reader to fetch object states as well. You probably will need to modify the function to allow for predicate parameters so that we can add appropriate eventhandlers for each event.

https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/6bb1a49e3c3351856c6d572da92801472ce8b836/pkg/controller/common.go#L1105.

pkg/controller/replica.go
```diff
@@ -91,6 +91,11 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
 		if volumeDetachRequested(azVolumeAttachment) {
 			switch azVolumeAttachment.Status.State {
 			case azdiskv1beta2.DetachmentFailed:
+				// if detachment failed and the driver is uninstalling, delete azvolumeattachment CRI to let uninstallation proceed
+				if r.isDriverUninstall() {
+					_ = r.cachedClient.Delete(ctx, azVolumeAttachment)
```
sunpa93, line 96: It would probably be good to emit a log to indicate that the replica attachment CRI is being deleted without detachment.

pkg/controller/azvolume.go
```diff
@@ -334,7 +334,8 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
 
 					deleteErr = r.deleteVolume(cloudCtx, azVolume)
 				}
-				if deleteErr != nil {
+				// if volume deletion failed and the driver is uninstalling, delete azvolume CRI to let uninstallation proceed
+				if deleteErr != nil && !r.isDriverUninstall() {
```
sunpa93, line 338: I am not sure if we should be handling this here. I think we should update the error first and handle driver uninstall case on the next reconciler iteration on the update event this will be triggering.

pkg/azureconstants/azure_constants.go
```diff
@@ -122,6 +122,7 @@ const (
 	VolumeAttachRequestAnnotation         = "disk.csi.azure.com/volume-attach-request"
 	VolumeDeleteRequestAnnotation         = "disk.csi.azure.com/volume-delete-request"
 	VolumeDetachRequestAnnotation         = "disk.csi.azure.com/volume-detach-request"
+	CloudDeleteVolume                     = "cloud-delete-volume"
```
edreed, line 125: np: Let's move this up next to the `VolumeDeleteRequestAnnotation` constant since that is what it is used with.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -45,6 +45,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/klog/v2/klogr"
 
+	crd "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
edreed, line 48: np: Consider `apiextensionsv1` instead of `crd` for consistency in package aliasing.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -405,11 +408,46 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
...
+		},
+		UpdateFunc: func(_ event.UpdateEvent) bool {
+			return true
+		},
+		GenericFunc: func(_ event.GenericEvent) bool {
+			return false
+		},
+		DeleteFunc: func(_ event.DeleteEvent) bool {
+			return false
+		},
```
edreed, line 423: #1542 added the watcher to keep namespaces cached (I thought). Ignoring create and delete events would cause the cache to be stale. @sunpa93 Is this change correct?

sunpa93: As far as I know, the predicates are strictly for event handler and wouldn't affect the caching behavior. So, we actually should be good if all predicates were to return false.

pkg/controller/attach_detach.go
```diff
@@ -312,64 +314,60 @@ func (r *ReconcileAttachDetach) triggerDetach(ctx context.Context, azVolumeAttac
...
 	defer func() { w.Finish(err) }()
-	// only detach if detachment request was made for underlying volume attachment object
-	detachmentRequested := volumeDetachRequested(azVolumeAttachment)
 
-	if detachmentRequested {
-		defer r.stateLock.Delete(azVolumeAttachment.Name)
-		if _, ok := r.stateLock.LoadOrStore(azVolumeAttachment.Name, nil); ok {
-			return getOperationRequeueError("detach", azVolumeAttachment)
-		}
+	defer r.stateLock.Delete(azVolumeAttachment.Name)
```
edreed, line 318: Should the delete of the azva state lock occur only if we acquire the lock on the next line? Otherwise, we'll delete the state lock when we haven't acquired it.

pkg/controller/azvolume.go
```diff
@@ -328,12 +328,18 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
 			}
 
 			if err == nil {
+				// if the driver is uninstalling, delete azvolume CRI to let uninstallation proceed
 				if volumeDeleteRequested {
-					cloudCtx, cloudCancel := context.WithTimeout(goCtx, cloudTimeout)
-					defer cloudCancel()
+					if r.isDriverUninstall(ctx) {
+						goWorkflow.Logger().Errorf(fmt.Errorf("VolumeDeleteRequestError"), "trying to delete volume (%s) as the driver is uninstalling, deletion will be ignored", azVolume.Name)
```
edreed, line 334: This is an expected condition and shouldn't be logged as an error. If logged, it should be done at a higher verbosity level, e.g. `goWorkflow.Logger().V(12).Infof(...)`.

pkg/azuredisk/azuredisk_v2.go
edreed: I see now that this was the predicate originally used in the `controller.WatchObject` implementation for #1542. I'll assume this is correct, then.

pkg/azureconstants/azure_constants.go
```diff
@@ -122,6 +122,7 @@ const (
...
 	RecoverAnnotation                     = "disk.csi.azure.com/recovery" // used to ensure reconciliation is triggered for recovering CRIs
 	VolumeNameLabel                       = "disk.csi.azure.com/volume-name"
 	VolumeIDLabel                         = "disk.csi.azure.com/volume-id"
```
edreed, line 128: It looks like we no longer use `PreProvisionedVolumeCleanupAnnotation`. It should be removed.

pkg/controller/replica.go
edreed: Use `workflow.GetObjectDetails` to include `AzVolumeAttachment` details, e.g.

```go
r.logger.Info("Deleting AzVolumeAttachment without detaching disk", workflow.GetObjectDetails(azVolumeAttachment)...)
```

pkg/controller/shared_state.go
```diff
@@ -81,23 +82,35 @@ type SharedState struct {
...
+		conditionWatcher:       watcher.New(context.Background(), azClient, azdiskinformers.NewSharedInformerFactory(azClient, consts.DefaultInformerResync), objectNamespace),
 		azureDiskCSITranslator: csitranslator.NewAzureDiskCSITranslator(),
 	}
 	newSharedState.createReplicaRequestsQueue()
 
 	return newSharedState
 }
 
+func (c *SharedState) isDriverUninstall(ctx context.Context) bool {
+	w, _ := workflow.GetWorkflowFromContext(ctx)
```
edreed, line 102: The workflow doesn't seem to be used in this function. Do we need to call `GetWorkflowFromContext` here?

pkg/provisioner/crdprovisioner.go
```diff
@@ -1228,3 +1253,24 @@ func startWatchingKube(ctx context.Context, kubeInformerFactory informers.Shared
 		os.Exit(1)
 	}
 }
+
+func startWatchingCrd(ctx context.Context, crdInformerFactory crdInformers.SharedInformerFactory, driverUninstall *uint32) {
+	crdInformer := crdInformerFactory.Apiextensions().V1().CustomResourceDefinitions().Informer()
+	crdInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: nil,
+		UpdateFunc: func(_, new interface{}) {
```
edreed, line 1261: Shouldn't we do this on delete rather than update? I don't believe we'd be guaranteed to see the update for the CRD deletion. There do not appear to be any finalizers, so deletion should be near instantaneous.

pkg/azuredisk/azuredisk_v2.go
hccheng72: I replace `crd` with `apiext` to be consistent with other packages in `sigs.k8s.io`.

pkg/controller/attach_detach.go
hccheng72: Yeah, it makes sense to me. I'll change all of them in our codebase.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -405,11 +408,46 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
 		klog.Errorf("Failed to initialize NodeAvailabilityController. Error: %v. Exiting application...", err)
 		os.Exit(1)
 	}
-	err = controller.WatchObject(mgr, source.Kind{Type: &v1.Namespace{}})
+	err = controller.WatchObject(mgr, source.Kind{Type: &v1.Namespace{}}, predicate.Funcs{
+		CreateFunc: func(_ event.CreateEvent) bool {
```
landreasyan, line 412: @sunpa93 seems like we are creating a noop controller here. What is the point of passing the predicates in this case? Even if the predicate returns true, nothing will be triggered.

landreasyan: It might also be good to rename this `CacheObject` if that is the intend of creating the noop controller.

pkg/provisioner/crdprovisioner.go
```diff
@@ -1228,3 +1253,24 @@ func startWatchingKube(ctx context.Context, kubeInformerFactory informers.Shared
 		os.Exit(1)
 	}
 }
+
+func startWatchingCrd(ctx context.Context, crdInformerFactory crdInformers.SharedInformerFactory, driverUninstall *uint32) {
+	crdInformer := crdInformerFactory.Apiextensions().V1().CustomResourceDefinitions().Informer()
+	crdInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: nil,
```
landreasyan, line 1260: remove `AddFunc` and `DeleteFunc` instead of setting to `nil`

pkg/provisioner/crdprovisioner.go
```diff
@@ -175,13 +181,18 @@ func NewCrdProvisioner(kubeConfig *rest.Config, objNamespace string) (*CrdProvis
...
+	crdClient, err := crdClientset.NewForConfig(kubeConfig)
+	if err != nil {
+		return nil, err
+	}
 
 	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, consts.DefaultInformerResync)
 	startWatchingKube(context.Background(), kubeInformerFactory)
 
 	azInformerFactory := azdiskinformers.NewSharedInformerFactory(azdiskClient, consts.DefaultInformerResync)
+	crdInformerFactory := crdInformers.NewSharedInformerFactory(crdClient, consts.DefaultInformerResync)
```
landreasyan, line 193: Refactor the setup of all informers to a separate method `SetupInformers` that will initialize all the informers together and will wait for cache sync for all at the same time rather one at a time, i.e. ```cache.WaitForCacheSync(
		ctx.Done(),
                azInformerFactory.Informer().HasSynced,
                kubeInformerFactory.Informer().HasSynced, ...```
                
I am also not sure we currently wait for cacheSync for azInformerFactory.

pkg/controller/shared_state.go
edreed: Ah, sorry I missed the call to the workflow logger below.

pkg/provisioner/crdprovisioner.go
hccheng72: We wait for cacheSync for azInformerFactory [here](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/ed9c00b395b7e6c65ae0626dd0000a16586ba0f0/pkg/watcher/conditionwatcher.go#L70-L106).

pkg/controller/attach_detach.go
```diff
@@ -110,8 +110,10 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
 
 	// deletion request
 	if objectDeletionRequested(azVolumeAttachment) {
-		if err := r.removeFinalizer(ctx, azVolumeAttachment); err != nil {
-			return reconcileReturnOnError(ctx, azVolumeAttachment, "delete", err, r.retryInfo)
+		if azVolumeAttachment.Status.State != azdiskv1beta2.Attaching && azVolumeAttachment.Status.State != azdiskv1beta2.Detaching {
```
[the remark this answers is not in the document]
hccheng72: I redefine `isOperationInProcess` function and let the driver will update the state to `Detached` before deleting the `AzVoluemAttachment` CRI.

pkg/controller/shared_state.go
```diff
@@ -81,23 +82,35 @@ type SharedState struct {
...
+	crdList := &apiext.CustomResourceDefinitionList{}
+	if err := c.cachedClient.List(ctx, crdList, &client.ListOptions{}); err != nil {
+		w.Logger().Error(err, "failed to list CRDs to detect if the driver is uninstalling")
+	}
+
+	if len(crdList.Items) > 0 {
```
sunpa93, line 108: It seems like we are returning `true` on `isDriverUninstall()` if we have crd in the cluster. We should only be returning true if `AzVolume` or `AzVolumeAttachment` is marked with a deletionTimestamp.

pkg/azuredisk/azuredisk_v2.go
sunpa93: > @sunpa93 seems like we are creating a noop controller here. What is the point of passing the predicates in this case? Even if the predicate returns true, nothing will be triggered.

The reason we create a noop controller is so that we can use the same cached client to fetch namespaces instead of creating a dedicated informer for it to store in the shared state, which will require a separate syntax for fetching objects.

sunpa93: Unfortunately, the pkg doesn't seem to offer us a way to inject our own informer to its cached reader.

landreasyan: Yeah, I understand that part, but my question here is - do we need the predicates? We shouldn't need then for a noop controller. Also, we should probably rename the function like suggested above for more clarity.

sunpa93: Without the predicates returning false, we would have the namespace objects get added to the internal queue for reconcile operation, which will return with a noop. So, I figured it would be a waste of memory.

sunpa93: I agree with the naming. I think Huichan has already changed the name of the functions.

hccheng72: I renamed the functions to `CacheNamespace` and `CacheCRD`.

landreasyan: What I was saying is that all predicates should return false, because there is no controller to reconcile even if they return true

pkg/controller/common.go
```diff
@@ -1130,3 +1129,36 @@ func WatchObject(mgr manager.Manager, objKind source.Kind) error {
...
+
+	// Watch for CRUD events on CRD objects
+	err = c.Watch(&source.Kind{Type: &apiext.CustomResourceDefinition{}}, &handler.EnqueueRequestForObject{}, predicate.Funcs{
+		CreateFunc: func(_ event.CreateEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			crdObj, ok := e.ObjectNew.(*apiext.CustomResourceDefinition)
+			if ok && (crdObj.Name == consts.AzVolumeAttachmentCRDName || crdObj.Name == consts.AzVolumeCRDName) && crdObj.DeletionTimestamp != nil {
+				return true
```
sunpa93, line 1152: We might want to dispatch a goroutine to update the atomic value representing `isDriverUninstall` like we do for `isRecoveryComplete` instead of listing all CRDs to determine if the driver is being uninstalled or not.

https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/95813ab637a8b79a0646701d94209929fa3a159e/pkg/controller/shared_state.go#L118

pkg/azuredisk/azuredisk_v2.go
sunpa93: As I left in the earlier comments of the PR, the predicates for these noop reconcilers should return false.

pkg/controller/common.go
```diff
@@ -1130,3 +1129,36 @@ func WatchObject(mgr manager.Manager, objKind source.Kind) error {
...
+	}
+
+	c.GetLogger().Info("Starting to watch CRDs")
+
+	// Watch for CRUD events on CRD objects
+	err = c.Watch(&source.Kind{Type: &apiext.CustomResourceDefinition{}}, &handler.EnqueueRequestForObject{}, predicate.Funcs{
+		CreateFunc: func(_ event.CreateEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
```
landreasyan, line 1149: We don't need to return true in the predicate here because we have another informer for CRDs and this is a noop controller. I am not sure why `ChacheCRD` and `CacheNamespace` have been refactored into two different methods when they could have been one `CacheObject` method taking the object type as a parameter.

hccheng72: If the predicate isn't needed, I can change it to only one `CacheObject` method.

sunpa93: > CacheObject method taking the object type as a parameter.
This wouldn't do for CRD case as it requires a manual predicate. 

We either need to create `CacheObject(mgr manager.Manager, obj runtime.Object, options ...controller.WatchOption)` to allow for option to manually configure predicates (like we do for workflow objects) https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/a2fde828b8a4fd9e3de24543861583b5a742a048/pkg/workflow/workflow.go#L84-L120

Or we need them as separate functions.

I personally think using `WatchOption` would give the code more versatility

pkg/provisioner/crdprovisioner.go
landreasyan: This is not related to your change, but there is an indirect dependency here. condition watcher starts the `azInformerFactory` that is initialized in this method, later CacheReader relies on the `azInformerFactory` to be started. So, maybe a better approach would be for watcher.New to initialize and return the started `azInformerFactory` that will be passed to `CachedReader`. Yet I think the best approach would be to initialize and start all controllers in one `SetupInformers` method and then pass them to appropriate components that rely on them.

pkg/controller/common.go
landreasyan: We have an informer that triggers a defined update function for the CRD. Why do we need a predicate that returns true for a noop controller here?

sunpa93: We don't have an informer that triggers anything for the CRD. We need a predicate to detect if `AzVolume` or `AzVolumeAttachment` CRD is requested for deletion. 

As I commented above, I believe we should use atomic value to denote if the driver is being uninstalled or not, like we do for `isRecoveryComplete`, instead of listing all the CRDs every time `isDriverUninstall()` is called.

landreasyan: Here is the informer: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1570/files#diff-969dc63b611428d6a3320f889ab448a42f1288c3a211533f19dc2a7bf526ef75R1250

hccheng72: Oh, I just realize that we can have predicate but it only returns true. If that, I created two functions because their predicates are different, and defining predicates as function parameter in `azureisk_v2` makes the code difficult to read that you commented earlier in the PR.

sunpa93: That informer lives in `crdProvsioner` not in the controllers. We don't want the `crdProvisioner` and controller to have shared memory.

landreasyan: I believe we should use the informer instead of the predicate to achieve that functionality as it is currently done. All predicates for caching should return false for a noop controller. And predacities should not be initializing any values as their purpose is to decide if an event should be triggered for reconciliation or not.

landreasyan: We should create an informer in the controller for the CRDs as well if that is the case.

sunpa93: Internally, it's exactly the same thing unless we use a queue, and worker goroutines to dequeue objects and run reconciliation on them, which is the standard way to write controllers. Handling things directly in event handler vs. handling in controller runtime predicate func is essentially going to be no different.

sunpa93: I understand that such usage of predicate func goes against its intended use case but I also am not a big fan of adding bunch of shared informers when we can leverage controller-runtime pkg to centrally manage cache.

pkg/controller/common.go
```diff
@@ -1100,26 +1101,57 @@ func verifyObjectFailedOrDeleted(obj interface{}, objectDeleted bool) (bool, err
...
+			crdObj, ok := e.ObjectNew.(*apiext.CustomResourceDefinition)
+			if ok && (crdObj.Name == consts.AzVolumeAttachmentCRDName || crdObj.Name == consts.AzVolumeCRDName) && crdObj.DeletionTimestamp != nil {
+				atomic.StoreUint32(&sharedState.driverUninstall, 1)
```
sunpa93, line 1152: We don't want to block the main thread so if we were going to handle things in a predicate, it would need to be done in a goroutine. And we probably should only update `if !sharedState.IsDriverUninstall()`

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -405,11 +406,16 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
 		klog.Errorf("Failed to initialize NodeAvailabilityController. Error: %v. Exiting application...", err)
 		os.Exit(1)
 	}
-	err = controller.WatchObject(mgr, source.Kind{Type: &v1.Namespace{}})
+	err = controller.CacheNamespace(mgr)
 	if err != nil {
 		klog.Errorf("Failed to watch Namespace. Error: %v. Exiting application...", err)
 		os.Exit(1)
 	}
+	err = controller.CacheCRD(mgr, sharedState)
```
landreasyan, line 414: If we are going with the predicate approach here, this is no longer caching, this method should be something like `WatchForDriverUninstall`

hccheng72: Sure! When we finalize which approach to go, I'll change it.

pkg/provisioner/crdprovisioner.go
```diff
@@ -1217,14 +1245,25 @@ func waitForDetachVolumeFunc(obj interface{}, objectDeleted bool) (bool, error)
 	return false, nil
 }
 
-func startWatchingKube(ctx context.Context, kubeInformerFactory informers.SharedInformerFactory) {
+// Initializing nodeInformer and crdInformer and waiting for cache sync
+func startWatchingObjs(ctx context.Context, kubeInformerFactory informers.SharedInformerFactory, crdInformerFactory crdInformers.SharedInformerFactory, driverUninstallState *uint32) {
```
edreed, line 1249: Consider making `startWatchingObjs` a method of `crdProvisioner` to avoid passing the informers and install state as arguments.

pkg/controller/shared_state.go
```diff
@@ -77,27 +77,32 @@ type SharedState struct {
 	crdClient                     crdClientset.Interface
 	conditionWatcher              *watcher.ConditionWatcher
 	azureDiskCSITranslator        csitranslator.InTreePlugin
+	driverUninstallState          *uint32
```
edreed, line 80: Rather the store a pointer to the uninstall state, define an interface that defines a single method, `IsDriverUninstalling()`, to be implemented by the `CrdProvisioner`.

pkg/controller/attach_detach.go
```diff
@@ -312,64 +311,69 @@ func (r *ReconcileAttachDetach) triggerDetach(ctx context.Context, azVolumeAttac
...
+				return derr
+			}
+			//nolint:contextcheck // final status update of the CRI must occur even when the current context's deadline passes.
+			if _, uerr := azureutils.UpdateCRIWithRetry(goCtx, nil, r.cachedClient, r.azClient, azVolumeAttachment, updateFunc, consts.ForcedUpdateMaxNetRetry, updateMode); uerr != nil {
+				w.Logger().Errorf(uerr, "failed to update final status of AzVolumeAttachement")
+			}
+		} else {
+			updateFunc = func(obj client.Object) error {
+				azv := obj.(*azdiskv1beta2.AzVolumeAttachment)
+				_, derr := updateState(azv, azdiskv1beta2.Detached, normalUpdate)
```
edreed, line 362: Why are we updating the state to `Detached` here since we're immediately marking it for delete? (I have a vague recollection of some discussion, but I forget the details.)

pkg/controller/attach_detach.go
```diff
@@ -312,64 +311,69 @@ func (r *ReconcileAttachDetach) triggerDetach(ctx context.Context, azVolumeAttac
...
+		_, goWorkflow := workflow.New(ctx)
+		defer func() { goWorkflow.Finish(detachErr) }()
+		waitCh <- goSignal{}
+
+		goCtx := goWorkflow.SaveToContext(context.Background())
+		cloudCtx, cloudCancel := context.WithTimeout(goCtx, cloudTimeout)
+		defer cloudCancel()
+
+		var updateFunc func(obj client.Object) error
+		updateMode := azureutils.UpdateCRIStatus
```
sunpa93, line 347: I think we can get rid of the `updateMode` variable here and just feed `UpdateCRIStatus` directly to `UpdateCRIWithRetry` function.

pkg/controller/attach_detach.go
hccheng72: This update is to prevent orphaned `AzVolumeAttachment` CRIs mentioned in the [comment](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1570#discussion_r1017012162).

sunpa93: Let's seek different way to avoid orphaned `AzVolumeAttachment` scenario as this will add to our already heavy apiserver load.

hccheng72: How about removing finalizer before making the [DELETE request](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/edd3cb8795b269297cc68820418830b4eab43623/pkg/controller/attach_detach.go#L371)?

edreed: Both the `replica` and `attach_detach` controllers can delete an `AzVolumeAttachment`. The finalizer is removed by the `attach_detach` controller's `Reconcile` function and allows the `replica` controller to coordinate on the deletion of a failed replica attachment. At least, this is my understanding of the flow.

@sunpa93 Would it work for both the `replica` and `attach_detach` controllers to remove the finalizer and delete the `AzVolumeAttachment` objects as needed?

hccheng72: Confirmed with Sung Jun, it seems like changing definition of `isOperationInProcess` function back to `return target.Status.State == azdiskv1beta2.Attaching || (target.Status.State == azdiskv1beta2.Detaching && !deleteRequested)` can solve this issue.

pkg/controller/shared_state.go
```diff
@@ -79,21 +79,26 @@ type SharedState struct {
 	conditionWatcher              *watcher.ConditionWatcher
 	azureDiskCSITranslator        csitranslator.InTreePlugin
 	availableAttachmentsMap       sync.Map
+	crdProvisioner                CrdProvisioner
 }
 
-func NewSharedState(driverName, objectNamespace, topologyKey string, eventRecorder record.EventRecorder, cachedClient client.Client, azClient azdisk.Interface, kubeClient kubernetes.Interface, crdClient crdClientset.Interface) *SharedState {
+type CrdProvisioner interface {
```
edreed, line 85: Consider `DriveLifecycle` to avoid confusion with `CrdProvisioner` defined elsewhere.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -348,7 +350,7 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
 		os.Exit(1)
 	}
 
-	sharedState := controller.NewSharedState(d.Name, d.config.ObjectNamespace, topologyKey, eventRecorder, mgr.GetClient(), d.crdProvisioner.GetDiskClientSet(), d.kubeClient, crdClient)
+	sharedState := controller.NewSharedState(d.Name, d.config.ObjectNamespace, topologyKey, eventRecorder, mgr.GetClient(), d.crdProvisioner.GetDiskClientSet(), d.crdProvisioner.(*provisioner.CrdProvisioner), d.kubeClient, crdClient)
```
edreed, line 353: Can we avoid the cast to the underlying `provisioner.CrdProvisioner` type by adding the `IsDriverUninstall()` method to the `azuredisk.CrdProvisioner` interface? If not, it would be better to remove the `azuredisk.CrdProvisioner` interface and use the `provisioner.CrdProvisioner` type directly in `DriverV2`.

edreed: Another alternative is to embed the `controller.DriverLifecycle` interface in the `azuredisk.CrdProvisioner` interface.

pkg/controller/shared_state.go
```diff
@@ -79,21 +79,26 @@ type SharedState struct {
 	conditionWatcher              *watcher.ConditionWatcher
 	azureDiskCSITranslator        csitranslator.InTreePlugin
 	availableAttachmentsMap       sync.Map
+	crdProvisioner                CrdProvisioner
```
edreed, line 82: Let's use `driverLifecycle` here.

pkg/controller/common_test.go
```diff
@@ -451,7 +452,7 @@ func convertToNode(obj client.Object) *v1.Node {
 }
 
 func initState(client client.Client, azClient azdisk.Interface, kubeClient kubernetes.Interface, objs ...runtime.Object) (c *SharedState) {
-	c = NewSharedState(consts.DefaultDriverName, consts.DefaultAzureDiskCrdNamespace, consts.WellKnownTopologyKey, &record.FakeRecorder{}, client, azClient, kubeClient, nil)
+	c = NewSharedState(consts.DefaultDriverName, consts.DefaultAzureDiskCrdNamespace, consts.WellKnownTopologyKey, &record.FakeRecorder{}, client, azClient, &provisioner.CrdProvisioner{}, kubeClient, nil)
```
edreed, line 455: Instead of using an uninitialized `provisioner.CrdProvisioner` here, consider adding a `FakeDriverLifecycle` or using [`gomock`](https://github.com/golang/mock) to create a mock.

pkg/controller/shared_state.go
```diff
@@ -53,6 +53,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type DriveLifecycle interface {
```
edreed, line 56: sp: `DriverLifecyle`, `driverLifecyle`

pkg/controller/shared_state.go
```diff
@@ -53,6 +53,9 @@ import (
...
+	IsDriverUninstall() bool
+}
 type SharedState struct {
```
edreed, line 59: np: insert blank line between structure definitions.

edreed:
I left a couple of small nitpick comments, but otherwise LGTM.

hccheng72:
> I left a couple of small nitpick comments, but otherwise LGTM.

I've fixed those issues. Could you please approve this PR so I can incorporate new changes to [the other related PR](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1607) that would refactor informers in one place?

PR #1572, opened 2022-10-31
[V2] feat: enable controller pod customization

edreed:
**What type of PR is this?**

/kind feature

Enables limited customization of the controller pod.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


charts/latest-v2/azuredisk-csi-driver/values.yaml
```diff
@@ -135,6 +135,19 @@ controller:
       requests:
         cpu: 10m
         memory: 20Mi
+  environments:
```
sunpa93, line 138: Are the values under environments used anywhere?

edreed: This was from an early iteration. Removed.

PR #1581, opened 2022-11-07
[V2] fix: cloud-provisioner should always report through result channel before returning

sunpa93:
**What type of PR is this?**
/kind bug

* Fixes a bug where cloudprovisioner will not return result through result channel if disk is already attached.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/provisioner/cloudprovisioner.go
+++ b/pkg/provisioner/cloudprovisioner.go
@@ -370,6 +370,7 @@ func (c *CloudProvisioner) PublishVolume(
 	attachResult = NewCloudAttachResult()
 	defer func() {
 		if !waitForCloud {
+			attachResult.ResultChannel() <- err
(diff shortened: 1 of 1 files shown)


pkg/provisioner/cloudprovisioner.go
```diff
@@ -370,6 +370,7 @@ func (c *CloudProvisioner) PublishVolume(
 	attachResult = NewCloudAttachResult()
 	defer func() {
 		if !waitForCloud {
+			attachResult.ResultChannel() <- err
 			close(attachResult.ResultChannel())
```
landreasyan, line 374: you are closing the channel twice - here and in the goroutine. This can result in a panic.

PR #1583, opened 2022-11-07
[V2]feat: adding mount replica managed resource

connorbradshaw10:
**What type of PR is this?**

/kind feature

- adding disk.csi.azure.com/mount-replicas resource to scheduler extender
- changed all of the e2e tests to use new managed resource

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


edreed:
@abhisheksinghbaghel Do we want to use the cluster managed resource approach upstream? We already deploy a SxS scheduler and need the scheduler name specified in the pod, so cluster managed resource is redundant.

PR #1592, opened 2022-11-10
[V2] feat: Add feature flag enableMountReplicas to v2 driver

mukhoakash:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/provisioner/crdprovisioner.go
```diff
@@ -846,6 +849,10 @@ func (c *CrdProvisioner) UnpublishVolume(
 		return err
 	}
 
+	if !c.mountReplicasEnabled {
```
landreasyan, line 852: MaxShares on AzVolume should already be set to 0, right? I don't think we need to check mountReplcias here.

mukhoakash: Consider a driver update scenario. Initially the mountReplicasEnabled was true, then we updated it to false.

landreasyan: On driver upgrade the recovery function should set the correct value in the CRDs. I don't think we should keep the maxshares value and the replica management inconsistent.

charts/latest-v2/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
```diff
@@ -207,6 +207,7 @@ spec:
           args:
             - "--v={{ .Values.controller.logLevel }}"
             - "--config=/etc/{{ .Values.controller.name }}/config.yaml"
+            - "--enable-mount-replicas={{ .Values.controller.enableMountReplicas }}"
```
landreasyan, line 210: We should add `enable-mount-replicas` to the configMap defined before the deployment here instead of adding a command line parameter.

sunpa93, line 210: We should probably use configMap for this.

charts/latest-v2/azuredisk-csi-driver/values.yaml
```diff
@@ -58,6 +58,7 @@ controller:
   cloudConfigSecretName: azure-cloud-provider
   cloudConfigSecretNamespace: kube-system
   allowEmptyCloudConfig: false
+  enableMountReplicas: false
```
sunpa93, line 61: The value for this in our latest-v2 chart should be `true` I believe.

pkg/controller/shared_state.go
```diff
@@ -1277,7 +1279,7 @@ func (c *SharedState) createAzVolumeFromPv(ctx context.Context, pv v1.Persistent
 	if pv.Spec.NodeAffinity != nil && pv.Spec.NodeAffinity.Required != nil {
 		desiredAzVolume.Status.Detail.AccessibleTopology = azureutils.GetTopologyFromNodeSelector(*pv.Spec.NodeAffinity.Required, c.topologyKey)
 	}
-	if azureutils.IsMultiNodePersistentVolume(pv) {
+	if azureutils.IsMultiNodePersistentVolume(pv) || !c.mountReplicasEnabled {
```
sunpa93, line 1282: This check would be redundant to https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/02ecd983aaf68aaa0184f5b04b574e2ad17bb830/pkg/controller/shared_state.go#L1321

sunpa93:
We probably need to add code to update `azVolume.Spec.MaxMountReplicaCount` in https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/02ecd983aaf68aaa0184f5b04b574e2ad17bb830/pkg/controller/shared_state.go#L1355 for upgrade scenario.

charts/latest-v2/azuredisk-csi-driver/values.yaml
mukhoakash: In the PRD, the default value is set to false.

pkg/provisioner/crdprovisioner.go
```diff
@@ -592,7 +594,8 @@ func (c *CrdProvisioner) PublishVolume(
 	}
 
 	// detach replica volume if volume's maxShares have been fully saturated.
-	if azVolume.Spec.MaxMountReplicaCount > 0 {
+	// need to validate these conditions only when mountReplicas is enabled.
+	if c.mountReplicasEnabled && azVolume.Spec.MaxMountReplicaCount > 0 {
```
landreasyan, line 598: here too checking `azVolume.Spec.MaxMountReplicaCount` should be enough

pkg/azureutils/azure_disk_utils.go
```diff
@@ -930,9 +930,15 @@ func GetMaxSharesAndMaxMountReplicaCount(parameters map[string]string, isMultiNo
 		maxShares = 1
 	}
 
-	if isMultiNodeVolume {
+	if isMultiNodeVolume || !mountReplicasEnabled {
```
landreasyan, line 933: this check is unnecessary, these two values are checked later before logging.

charts/latest-v2/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
mukhoakash: fixed

charts/latest-v2/azuredisk-csi-driver/values.yaml
mukhoakash: fixed

pkg/controller/shared_state.go
mukhoakash: fixed

pkg/provisioner/crdprovisioner.go
mukhoakash: fixed

pkg/provisioner/crdprovisioner.go
mukhoakash: fixed

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -118,6 +118,7 @@ type DriverV2 struct {
 	azdiskClient         azdisk.Interface
 	azDriverNodeInformer azdiskinformertypes.AzDriverNodeInformer
 	deviceChecker        *deviceChecker
+	enableMountReplicas  bool
```
edreed, line 121: Since the configuration is referenced in the `config` field above, there's no need to add an additional field for this setting here.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -348,7 +350,7 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
 		os.Exit(1)
 	}
 
-	sharedState := controller.NewSharedState(d.Name, d.config.ObjectNamespace, topologyKey, eventRecorder, mgr.GetClient(), d.crdProvisioner.GetDiskClientSet(), d.kubeClient, crdClient)
+	sharedState := controller.NewSharedState(d.Name, d.config.ObjectNamespace, topologyKey, eventRecorder, mgr.GetClient(), d.crdProvisioner.GetDiskClientSet(), d.kubeClient, crdClient, d.enableMountReplicas)
```
edreed, line 353: Instead of adding a new parameter, consider passing `d.config` instead of `d.Name` and `d.config.ObjectNamespace`.

pkg/controller/shared_state.go
```diff
@@ -1356,12 +1358,18 @@ func (c *SharedState) createAzVolume(ctx context.Context, desiredAzVolume *azdis
...
+	var specModified bool
 
 	if azVolume, err = c.azClient.DiskV1beta2().AzVolumes(c.objectNamespace).Get(ctx, desiredAzVolume.Name, metav1.GetOptions{}); err != nil {
 		if apiErrors.IsNotFound(err) {
 			if azVolume, err = c.azClient.DiskV1beta2().AzVolumes(c.objectNamespace).Create(ctx, desiredAzVolume, metav1.CreateOptions{}); err != nil {
 				return err
 			}
+			if azVolume.Spec.MaxMountReplicaCount != desiredAzVolume.Spec.MaxMountReplicaCount {
+				azVolume.Spec.MaxMountReplicaCount = desiredAzVolume.Spec.MaxMountReplicaCount
+				specModified = true
```
landreasyan, line 1370: np: `specModified` -> `specUpdateNeeded`

PR #1595, opened 2022-11-12
[V2] fix: add nodeDiskAvailabilityMap

alice-zheyan-yu:
**What type of PR is this?**
/kind bug

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Passed tests:**

(node's max attachment count is 8. disk MaxShares = 2)
2 nodes, 8 disks
3 nodes, 11 disks
20 nodes, 20 disks
30 nodes, 30 disks

**How I verified:**

$kubectl get azvolumeattachment -A | grep Replica and 
$kubectl get azvolumeattachment -A | grep Primary 
can print out correct azvolumeattachments

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/attach_detach.go
+++ b/pkg/controller/attach_detach.go
@@ -116,6 +116,8 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
 		if err := r.removeFinalizer(ctx, azVolumeAttachment); err != nil {
 			return reconcileReturnOnError(ctx, azVolumeAttachment, "delete", err, r.retryInfo)
 		}
+		// deletion of azVolumeAttachment is succeeded, the node's remaining capacity of disk attachment should be increased by 1
+		r.incrementAttachmentCount(ctx, azVolumeAttachment.Spec.NodeName)
 		// detachment request
 	} else if volumeDetachRequested(azVolumeAttachment) {
 		if err := r.triggerDetach(ctx, azVolumeAttachment); err != nil {
@@ -285,6 +287,8 @@ func (r *ReconcileAttachDetach) triggerAttach(ctx context.Context, azVolumeAttac
 						r.eventRecorder.Eventf(pod.DeepCopyObject(), v1.EventTypeNormal, consts.ReplicaAttachmentSuccessEvent, "Replica mount for volume %s successfully attached to node %s", azVolumeAttachment.Spec.VolumeName, azVolumeAttachment.Spec.NodeName)
 					}
 				}
+				// the node's remaining capacity of disk attachment should be decreased by 1, since the disk attachment is succeeded.
+				r.decrementAttachmentCount(ctx, azVolumeAttachment.Spec.NodeName)
 			}
 
 			updateFunc := func(obj client.Object) error {
--- a/pkg/controller/azdrivernode.go
+++ b/pkg/controller/azdrivernode.go
@@ -55,8 +55,12 @@ func (r *ReconcileAzDriverNode) Reconcile(ctx context.Context, request reconcile
 	n := &corev1.Node{}
 	err := r.cachedClient.Get(ctx, request.NamespacedName, n)
 
-	// If the node still exists don't delete the AzDriverNode
 	if err == nil {
+		if n.ObjectMeta.DeletionTimestamp.IsZero() {
+			// for create event, add the new node in availableAttachmentsMap
+			r.addNodeToAvailableAttachmentsMap(ctx, n.Name, n.GetLabels())
+		}
+		// for delete even, if the node still exists don't delete the AzDriverNode
 		return reconcile.Result{}, nil
 	}
 
@@ -70,6 +74,7 @@ func (r *ReconcileAzDriverNode) Reconcile(ctx context.Context, request reconcile
 		if err != nil && !errors.IsNotFound(err) {
 			return reconcile.Result{Requeue: true}, err
 		}
+		r.deleteNodeFromAvailableAttachmentsMap(ctx, request.Name)
 
 		// Delete all volumeAttachments attached to this node, if failed, requeue
 		if _, err = r.cleanUpAzVolumeAttachmentByNode(ctx, request.Name, azdrivernode, azureutils.AllRoles, cleanUpAttachment); err != nil {
@@ -101,6 +106,7 @@ func (r *ReconcileAzDriverNode) Recover(ctx context.Context) error {
 		if _, err = r.azClient.DiskV1beta2().AzDriverNodes(r.objectNamespace).Update(ctx, updated, metav1.UpdateOptions{}); err != nil {
 			return err
 		}
+		r.addNodeToAvailableAttachmentsMap(ctx, node.Name, node.GetLabels())
 	}
 	return nil
 }
@@ -127,7 +133,7 @@ func NewAzDriverNodeController(mgr manager.Manager, controllerSharedState *Share
 	// Predicate to only reconcile deleted nodes
 	p := predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
-			return false
+			return true
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			return false
--- a/pkg/controller/common_test.go
+++ b/pkg/controller/common_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/golang/mock/gomock"
@@ -197,6 +198,9 @@ var (
 	}
 
 	testVolumeAttachmentName = "test-attachment"
+
+	testNodeAvailableAttachmentCount   = 8
+	testNodeNoAvailableAttachmentCount = 0
 
 	testVolumeAttachment = storagev1.VolumeAttachment{
 		ObjectMeta: metav1.ObjectMeta{
@@ -750,4 +754,10 @@ func mockClients(mockClient *mockclient.MockClient, azVolumeClient azdisk.Interf
 			return nil
 		}).
 		AnyTimes()
+}
+
+func addTestNodeInAvailableAttachmentsMap(sharedState *SharedState, nodeName string, capacity int) {
+	var count atomic.Int32
+	count.Store(int32(capacity))
+	sharedState.availableAttachmentsMap.Store(nodeName, &count)
 }
--- a/pkg/controller/shared_state.go
+++ b/pkg/controller/shared_state.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
 	csitranslator "k8s.io/csi-translation-lib/plugins"
+	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/features"
 	azdiskv1beta2 "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/azuredisk/v1beta2"
 	azdisk "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/client/clientset/versioned"
@@ -77,6 +78,7 @@ type SharedState struct {
 	crdClient                     crdClientset.Interface
 	conditionWatcher              *watcher.ConditionWatcher
 	azureDiskCSITranslator        csitranslator.InTreePlugin
+	availableAttachmentsMap       sync.Map
 }
 
 func NewSharedState(driverName, objectNamespace, topologyKey string, eventRecorder record.EventRecorder, cachedClient client.Client, azClient azdisk.Interface, kubeClient kubernetes.Interface, crdClient crdClientset.Interface) *SharedState {
@@ -860,7 +862,7 @@ func (c *SharedState) selectNodesPerTopology(ctx context.Context, nodes []v1.Nod
 		}
 	}
 
-	return selectedNodes[:min(len(selectedNodes), numReplicas)], nil
+	return selectedNodes, nil
 }
 
 func (c *SharedState) getNodesWithReplica(ctx context.Context, volumeName string) ([]string, error) {
@@ -1173,12 +1175,14 @@ func (c *SharedState) createReplicas(ctx context.Context, remainingReplicas int,
 	for _, node := range nodes {
 		if err = c.createReplicaAzVolumeAttachment(ctx, volumeID, node, volumeContext); err != nil {
 			w.Logger().Errorf(err, "failed to create replica AzVolumeAttachment for volume %s", volumeName)
-			//Push to queue the failed replica number
-			request := ReplicaRequest{VolumeName: volumeName, Priority: remainingReplicas}
-			c.priorityReplicaRequestsQueue.Push(ctx, &request)
-			return err
+			// continue to try attachment with next node
+			continue
 		}
 		remainingReplicas--
+		if remainingReplicas <= 0 {
+			// no more remainingReplicas, don't need to create replica AzVolumeAttachment
+			break
+		}
 	}
 
 	if remainingReplicas > 0 {
@@ -1228,16 +1232,17 @@ func (c *SharedState) getNodesForReplica(ctx context.Context, volumeName string,
 	}
 
 	filtered := []string{}
-	numFiltered := 0
 	for _, node := range nodes {
-		if numFiltered >= numReplica {
-			break
-		}
 		if skipSet[node] {
 			continue
 		}
+		// if the node has no capacity for disk attachment, we should skip it
+		remainingCapacity, nodeExists := c.availableAttachmentsMap.Load(node)
+		if !nodeExists || remainingCapacity == nil || remainingCapacity.(*atomic.Int32).Load() <= int32(0) {
+			w.Logger().V(5).Infof("skip node(%s) because it has no capacity for disk attachment", node)
+			continue
+		}
 		filtered = append(filtered, node)
-		numFiltered++
 	}
 
 	return filtered, nil
@@ -1564,4 +1569,59 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
 		topologySelector = topologySelector.Add(*topologyRequirement)
 	}
 	return topologySelector
+}
+
+func (c *SharedState) addNodeToAvailableAttachmentsMap(ctx context.Context, nodeName string, nodeLables map[string]string) {
+	if _, ok := c.availableAttachmentsMap.Load(nodeName); !ok {
+		capacity, err := azureutils.GetNodeRemainingDiskCount(ctx, c.cachedClient, nodeName)
+		if err != nil {
+			klog.Errorf("Failed to get node(%s) remaining disk count with error: %v", nodeName, err)
+			// store the maximum capacity if an entry for the node doesn't exist.
+			capacity, err = azureutils.GetNodeMaxDiskCountWithLabels(nodeLables)
+			if err != nil {
+				klog.Errorf("Failed to add node(%s) in availableAttachmentsMap, because get capacity of available attachments is failed with error: %v", nodeName, err)
+				return
+			}
+		}
+		var count atomic.Int32
+		count.Store(int32(capacity))
+		klog.Infof("Added node(%s) to availableAttachmentsMap with capacity: %d", nodeName, capacity)
+		c.availableAttachmentsMap.LoadOrStore(nodeName, &count)
+	}
+}
+
+func (c *SharedState) deleteNodeFromAvailableAttachmentsMap(ctx context.Context, node string) {
+	klog.Infof("Deleted node(%s) from availableAttachmentsMap", node)
+	c.availableAttachmentsMap.Delete(node)
+}
+func (c *SharedState) decrementAttachmentCount(ctx context.Context, node string) bool {
+	remainingCapacity, nodeExists := c.availableAttachmentsMap.Load(node)
+	if nodeExists && remainingCapacity != nil {
+		currentCapacity := int32(0)
+		for {
+			currentCapacity = remainingCapacity.(*atomic.Int32).Load()
+			if currentCapacity == int32(0) || remainingCapacity.(*atomic.Int32).CompareAndSwap(currentCapacity, currentCapacity-1) {
+				if currentCapacity == int32(0) {
+					klog.Errorf("Failed to decrement attachment count for node(%s), because no available attachment", node)
+					return false
+				}
+				return true
+			}
+		}
+	}
+
+	klog.Errorf("Failed to decrement attachment count, because node(%s) not found", node)
+
+	return false
+}
+func (c *SharedState) incrementAttachmentCount(ctx context.Context, node string) bool {
+	remainingCapacity, nodeExists := c.availableAttachmentsMap.Load(node)
+	if nodeExists && remainingCapacity != nil {
+		remainingCapacity.(*atomic.Int32).Add(1)
+		return true
+	}
+
+	klog.Errorf("Failed to increment attachment count, because node(%s) not found", node)
+
+	return false
 }
--- a/pkg/controller/attach_detach_test.go
+++ b/pkg/controller/attach_detach_test.go
@@ -98,6 +98,7 @@ func TestAttachDetachReconcile(t *testing.T) {
 					newAttachment)
 
 				controller.azVolumeAttachmentToVaMap.Store(newAttachment.Name, newVolumeAttachment.Name)
+				addTestNodeInAvailableAttachmentsMap(controller.SharedState, newAttachment.Spec.NodeName, testNodeAvailableAttachmentCount)
 
 				mockClientsAndAttachmentProvisioner(controller)
 
@@ -137,6 +138,7 @@ func TestAttachDetachReconcile(t *testing.T) {
 					newVolumeAttachment,
 					newAttachment)
 
+				addTestNodeInAvailableAttachmentsMap(controller.SharedState, newAttachment.Spec.NodeName, testNodeAvailableAttachmentCount)
 				mockClientsAndAttachmentProvisioner(controller)
 
 				return controller
@@ -168,6 +170,7 @@ func TestAttachDetachReconcile(t *testing.T) {
 					&testAzVolume0,
 					newAttachment)
 
+				addTestNodeInAvailableAttachmentsMap(controller.SharedState, newAttachment.Spec.NodeName, testNodeAvailableAttachmentCount)
 				mockClientsAndAttachmentProvisioner(controller)
 
 				return controller
--- a/pkg/controller/azdrivernode_test.go
+++ b/pkg/controller/azdrivernode_test.go
@@ -18,14 +18,18 @@ package controller
 
 import (
 	"context"
+	"fmt"
+	"reflect"
 	"testing"
 
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2/klogr"
 	azdiskfakes "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/client/clientset/versioned/fake"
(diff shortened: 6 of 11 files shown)


pkg/controller/attach_detach.go
```diff
@@ -113,6 +113,11 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
 		if err := r.removeFinalizer(ctx, azVolumeAttachment); err != nil {
 			return reconcileReturnOnError(ctx, azVolumeAttachment, "delete", err, r.retryInfo)
 		}
+		// deletion of azVolumeAttachment is succeeded, the node's remaining capacity of disk attachment should be increased by 1
+		remainingCapacity, nodeExists := r.SharedState.nodeDiskAvailabilityMap.Load(azVolumeAttachment.Spec.NodeName)
```
edreed, line 117: You can omit the explicit use of `SharedState` since it's an anonymous field, i.e. `r.nodeDiskAvailabilityMap`.

Consider encapsulating the allocation and deallocation of the attachment count in methods of `SharedState`, e.g.

```go
func (s *SharedState) allocateAttachment(node string) error
func (s* SharedState) deallocateAttachment(node string)
```

alice-zheyan-yu: Sure. Will update and fix the failed unit test together.

pkg/controller/attach_detach.go
```diff
@@ -282,6 +287,17 @@ func (r *ReconcileAttachDetach) triggerAttach(ctx context.Context, azVolumeAttac
...
 					}
 				}
+				// the node's remaining capacity of disk attachment should be decreased by 1, since the disk attachment is succeeded.
+				remainingCapacity, nodeExists := r.SharedState.nodeDiskAvailabilityMap.Load(azVolumeAttachment.Spec.NodeName)
+				if nodeExists && remainingCapacity != nil {
+					currentCapacity := int32(0)
+					for {
+						currentCapacity = remainingCapacity.(*atomic.Int32).Load()
+						if currentCapacity == int32(0) || remainingCapacity.(*atomic.Int32).CompareAndSwap(currentCapacity, currentCapacity-1) {
+							break
```
edreed, line 297: We should probably log an error(?) if the remaining capacity was already zero.

pkg/controller/azdrivernode.go
```diff
@@ -101,10 +107,18 @@ func (r *ReconcileAzDriverNode) Recover(ctx context.Context) error {
 		if _, err = r.azClient.DiskV1beta2().AzDriverNodes(r.objectNamespace).Update(ctx, updated, metav1.UpdateOptions{}); err != nil {
 			return err
 		}
+		r.AddNodeInNodeDiskAvailabilityMap(ctx, node.Name, node.Labels)
 	}
 	return nil
 }
 
+func (r *ReconcileAzDriverNode) AddNodeInNodeDiskAvailabilityMap(ctx context.Context, nodeName string, nodeLabels map[string]string) {
```
edreed, line 115: We are not using `nodeLabels`. Why pass them?

pkg/controller/azdrivernode.go
```diff
@@ -101,10 +107,18 @@ func (r *ReconcileAzDriverNode) Recover(ctx context.Context) error {
...
+	capacity, _ := azureutils.GetNodeRemainingDiskCount(ctx, r.SharedState.cachedClient, nodeName)
+	var count atomic.Int32
+	count.Store(int32(capacity))
+	r.SharedState.nodeDiskAvailabilityMap.Store(nodeName, &count)
```
edreed, line 119: Instead of storing unconditionally, it would be safer to only store the maximum capacity if an entry for the node doesn't exist.

pkg/controller/azdrivernode.go
```diff
@@ -70,6 +75,7 @@ func (r *ReconcileAzDriverNode) Reconcile(ctx context.Context, request reconcile
 		if err != nil && !errors.IsNotFound(err) {
 			return reconcile.Result{Requeue: true}, err
 		}
+		r.SharedState.nodeDiskAvailabilityMap.Delete(request.Name)
```
edreed, line 78: For symetry, consider making a `deleteNodeFromDiskAvailabilityMap`

pkg/controller/attach_detach.go
alice-zheyan-yu: Done.

pkg/controller/attach_detach.go
alice-zheyan-yu: Done.

pkg/controller/azdrivernode.go
alice-zheyan-yu: Done.

pkg/controller/azdrivernode.go
alice-zheyan-yu: Done.

pkg/controller/azdrivernode.go
alice-zheyan-yu: Done.

alice-zheyan-yu:
I'm still working on the unit tests. In the meantime, feel free to leave any comments for the current iteration.

pkg/controller/common_test.go
```diff
@@ -751,3 +752,9 @@ func mockClients(mockClient *mockclient.MockClient, azVolumeClient azdisk.Interf
 		}).
 		AnyTimes()
 }
+
+func addTestNodeInAvailableAttachmentsMap(sharedState SharedState, nodeName string) {
```
edreed, line 756: The `sharedState` parameter needs to be a pointer.

pkg/controller/shared_state.go
```diff
@@ -1175,9 +1176,14 @@ func (c *SharedState) createReplicas(ctx context.Context, remainingReplicas int,
 			//Push to queue the failed replica number
 			request := ReplicaRequest{VolumeName: volumeName, Priority: remainingReplicas}
 			c.priorityReplicaRequestsQueue.Push(ctx, &request)
-			return err
+			// don't return err yet, we can still try the next node
+			continue
```
edreed, line 1180: This means that as long as there are nodes we will attempt to create replicas. If all fail (or one fails when the last node is attempted), we will not requeue an attempt to retry (lines 1189..1195 below). I do not believe we should continue on error to create the `AzVolumeAttachment` here.

pkg/controller/shared_state.go
```diff
@@ -1227,16 +1233,17 @@ func (c *SharedState) getNodesForReplica(ctx context.Context, volumeName string,
 	}
 
 	filtered := []string{}
-	numFiltered := 0
 	for _, node := range nodes {
-		if numFiltered >= numReplica {
-			break
-		}
```
edreed, line 1234: Why remove this check? We only need as many nodes as we have requested replicas in the `filtered` list. (That being said, I believe we can rely on the length of the `filtered` list rather than maintain a separate count.

pkg/controller/shared_state.go
edreed: Oh I see, we would requeue an attempt on failure (line 1175..1179 above).

Still, it seems like we could simplify a bit and only decrement `remainingReplicas` on success (exiting early when we reach zero as you've added below). That way, we can rely on `remainingReplicas > 0` to requeue the retry as long as there are more replicas needed. This would eliminate the need to separately requeue on failure to create the `AzVolumeAttachment`.

pkg/controller/shared_state.go
```diff
@@ -1564,3 +1571,60 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
 	}
 	return topologySelector
 }
+
+func (s *SharedState) addNodeToAvailableAttachmentsMap(ctx context.Context, node string) {
```
edreed, line 1575: Where this function is called, it seems like you have the `Node` object already available. You should pass it to this function instead of getting it again.

pkg/controller/shared_state.go
```diff
@@ -1564,3 +1571,60 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+		capacity, _ := azureutils.GetNodeMaxDiskCountWithLabels(nodeObj.Labels)
+		var count atomic.Int32
+		count.Store(int32(capacity))
+		s.availableAttachmentsMap.Store(node, &count)
+		return
+	}
+	capacity, _ := azureutils.GetNodeRemainingDiskCount(ctx, s.cachedClient, node)
+	var count atomic.Int32
+	count.Store(int32(capacity))
+	s.availableAttachmentsMap.Store(node, &count)
```
edreed, line 1587: You should check to see if the entry is already in the map before unconditionally adding it. Although this change may ensure that this isn't the case through the reconciler watch predicate, future changes may break this assumption.

```go
if _, ok := s.availableAttachmentsMap.Load(node); !ok {
    capacity, _ := azureutils.GetNodeRemainingDiskCount(ctx, s.cachedClient, node)
    var count atomic.Int32
    count.Store(int32(capacity))
    _, _ := s.availableAttachmentsMap.LoadOrStore(node, &count)
```

Adding the count using `LoadOrStore` instead of just `Store` handle races where this function is called concurrently for the same node.

pkg/controller/shared_state.go
```diff
@@ -1564,3 +1571,60 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+}
+
+func (s *SharedState) deleteNodeFromAvailableAttachmentsMap(ctx context.Context, node string) {
+	w, _ := workflow.GetWorkflowFromContext(ctx)
+	w.Logger().V(5).Infof("Delete node (%s) from availableAttachmentsMap", node)
```
edreed, line 1592: Rather the log a formatted string, use `Info` and pass the node as a key value pair, e.g.

```go
w.Logger().V(5).Info("Deleted node from availableAttachmentsMap", consts.NodeNameLabel, node)
```

pkg/controller/shared_state.go
```diff
@@ -1564,3 +1571,60 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+	w, _ := workflow.GetWorkflowFromContext(ctx)
+	remainingCapacity, nodeExists := s.availableAttachmentsMap.Load(node)
+	if nodeExists && remainingCapacity != nil {
+		currentCapacity := int32(0)
+		for {
+			currentCapacity = remainingCapacity.(*atomic.Int32).Load()
+			if currentCapacity == int32(0) || remainingCapacity.(*atomic.Int32).CompareAndSwap(currentCapacity, currentCapacity-1) {
+				if currentCapacity == int32(0) {
+					err := status.Error(codes.OutOfRange, fmt.Sprintf("Current Capacity of disk attachment is 0. Can't decrement attachment count anymore."))
+					w.Logger().Errorf(err, "failed to decrement attachment count for node (%s) ", node)
```
edreed, line 1605: Rather the log a formatted string, use `Info` and pass the node as a key value pair, e.g.

```go
w.Logger().Error(err, "Failed to decrement attachment count", consts.NodeNameLabel, node)
```

pkg/controller/shared_state.go
```diff
@@ -1564,3 +1571,60 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+func (s *SharedState) decrementAttachmentCount(ctx context.Context, node string) bool {
+	w, _ := workflow.GetWorkflowFromContext(ctx)
+	remainingCapacity, nodeExists := s.availableAttachmentsMap.Load(node)
+	if nodeExists && remainingCapacity != nil {
+		currentCapacity := int32(0)
+		for {
+			currentCapacity = remainingCapacity.(*atomic.Int32).Load()
+			if currentCapacity == int32(0) || remainingCapacity.(*atomic.Int32).CompareAndSwap(currentCapacity, currentCapacity-1) {
+				if currentCapacity == int32(0) {
+					err := status.Error(codes.OutOfRange, fmt.Sprintf("Current Capacity of disk attachment is 0. Can't decrement attachment count anymore."))
```
edreed, line 1604: You can simply use `errors.New("no more available attachments")` here.

pkg/controller/shared_state.go
```diff
@@ -1564,3 +1571,60 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+					return false
+				} else {
+					return true
+				}
+			}
+		}
+	}
+
+	err := status.Error(codes.FailedPrecondition, fmt.Sprintf("Node does not exist in availableAttachmentsMap."))
+	w.Logger().Errorf(err, "Can't decrement attachment count for the node (%s).", node)
```
edreed, line 1615:
```go
err := errors.New("node not found")
w.Logger().Error(err, "Failed to decrement attachment count", consts.NodeNameLabel, node)
```

pkg/controller/shared_state.go
```diff
@@ -1564,3 +1571,60 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+}
+func (s *SharedState) incrementAttachmentCount(ctx context.Context, node string) bool {
+	w, _ := workflow.GetWorkflowFromContext(ctx)
+	remainingCapacity, nodeExists := s.availableAttachmentsMap.Load(node)
+	if nodeExists && remainingCapacity != nil {
+		remainingCapacity.(*atomic.Int32).Add(1)
+		return true
+	} else {
+		err := status.Error(codes.FailedPrecondition, fmt.Sprintf("The node does not exist in availableAttachmentsMap."))
+		w.Logger().Errorf(err, "failed to increment attachment count for node (%s) ", node)
```
edreed, line 1627: Use same approach recommended above for error reporting.

pkg/controller/common_test.go
alice-zheyan-yu: Done.

pkg/controller/shared_state.go
alice-zheyan-yu: The idea is that getNodesForReplica() should return a full list of available nodes. And before creating replica attachments, it will decrement available attachments count. If the attachment fails(because race condition like the node has no capacity to attach more disks), try next node (here is why we need more nodes than numReplica). Let me know if this works.

pkg/controller/shared_state.go
alice-zheyan-yu: What's the benefit of using Info() rather than Infof()?

edreed: It uses structured logging which is easier to parse for automated analysis.

pkg/controller/shared_state.go
alice-zheyan-yu: Do you mean: 
w.Logger().Errorf(err, "Failed to decrement attachment count", consts.NodeNameLabel, node)

I've checked the implementation of Errorf() and Error(). It seems Error() can only have 2 arguments. Below is the original code:
func (l Logger) Errorf(err error, msg string, params ...interface{}) {
	l.WithCallDepth(1).Error(err, fmt.Sprintf(msg, params...))
}

pkg/controller/shared_state.go
alice-zheyan-yu: But Info() can only have 1 argument (see bottom). You must mean using infof() right?
w.Logger().V(5).Infof("Deleted node from availableAttachmentsMap", consts.NodeNameLabel, node)

```
func (l Logger) Info(msg string) {
	l.WithCallDepth(1).Info(msg)
}
```

edreed: Ah, my bad. I forgot that the workflow logger interface does not implement the additional `keysAndValues` parameters that the [`logr.Logger.Error`](https://pkg.go.dev/github.com/go-logr/logr#Logger.Error) method does.

pkg/controller/shared_state.go
edreed: Ah, my bad. I forgot that the workflow logger interface does not implement the additional `keysAndValues` parameters that the [`logr.Logger.Error`](https://pkg.go.dev/github.com/go-logr/logr#Logger.Error) method does.

pkg/controller/shared_state.go
```diff
@@ -1564,3 +1571,60 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+	return false
+}
+func (s *SharedState) incrementAttachmentCount(ctx context.Context, node string) bool {
+	w, _ := workflow.GetWorkflowFromContext(ctx)
+	remainingCapacity, nodeExists := s.availableAttachmentsMap.Load(node)
+	if nodeExists && remainingCapacity != nil {
+		remainingCapacity.(*atomic.Int32).Add(1)
+		return true
+	} else {
+		err := status.Error(codes.FailedPrecondition, fmt.Sprintf("The node does not exist in availableAttachmentsMap."))
```
edreed, line 1626: Just use `errors.New()` to create the error for logging here.

pkg/controller/shared_state.go
```diff
@@ -1564,3 +1571,60 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+		return
+	}
+	capacity, _ := azureutils.GetNodeRemainingDiskCount(ctx, s.cachedClient, node)
+	var count atomic.Int32
+	count.Store(int32(capacity))
+	s.availableAttachmentsMap.Store(node, &count)
+}
+
+func (s *SharedState) deleteNodeFromAvailableAttachmentsMap(ctx context.Context, node string) {
+	w, _ := workflow.GetWorkflowFromContext(ctx)
```
edreed, line 1591: `deleteNodeFromAvailableAttachmentMap` is mostly called from the `Reconcile` function which doesn't create a workflow. Relying on the temporary workflow that will be returned by `GetWorkflowFromContext` for logging purposes is overkill. You can just use `klog` directly.

pkg/controller/shared_state.go
alice-zheyan-yu: Because the function getNodesForReplica() will return a full list of nodes, so if the attachment fails here, we can still try the next node until remainingReplicas is 0.

pkg/controller/shared_state.go
alice-zheyan-yu: Done.

pkg/controller/shared_state.go
```diff
@@ -1564,3 +1571,60 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
 	}
 	return topologySelector
 }
+
+func (s *SharedState) addNodeToAvailableAttachmentsMap(ctx context.Context, node string) {
+	nodeObj := &v1.Node{}
+	if err := s.cachedClient.Get(ctx, types.NamespacedName{Name: node}, nodeObj); err != nil {
```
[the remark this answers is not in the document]
alice-zheyan-yu: Done.

pkg/controller/shared_state.go
alice-zheyan-yu: Done.

pkg/controller/azdrivernode.go
```diff
@@ -55,8 +55,12 @@ func (r *ReconcileAzDriverNode) Reconcile(ctx context.Context, request reconcile
 	n := &corev1.Node{}
 	err := r.cachedClient.Get(ctx, request.NamespacedName, n)
 
-	// If the node still exists don't delete the AzDriverNode
 	if err == nil {
+		if n.ObjectMeta.DeletionTimestamp == nil {
```
landreasyan, line 59: np: Use `!n.ObjectMeta.DeletionTimestamp.IsZero()`

pkg/controller/azdrivernode.go
```diff
@@ -55,8 +55,12 @@ func (r *ReconcileAzDriverNode) Reconcile(ctx context.Context, request reconcile
 	n := &corev1.Node{}
 	err := r.cachedClient.Get(ctx, request.NamespacedName, n)
 
-	// If the node still exists don't delete the AzDriverNode
 	if err == nil {
```
landreasyan, line 58: If `error != nill` and is not `errors.IsNotFound(err)`, we should probably reconcile because we failed to read from the cacheClient.

alice-zheyan-yu: If `error != nil`l and is not `errors.IsNotFound(err)`, the code will reach the last line of current function:
`return reconcile.Result{Requeue: true}, err`
And `Requeue: true` means it will reconcile which is what we want. Right? Correct me if I misunderstand.

pkg/controller/shared_state.go
```diff
@@ -1565,3 +1573,62 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+
+func (c *SharedState) addNodeToAvailableAttachmentsMap(ctx context.Context, nodeName string, nodeLables map[string]string) {
+	if _, ok := c.availableAttachmentsMap.Load(nodeName); !ok {
+		capacity, err := azureutils.GetNodeRemainingDiskCount(ctx, c.cachedClient, nodeName)
+		if err != nil {
+			klog.Errorf("Failed to get node(%s) remaining disk count", nodeName)
+			// store the maximum capacity if an entry for the node doesn't exist.
+			capacity, err = azureutils.GetNodeMaxDiskCountWithLabels(nodeLables)
+			if err != nil {
+				klog.Errorf("Failed to add node(%s) in availableAttachmentsMap, because get capacity of available attachments is failed", nodeName)
```
alice-zheyan-yu, line 1585: Should log the err return by GetNodeMaxDiskCountWithLabels()

pkg/controller/shared_state.go
```diff
@@ -1565,3 +1573,62 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
 	}
 	return topologySelector
 }
+
+func (c *SharedState) addNodeToAvailableAttachmentsMap(ctx context.Context, nodeName string, nodeLables map[string]string) {
+	if _, ok := c.availableAttachmentsMap.Load(nodeName); !ok {
+		capacity, err := azureutils.GetNodeRemainingDiskCount(ctx, c.cachedClient, nodeName)
+		if err != nil {
+			klog.Errorf("Failed to get node(%s) remaining disk count", nodeName)
```
alice-zheyan-yu, line 1581: Should log the err return by GetNodeRemainingDiskCount()

pkg/controller/shared_state.go
```diff
@@ -1565,3 +1573,62 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+				return
+			}
+		}
+		var count atomic.Int32
+		count.Store(int32(capacity))
+		klog.Infof("Added node(%s) to availableAttachmentsMap with capacity: %d", nodeName, capacity)
+		c.availableAttachmentsMap.LoadOrStore(nodeName, &count)
+	}
+
+	klog.Infof("Node(%s) is already in availableAttachmentsMap. Don't need to add it again", nodeName)
```
alice-zheyan-yu, line 1595: Please also log the current capacity to make debug easier.

pkg/controller/azdrivernode.go
alice-zheyan-yu: Done.

pkg/controller/shared_state.go
alice-zheyan-yu: Done.

pkg/controller/shared_state.go
alice-zheyan-yu: Done.

pkg/controller/shared_state.go
```diff
@@ -1176,9 +1178,14 @@ func (c *SharedState) createReplicas(ctx context.Context, remainingReplicas int,
 			//Push to queue the failed replica number
 			request := ReplicaRequest{VolumeName: volumeName, Priority: remainingReplicas}
 			c.priorityReplicaRequestsQueue.Push(ctx, &request)
```
edreed, line 1180: If we exit the loop and `remainingReplicas > 0`, a `ReplicaRequest` will be pushed to the operation queue on line 1194 below, so we do not need to do this here.

pkg/controller/shared_state.go
```diff
@@ -1228,16 +1235,17 @@ func (c *SharedState) getNodesForReplica(ctx context.Context, volumeName string,
...
-		if numFiltered >= numReplica {
-			break
-		}
 		if skipSet[node] {
 			continue
 		}
+		// if the node has no capacity for disk attachment, we should skip it
+		remainingCapacity, nodeExists := c.availableAttachmentsMap.Load(node)
+		if !nodeExists || remainingCapacity == nil || remainingCapacity.(*atomic.Int32).Load() <= int32(0) {
+			w.Logger().Infof("skip node(%s) because it has no capacity for disk attachment", node)
```
edreed, line 1245: This log could be quite chatty when near full attachment capacity. I would recommend a higher logging level here, e.g. `V(5)`.

pkg/controller/shared_state.go
```diff
@@ -1565,3 +1573,62 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+				return
+			}
+		}
+		var count atomic.Int32
+		count.Store(int32(capacity))
+		klog.Infof("Added node(%s) to availableAttachmentsMap with capacity: %d", nodeName, capacity)
+		c.availableAttachmentsMap.LoadOrStore(nodeName, &count)
+	}
+
+	klog.Infof("Node(%s) is already in availableAttachmentsMap. Don't need to add it again", nodeName)
```
edreed, line 1595: This log isn't all that useful for debugging and can be removed.

pkg/controller/shared_state.go
```diff
@@ -1565,3 +1573,62 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+	if nodeExists && remainingCapacity != nil {
+		currentCapacity := int32(0)
+		for {
+			currentCapacity = remainingCapacity.(*atomic.Int32).Load()
+			if currentCapacity == int32(0) || remainingCapacity.(*atomic.Int32).CompareAndSwap(currentCapacity, currentCapacity-1) {
+				if currentCapacity == int32(0) {
+					klog.Errorf("Failed to decrement attachment count for node(%s), because no available attachment", node)
+					return false
+				}
+				klog.Infof("Decrement attachment count is succeeded for node(%s) with new capacity: %d", node, currentCapacity-1)
```
edreed, line 1613: This log isn't all that useful and can be removed.

pkg/controller/shared_state.go
```diff
@@ -1565,3 +1573,62 @@ func (c *SharedState) getNodesTopologySelector(ctx context.Context, nodes []v1.N
...
+
+	klog.Errorf("Failed to decrement attachment count, because node(%s) not found", node)
+
+	return false
+}
+func (c *SharedState) incrementAttachmentCount(ctx context.Context, node string) bool {
+	remainingCapacity, nodeExists := c.availableAttachmentsMap.Load(node)
+	if nodeExists && remainingCapacity != nil {
+		remainingCapacity.(*atomic.Int32).Add(1)
+		klog.Infof("Increment attachment count is succeeded for node(%s)", node)
```
edreed, line 1627: This log isn't all that useful and can be removed.

pkg/controller/shared_state.go
alice-zheyan-yu: Sure. I saw that in our code base we have V(2) V(5) V(10). Does 10 mean the most important log? How to use this level to help debug?

pkg/controller/shared_state.go
alice-zheyan-yu: Done.

pkg/controller/shared_state.go
alice-zheyan-yu: Done.

pkg/controller/shared_state.go
alice-zheyan-yu: Done.

pkg/controller/shared_state.go
alice-zheyan-yu: Done.

pkg/controller/azdrivernode.go
landreasyan: Oh, you are right! I was reviewing on github and missed the end of the method.

pkg/controller/shared_state.go
alice-zheyan-yu: Done. Yes, we shouldn't push the failed replica to the queue here, since we will keep trying next node. I missed it here and make the code so confused. Now the logic is clean!

pkg/controller/shared_state.go
edreed: Lower values in verbosity level are generally more important. We have to ensure that we log enough information to be useful in production without overwhelming the logs. This is important because storage is limited, and we must support many customers and clusters. For production, we generally use verbosity level 2 or 3. For testing, we will use higher levels because the additional logging helps analyze test failures.

PR #1599, opened 2022-11-15
[V2] fix: handle pre-provisioned volume deletion at the right time

sunpa93:
**What type of PR is this?**

/kind cleanup

* Current PV controller issues a delete request for `AzVolume` CRI upon seeing a deletionTimestamp for `PV` but this can be problematic when user issues `ControllerUnpublishVolume` after marking the preprovisioned PV with a deletion timestamp, as 
`AzVolume` may no longer be able to be found in cluster when trying to unpublish the volume, ultimately resulting in detach failures.
* Also, when deleting `AzVolume` for pre-provisioned volume, `PV controller` needs to be able to retrigger a failed deletion.
* The PR addresses these issue.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/pv.go
+++ b/pkg/controller/pv.go
@@ -71,11 +71,19 @@ func (r *ReconcilePV) Reconcile(ctx context.Context, request reconcile.Request) 
 	// Ignore not found errors as they cannot be fixed by a requeue
 	if err := r.cachedClient.Get(ctx, request.NamespacedName, &pv); err != nil {
 		if apierrors.IsNotFound(err) {
-			r.deletePV(request.Name)
+			//nolint:contextcheck // deletePV is asynchronous; context is not inherited by design
+			if err = r.deletePV(request.Name); err != nil {
+				return reconcileReturnOnError(ctx, &corev1.PersistentVolume{ObjectMeta: metav1.ObjectMeta{Name: request.Name}}, "delete", err, r.controllerRetryInfo)
+			}
 			return reconcileReturnOnSuccess(request.Name, r.controllerRetryInfo)
 		}
 		logger.Error(err, "failed to get PV")
 		return reconcileReturnOnError(ctx, &pv, "get", err, r.controllerRetryInfo)
+	}
+
+	// no need to reconcile further if pv is marked with a deletionTimestamp
+	if deleteRequested, _ := objectDeletionRequested(&pv); deleteRequested {
+		return reconcileReturnOnSuccess(pv.Name, r.controllerRetryInfo)
 	}
 
 	// migration case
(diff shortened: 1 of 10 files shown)


pkg/controller/pv.go
```diff
@@ -100,60 +103,27 @@ func (r *ReconcilePV) Reconcile(ctx context.Context, request reconcile.Request)
...
-		}
-		// deletion timestamp is set and AzVolume reconcliler will handle the delete request.
-		// For pre-provisioned case when only PV is deleted, we will only be deleting the CRI as VolumeDeleteRequestAnnotation will not be set.
-		// The volume itself will not be deleted.
-		logger.Info("Deletion timestamp for AzVolume set")
-		return reconcileReturnOnSuccess(pv.Name, r.controllerRetryInfo)
-	}
-
 	// PV exists but AzVolume doesn't
+	deleteRequested, _ := objectDeletionRequested(&pv)
```
edreed, line 107: It doesn't appear we are using `deleteRequested` other than in the following `if` statement scope. Can we move the check to the point of use (i.e. line 115 below)?

pkg/controller/shared_state.go
```diff
@@ -555,8 +556,80 @@ func (c *SharedState) addVolumeAndClaim(azVolumeName, pvName, pvClaimName string
...
+	if err != nil {
+		if apiErrors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+	klog.Info("azVolume updated")
+
+	go func() {
+		_ = wait.PollImmediateInfinite(consts.CRIUpdateRetryDuration, func() (bool, error) {
```
edreed, line 612: Can we use a condition waiter instead of a polling wait? Polling impacts API server throttling.

pkg/controller/pv.go
landreasyan, line 107: If delete is requested, we should return a success on reconciliation and not proceed with getting the azvolume.

pkg/controller/shared_state.go
```diff
@@ -555,8 +555,97 @@ func (c *SharedState) addVolumeAndClaim(azVolumeName, pvName, pvClaimName string
...
+	azVolume := azVolumeList.Items[0]
+	w.AddDetailToLogger(workflow.GetObjectDetails(&azVolume)...)
+
+	if azureutils.MapContains(azVolume.Status.Annotations, consts.PreProvisionedVolumeAnnotation) {
+		return nil
+	}
+
+	// AzVolume does exist and needs to be deleted
+	// add annotation to mark AzVolumeAttachment cleanup
+	err = c.cachedClient.Delete(ctx, &azVolume)
```
edreed, line 591: Shouldn't we add the cleanup annotation before deleting the `AzVolume`? If the `azvolume` controller sees the delete  *before* the annotation is added, won't it try to delete the pre-provisioned disk?

pkg/controller/shared_state.go
```diff
@@ -555,8 +555,97 @@ func (c *SharedState) addVolumeAndClaim(azVolumeName, pvName, pvClaimName string
...
+			azVolume := obj.(*azdiskv1beta2.AzVolume)
+
+			if azVolume.Status.State == azdiskv1beta2.VolumeDeletionFailed || azVolume.Status.Error != nil {
+				updateFunc := func(obj client.Object) error {
+					azVolume := obj.(*azdiskv1beta2.AzVolume)
+					azVolume.Status.Error = nil
+					azVolume.Status.State = azdiskv1beta2.VolumeCreated
+					return nil
+				}
+				_, _ = azureutils.UpdateCRIWithRetry(ctx, nil, c.cachedClient, c.azClient, azVolume, updateFunc, consts.ForcedUpdateMaxNetRetry, azureutils.UpdateCRIStatus)
```
edreed, line 643: Use `goCtx`?

pkg/controller/shared_state.go
sunpa93: The azVolume controller would see that an `AzVolume` deletion was requested without a `volumeDetachRequestAnnotation`, in which case it would consider it a uninstall scenario and detach replicas without deleting the volume.

But good catch. The annotations need to be added before issuing a delete request. I was confusing upcoming Huichan's change for removing the cleanup annotation for pre-provisioned volumes.

PR #1603, opened 2022-11-16
[V2] fix: updating verify checks for helm tests.

landreasyan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/hack/verify-helm-chart-files.sh
+++ b/hack/verify-helm-chart-files.sh
@@ -80,6 +80,10 @@ if [[ -z "$(command -v helm)" ]]; then
   install_helm
 fi
 
-helm repo add azuredisk-csi-driver https://raw.githubusercontent.com/kubernetes-sigs/azuredisk-csi-driver/master/charts
-helm search repo -l azuredisk-csi-driver
+helm repo add azuredisk-csi-driver-v1 https://raw.githubusercontent.com/kubernetes-sigs/azuredisk-csi-driver/master/charts
+helm search repo -l azuredisk-csi-driver-v1
+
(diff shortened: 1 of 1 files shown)


hack/verify-helm-chart-files.sh
```diff
@@ -80,6 +80,10 @@ if [[ -z "$(command -v helm)" ]]; then
   install_helm
 fi
 
-helm repo add azuredisk-csi-driver https://raw.githubusercontent.com/kubernetes-sigs/azuredisk-csi-driver/master/charts
-helm search repo -l azuredisk-csi-driver
+helm repo add azuredisk-csi-driver-v1 https://raw.githubusercontent.com/kubernetes-sigs/azuredisk-csi-driver/master/charts
```
edreed, line 83: Rather than add both the `master` and `main_v2` branch repositories, we should only check the current branch. You can get the current branch using `git branch --show-current`.

We should also delete the Helm repository when done to make this script work repeatedly on developers' machines w/o having to manually clean up.

landreasyan: If we want to delete repo afterwards, we have to make sure it is added under a name that is not currently in use. Because this command will work repeatedly as long as we add the same repo under the same name.

edreed: We could append the Git commit to the repo name to make it unique.

landreasyan: Yeah, that sounds great!

PR #1607, opened 2022-11-18
[V2] refactor: start informers in one place

hccheng72:
**What type of PR is this?**
/kind cleanup

Start informers centrally and share them with crdprovisioner and controllers to prevent dependency.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azuredisk_v2.go
+++ b/pkg/azuredisk/azuredisk_v2.go
@@ -38,6 +38,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apiRuntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/informers"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
@@ -46,6 +47,7 @@ import (
 	"k8s.io/klog/v2/klogr"
 
 	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	crdInformers "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
 	clientmetrics "k8s.io/client-go/tools/metrics"
@@ -63,6 +65,7 @@ import (
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/optimization"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/provisioner"
 	volumehelper "sigs.k8s.io/azuredisk-csi-driver/pkg/util"
+	"sigs.k8s.io/azuredisk-csi-driver/pkg/watcher"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/workflow"
 	azurecloudconsts "sigs.k8s.io/cloud-provider-azure/pkg/consts"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -117,6 +120,7 @@ type DriverV2 struct {
 	kubeConfig           *rest.Config
 	kubeClient           *clientset.Clientset
 	azdiskClient         azdisk.Interface
+	crdClient            crdClientset.Interface
 	azDriverNodeInformer azdiskinformertypes.AzDriverNodeInformer
 	deviceChecker        *deviceChecker
 }
@@ -179,12 +183,28 @@ func (d *DriverV2) Run(endpoint, kubeconfig string, disableAVSetNodes, testingMo
 		klog.Fatalf("failed to get azdiskclient with kubeconfig (%s), error: %v. Exiting application...", kubeconfig, err)
 	}
 
+	d.crdClient, err = crdClientset.NewForConfig(d.kubeConfig)
+	if err != nil {
+		klog.Fatalf("failed to get crdclient with kubeconfig (%s), error: %v. Exiting application...", kubeconfig, err)
+	}
+
 	// d.crdProvisioner is set by NewFakeDriver for unit tests.
 	if d.crdProvisioner == nil {
-		d.crdProvisioner, err = provisioner.NewCrdProvisioner(d.kubeConfig, d.config.ObjectNamespace)
-		if err != nil {
-			klog.Fatalf("Failed to get crd provisioner. Error: %v", err)
-		}
+		kubeInformerFactory := informers.NewSharedInformerFactory(d.kubeClient, consts.DefaultInformerResync)
+		azInformerFactory := azdiskinformers.NewSharedInformerFactory(d.azdiskClient, consts.DefaultInformerResync)
+		crdInformerFactory := crdInformers.NewSharedInformerFactory(d.crdClient, consts.DefaultInformerResync)
+
+		nodeInformer := azureutils.NewNodeInformer(kubeInformerFactory)
+		azNodeInformer := azureutils.NewAzNodeInformer(azInformerFactory)
+		azVolumeAttachmentInformer := azureutils.NewAzVolumeAttachmentInformer(azInformerFactory)
+		azVolumeInformer := azureutils.NewAzVolumeInformer(azInformerFactory)
+		crdInformer := azureutils.NewCrdInformer(crdInformerFactory)
+
+		conditionWatcher := watcher.NewConditionWatcher(azInformerFactory, d.config.ObjectNamespace, azNodeInformer, azVolumeAttachmentInformer, azVolumeInformer)
+
+		d.crdProvisioner = provisioner.NewCrdProvisioner(d.azdiskClient, d.config.ObjectNamespace, conditionWatcher, provisioner.NewCachedReader(kubeInformerFactory, azInformerFactory, d.config.ObjectNamespace), crdInformer)
+
+		azureutils.StartInformersAndWaitForCacheSync(context.Background(), nodeInformer, azNodeInformer, azVolumeAttachmentInformer, azVolumeInformer, crdInformer)
 	}
 
 	// d.cloudProvisioner is set by NewFakeDriver for unit tests.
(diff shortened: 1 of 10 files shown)


pkg/azuredisk/azuredisk_v2.go
```diff
@@ -180,10 +183,11 @@ func (d *DriverV2) Run(endpoint, kubeconfig string, disableAVSetNodes, testingMo
...
 	if d.crdProvisioner == nil {
-		d.crdProvisioner, err = provisioner.NewCrdProvisioner(d.kubeConfig, d.config.ObjectNamespace)
-		if err != nil {
-			klog.Fatalf("Failed to get crd provisioner. Error: %v", err)
-		}
+		kubeInformerFactory := informers.NewSharedInformerFactory(d.kubeClient, consts.DefaultInformerResync)
+		azInformerFactory := azdiskinformers.NewSharedInformerFactory(d.azdiskClient, consts.DefaultInformerResync)
+		d.conditionWatcher = watcher.NewConditionWatcher(azInformerFactory, d.config.ObjectNamespace)
+		d.crdProvisioner = provisioner.NewCrdProvisioner(d.azdiskClient, d.config.ObjectNamespace, d.conditionWatcher, kubeInformerFactory, azInformerFactory)
+		d.crdProvisioner.AddPredicatesAndStartInformers(context.Background(), kubeInformerFactory, azInformerFactory, d.conditionWatcher)
```
edreed, line 190: Architecturally, it's not a great idea for one object to rely on the side-effect of initializing another. Here, `ConditionWatcher` relies on `CrdProvisioner` to register its handlers and synchronize the informers. It makes sense to create the informers to be passed to `NewConditionWatcher` and `NewCrdProvisioner`, each adding its respective event handlers as necessary. Then have a global function that handles starting and synchronizing the informers, e.g.

```go
func StartAndWaitForCacheSync(ctx context.Context, informers... cache.SharedInformer)
```

This function can be called after all of the informers, `ConditionWatcher` and `CrdProvisioner` have been created.

pkg/provisioner/crdprovisioner.go
```diff
@@ -1217,14 +1201,33 @@ func waitForDetachVolumeFunc(obj interface{}, objectDeleted bool) (bool, error)
 	return false, nil
 }
 
-func startWatchingKube(ctx context.Context, kubeInformerFactory informers.SharedInformerFactory) {
+func (c *CrdProvisioner) AddPredicatesAndStartInformers(ctx context.Context, kubeInformerFactory informers.SharedInformerFactory, azInformerFactory azdiskinformers.SharedInformerFactory, cw *watcher.ConditionWatcher) {
```
landreasyan, line 1204: I don't think `AddPredicatesAndStartInformers` should be a member method of the `CrdProvisioner` as there is no direct dependency between the two. I think we can make it an independent function in one of the util or common files.

pkg/provisioner/crdprovisioner.go
```diff
@@ -166,31 +164,17 @@ func (a CachedReader) List(_ context.Context, objectList client.ObjectList, opts
...
-	if err != nil {
-		return nil, err
-	}
-
-	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, consts.DefaultInformerResync)
-	startWatchingKube(context.Background(), kubeInformerFactory)
-
-	azInformerFactory := azdiskinformers.NewSharedInformerFactory(azdiskClient, consts.DefaultInformerResync)
-
+func NewCrdProvisioner(azdiskClient azdisk.Interface, objNamespace string, cw *watcher.ConditionWatcher, kubeInformerFactory informers.SharedInformerFactory, azInformerFactory azdiskinformers.SharedInformerFactory) *CrdProvisioner {
```
landreasyan, line 167: We could directly pass the `CacheReader1 instead of each parameter.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -461,7 +465,7 @@ func (d *DriverV2) registerAzDriverNodeOrDie(ctx context.Context) {
 
 	nodeSelector := fmt.Sprintf("metadata.name=%s", d.NodeID)
 
-	azdiskInformer := azdiskinformers.NewSharedInformerFactoryWithOptions(
+	azdiskInformerFactory := azdiskinformers.NewSharedInformerFactoryWithOptions(
```
landreasyan, line 468: Since this method starts a member informer of `DriverV2`, I am wondering if we should refactor and start it at the beginning of driver setup with the rest of the informers if `d.config.NodeConfig.Enabled`

pkg/azuredisk/azuredisk_v2.go
hccheng72: I found SharedInformerFactory could be from different packages (but not cache), it seems I can't make this function that global. Given that, do we still want to separate `WaitForCacheSync` into a function?

pkg/provisioner/crdprovisioner.go
hccheng72: Do you mean to create a `CacheReader` object in `azuredisk_v2.go` and pass to this function?

pkg/azuredisk/azuredisk_v2.go
edreed: Ah, yes you are right. However, this could be resolved using the following interface definition instead of `cache.SharedInformerFactory`:

```go
type InformerFactory interface {
        Start(stopCh <-chan struct{})
	Informer() cache.SharedIndexInformer
}
```

This should allow arbitrary `SharedInformerFactory` types to be passed to the method.

edreed: Passing the `HasSynced` method to `cache.WaitForCacheSync` is a little complicated due to the indirection through the `Informer` method. You will need to iterate over the `informers` variadic argument to build an array of `cache.InformerSynced` functions which then gets passed to `cache.WaitForCacheSync`, e.g.

```go
informerSynced := make([]InformerSynced, 0)
for _, factory := range informers {
    informerSynced = append(informerSynced, factory.Informer().HasSynced)
}

cache.WaitForCacheSync(ctx, informers...)
```

pkg/azureutils/informer_common.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (i crdInformer) Start(stopCh <-chan struct{}) {
+	i.factory.Start(stopCh)
+}
+
+func (i crdInformer) Informer() cache.SharedIndexInformer {
+	return i.informer
+}
+
+func StartAndWaitForCacheSync(ctx context.Context, factoris ...InformerFactory) {
```
edreed, line 134: sp: `factories`

pkg/azureutils/informer_common.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func NewAzNodeInformer(factory azdiskinformers.SharedInformerFactory) InformerFactory {
+	return &azNodeInformer{
+		factory:  factory,
+		informer: factory.Disk().V1beta2().AzDriverNodes().Informer(),
+	}
+}
+
+func (a azNodeInformer) Start(stopCh <-chan struct{}) {
```
alice-zheyan-yu, line 46: I'm not sure should we use `*azNodeInformer` here, like: `func (a *azNodeInformer) Start(stopCh <-chan struct{}) {`

edreed: Good catch! Yes, the receivers should be pointers instead of by value.

pkg/azuredisk/azuredisk_v2.go
hccheng72: I don't include it to this refactoring for now. Since this informer has an additional need for `Lister()` method [here](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/10e86f03596907ec945131c1eb4d17bc9e32b74f/pkg/azuredisk/azuredisk_v2.go#L561), I don't see many benefits to combine it with other informers. We could reconsider if we have a better idea.

pkg/azureutils/informer_common.go
```diff
@@ -0,0 +1,145 @@
...
+	"context"
+
+	crdInformers "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+	azdiskinformers "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/client/informers/externalversions"
+)
+
+type InformerFactory interface {
```
landreasyan, line 29: This interface does not really describe a factory, I think we should rename it tp `GenericInformer`

pkg/azureutils/informer_common.go
```diff
@@ -0,0 +1,145 @@
...
+	Start(stopCh <-chan struct{})
+	Informer() cache.SharedIndexInformer
+}
+
+type azNodeInformer struct {
```
landreasyan, line 34: We since `azNodeInformer`, `azVolumeAttachmentInformer` and `azVolumeInformer` essentially have the same type definition and the same implementation of the interface, we should define one type, say `GenericInformerInfo` and implement the interface for that type. We will later initialize `azNodeInformer`, `azVolumeAttachmentInformer` and `azVolumeInformer` as `GenericInformerInfo`.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -475,7 +481,7 @@ func (d *DriverV2) registerAzDriverNodeOrDie(ctx context.Context) {
 		},
 	})
 
-	go azdiskInformer.Start(ctx.Done())
+	azdiskInformerFactory.Start(ctx.Done())
```
landreasyan, line 484: We should utilize `azureutils.StartAndWaitForCacheSync` instead of manually starting the informer here.

pkg/azureutils/informer_common.go
```diff
@@ -0,0 +1,145 @@
...
+
+func NewCrdInformer(factory crdInformers.SharedInformerFactory) InformerFactory {
+	return &crdInformer{
+		factory:  factory,
+		informer: factory.Apiextensions().V1().CustomResourceDefinitions().Informer(),
+	}
+}
+
+func (i *crdInformer) Start(stopCh <-chan struct{}) {
+	i.factory.Start(stopCh)
```
landreasyan, line 127: I think we should either start it in a goroutine here, or return `i.factory.Start(stopCh)` and start in a goroutine where it is called.

hccheng72: `Start` method already starts each informer in a go routine [here](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/91e9fed829ea6447979156004bd4174b7f51f0d5/pkg/apis/client/informers/externalversions/factory.go#L117), do we still run this method in another go routine?

edreed: There's no need to call `Start` in a goroutine since it creates one in its implementation.

pkg/azureutils/informer_common.go
hccheng72: You're right, there's no need to distinct them.

pkg/azuredisk/azuredisk_v2.go
hccheng72: If we want to use `azureutils.StartAndWaitForCacheSync`, we need to create an object as GenericInformer interface. But this informer needs `Lister()` method [here](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/9356028355480bb971ea4bd811ae98b5d3498f2f/pkg/azuredisk/azuredisk_v2.go#L555), we need to add this method/field to GenericInformer interface and GenericAzInformerInfo struct that would cause all other informers need to initiate `lister` field which won't be used. That's why I manually start this informer. Any other suggestion?

pkg/azureutils/informer_common.go
hccheng72: Thanks for confirming.

Issue #1608, opened 2022-11-18
CSI Driver should support E2E tests for VMSS Flex


zmyzheng:
**Is your feature request related to a problem?/Why is this needed**

We should add a new PR presubmit pipeline for E2E tests in VMSS Flex cluster

**Describe the solution you'd like in detail**

- use capz to create the VMSS Flex cluster
- install out-of-tree CCM/CNM
- install CSI Driver
- run CSI E2E tests

**Describe alternatives you've considered**

**Additional context**

PR #1609, opened 2022-11-19
[chore] Fix E2E failure in VMSS Flex cluster

zmyzheng:
**What type of PR is this?**
/kind bug
/kind test

When TEST_CCM is enabled, CCM will be built with [ci-build-azure-ccm.sh](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/scripts/ci-build-azure-ccm.sh). However, this will cause the env variable naming conflict: env variable IMAGE_TAG is used by this file for CCM tag at line 46:https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/scripts/ci-build-azure-ccm.sh#L46, and at the same time it is also used by the CSI Driver make file with a different meaning for the CSI tag: https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/Makefile#L37. Because of that, the CSI image generated will have the wrong image URI and thus cannot be pushed to the ACR with authentication failure. This PR gives the CSI image with a different env variable for the tag to fix the issue.

Refer to #1608 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

Please merge PR https://github.com/kubernetes/test-infra/pull/28057 first before testing this PR

**Release note**:
```
none
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ ifndef PUBLISH
 override IMAGE_VERSION := $(IMAGE_VERSION)-$(GIT_COMMIT)
 endif
 endif
-IMAGE_TAG ?= $(REGISTRY)/$(IMAGE_NAME):$(IMAGE_VERSION)
+CSI_IMAGE_TAG ?= $(REGISTRY)/$(IMAGE_NAME):$(IMAGE_VERSION)
 IMAGE_TAG_LATEST = $(REGISTRY)/$(IMAGE_NAME):latest
 REV = $(shell git describe --long --tags --dirty)
 BUILD_DATE ?= $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -34,7 +34,7 @@ ifndef PUBLISH
 override IMAGE_VERSION := $(IMAGE_VERSION)-$(GIT_COMMIT)
 endif
 endif
-IMAGE_TAG ?= $(REGISTRY)/$(IMAGE_NAME):$(IMAGE_VERSION)
+CSI_IMAGE_TAG ?= $(REGISTRY)/$(IMAGE_NAME):$(IMAGE_VERSION)
 IMAGE_TAG_LATEST = $(REGISTRY)/$(IMAGE_NAME):latest
```
edreed, line 38: Should we also rename `IMAGE_TAG_LATEST` for consistency, i.e. `CSI_IMAGE_TAG_LATEST`?

zmyzheng: Sure. Modified it just now.

andyzhangx (approved):
/lgtm
/approve

PR #1612, opened 2022-11-22
[V2] refactor: combine azdrivernode and node_availability into one controller

### LINKED ISSUE #1591: [V2] Combine `AzDriverNodeController` and `NodeAvailabilityController`

* `AzDriverNodeController` and `NodeAvailabilityController` botch watches the same objects, so it makes sense to combine them into a single controller.

**Additional context**

alice-zheyan-yu:
**What type of PR is this?**

/kind cleanup
refactor: combine azdrivernode and node_availability into a single controller.

Fixes #1591 

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


edreed:
To prevent confusion over what is being watched, we should rename `azdrivernode.go` to `node.go` and `ReconcileAzDriverNode` to `ReconcileNode`.

alice-zheyan-yu:
> To prevent confusion over what is being watched, we should rename `azdrivernode.go` to `node.go` and `ReconcileAzDriverNode` to `ReconcileNode`.

alice-zheyan-yu:
> 

Done

pkg/controller/node.go
```diff
@@ -38,29 +38,36 @@ import (
...
-func (r *ReconcileAzDriverNode) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+func (r *ReconcileNode) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
 	if !r.isRecoveryComplete() {
 		return reconcile.Result{Requeue: true}, nil
 	}
 
 	n := &corev1.Node{}
 	err := r.cachedClient.Get(ctx, request.NamespacedName, n)
 
+	logger := r.logger.WithValues(consts.NodeNameLabel, request.Name)
```
edreed, line 58: Try to avoid inserting statements between a function that returns an error and the `if` statement that checks the result. The declaration and assignment of `logger` should be moved above line 55.

pkg/controller/node.go
```diff
@@ -38,29 +38,35 @@ import (
...
+func (r *ReconcileNode) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
 	if !r.isRecoveryComplete() {
 		return reconcile.Result{Requeue: true}, nil
 	}
+	logger := r.logger.WithValues(consts.NodeNameLabel, request.Name)
 
 	n := &corev1.Node{}
 	err := r.cachedClient.Get(ctx, request.NamespacedName, n)
 
 	if err == nil {
```
landreasyan, line 59: np: could we reorder and check in the following order so that we don't have to reconcile in the "default" case: 

1. errors.IsNotFound(err)  ->  ...delete azdrivernode...
2. error != nil -> reconcile.Result{Requeue: true}, err
3. if n.ObjectMeta.DeletionTimestamp.IsZero() && !n.Spec.Unschedulable -> add to map and create replicas
4. default -> return reconcile.Result{}, nil

alice-zheyan-yu: Love it! Done.

Issue #1590, opened 2022-11-10
[V2] AzDriverNodeController can miss Node DELETE events.


sunpa93:
**What happened**:
* When the driver is installed on an agent node, and the cluster is scaled down, the driver pods will terminate and may not be able to observe the DELETE event of the nodes, in which case the `AzDriverNodeController` will not be able to delete the `AzDriverNode` accordingly leading to orphaned `AzDriverNode` objects. 

**What you expected to happen**:
* `AzDriverNodeController.Recover` should delete `AzDriverNode` CRIs for nodes that have been deleted whenever controller restarts so that objects can be cleaned up properly.

**How to reproduce it**:
* Install driver on an vmss agentpool and scale down the cluster.

**Anything else we need to know?**:

**Environment**:

- CSI Driver version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #1615, opened 2022-11-22
[V2] fix: delete orphaned azdrivernode and update error logs in Recover function

hccheng72:
**What type of PR is this?**
/kind bug

1. delete orphaned azdrivernode when the cluster scales in.
2. change the way to return error in `Recover` function.
3. fix a typo in `azuredisk.go`

#1590 #1617

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azuredisk_v2.go
+++ b/pkg/azuredisk/azuredisk_v2.go
@@ -411,7 +411,7 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
 		<-mgr.Elected()
 		var errors []error
 		ctx, w := workflow.New(ctx)
-		defer func() { w.Finish(err) }()
+		defer func() { w.Finish(errors...) }()
 		// recover lost states if necessary
 		w.Logger().Infof("Elected as leader; initiating CRI deperecation / recovery...")
 		if err := nodeReconciler.Recover(ctx); err != nil {
--- a/pkg/controller/node.go
+++ b/pkg/controller/node.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -29,6 +30,7 @@ import (
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/workflow"
 
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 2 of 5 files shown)


pkg/controller/azdrivernode.go
```diff
@@ -101,12 +101,21 @@ func (r *ReconcileAzDriverNode) Recover(ctx context.Context) error {
...
-		updated := node.DeepCopy()
-		updated.Annotations = azureutils.AddToMap(updated.Annotations, consts.RecoverAnnotation, "azDriverNode")
-		if _, err = r.azClient.DiskV1beta2().AzDriverNodes(r.objectNamespace).Update(ctx, updated, metav1.UpdateOptions{}); err != nil {
-			return err
+		// if the corresponding node has been deleted, delete the azdriverode object
+		_, err := r.kubeClient.CoreV1().Nodes().Get(ctx, node.Spec.NodeName, metav1.GetOptions{})
+		if err != nil && errors.IsNotFound(err) {
+			azN := r.azClient.DiskV1beta2().AzDriverNodes(r.objectNamespace)
+			if err = azN.Delete(ctx, node.Name, metav1.DeleteOptions{}); err != nil {
+				return nil
```
alice-zheyan-yu, line 109: If delete current node is failed, it `return nil` immediately here in the for-loop. What about other nodes in the loop? I think we can continue to recover other nodes.

pkg/controller/azdrivernode.go
```diff
@@ -101,12 +101,21 @@ func (r *ReconcileAzDriverNode) Recover(ctx context.Context) error {
...
+			}
+		} else {
+			updated := node.DeepCopy()
+			updated.Annotations = azureutils.AddToMap(updated.Annotations, consts.RecoverAnnotation, "azDriverNode")
+			if _, err = r.azClient.DiskV1beta2().AzDriverNodes(r.objectNamespace).Update(ctx, updated, metav1.UpdateOptions{}); err != nil {
+				return err
```
alice-zheyan-yu, line 115: Same concern here: if update current node fails, it `return err` immediately. I feel like we can log the err and continue to recover other nodes.

hccheng72: Good point! Since this isn't new code and I also see same logic used in pv controller, we can raise it in the meeting for a discussion. If it should be changed, it would be better to do it in another PR.

pkg/controller/azdrivernode.go
edreed: We should log the error and continue rather than return.

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -417,7 +417,7 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
 		<-mgr.Elected()
 		var errors []error
 		ctx, w := workflow.New(ctx)
-		defer func() { w.Finish(err) }()
+		defer func() { w.Finish(errors...) }()
```
edreed, line 420: Good catch!

pkg/controller/azdrivernode.go
```diff
@@ -88,25 +91,40 @@ func (r *ReconcileAzDriverNode) Reconcile(ctx context.Context, request reconcile
...
-		}
 		return err
 	}
 
 	for _, node := range nodes.Items {
-		updated := node.DeepCopy()
-		updated.Annotations = azureutils.AddToMap(updated.Annotations, consts.RecoverAnnotation, "azDriverNode")
-		if _, err = r.azClient.DiskV1beta2().AzDriverNodes(r.objectNamespace).Update(ctx, updated, metav1.UpdateOptions{}); err != nil {
-			return err
+		// if the corresponding node has been deleted, delete the azdriverode object
```
edreed, line 105: sp: `azdrivernode`

pkg/controller/azdrivernode.go
```diff
@@ -88,25 +91,40 @@ func (r *ReconcileAzDriverNode) Reconcile(ctx context.Context, request reconcile
...
+		_, err := r.kubeClient.CoreV1().Nodes().Get(ctx, node.Spec.NodeName, metav1.GetOptions{})
+		if err != nil && errors.IsNotFound(err) {
```
edreed, line 107: If `err != nil` but isn't a "not found" error, we'll end up in the else clause and update the `AzDriverNode` with the recover annotation. Is this correct? We won't really know the state of the node. I think we should just log the error and move on to the next node.

pkg/controller/azdrivernode.go
```diff
@@ -88,25 +91,40 @@ func (r *ReconcileAzDriverNode) Reconcile(ctx context.Context, request reconcile
...
+			if _, err = azureutils.UpdateCRIWithRetry(ctx, nil, r.cachedClient, r.azClient, &node, updateFunc, consts.NormalUpdateMaxNetRetry, azureutils.UpdateCRI); err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			r.addNodeToAvailableAttachmentsMap(ctx, node.Name, node.GetLabels())
 		}
-		r.addNodeToAvailableAttachmentsMap(ctx, node.Name, node.GetLabels())
+	}
+	if len(errs) > 0 {
+		err = status.Errorf(codes.Internal, "azDriverNode recovery failed with: %+v", errs)
```
edreed, line 126: Rather than aggregating all of the errors in one error return, we should log the individual node errors as they are encountered. If there are any failures, it's fine to return a generic error from this method. I am concerned that if there are 100s to 1000s of nodes, we may end up with one really large error log which is difficult to use for debugging.

pkg/controller/azdrivernode.go
hccheng72: Okay, I thought it'd be saver to still recover the CRI if we can't confirm the node doesn't exist. I'll change it to log the error and keep the CRI without recovery.

pkg/controller/azdrivernode.go
hccheng72: Good idea!

pkg/controller/node.go
```diff
@@ -98,23 +100,41 @@ func (r *ReconcileNode) Recover(ctx context.Context) error {
...
+				w.Logger().Errorf(err, "failed to recover azDriverNode (%s) with annotation", azNode.Name)
+				continue
+			}
+			r.addNodeToAvailableAttachmentsMap(ctx, azNode.Name, azNode.GetLabels())
 		}
-		r.addNodeToAvailableAttachmentsMap(ctx, node.Name, node.GetLabels())
 	}
-	return nil
+
+	if err != nil {
```
edreed, line 134: You'll need to count the errors separately since `err` contains the result of the last iteration. An error won't be returned if the last iteration is successful but earlier iterations failed.

edreed: Actually, I just noticed that `err` was recreated (using `:=`) in the loop scope, so `err` will *always* `== nil` at this point.

PR #1616, opened 2022-11-23
[WIP]fix: resource group issue in disk attach

andyzhangx:
**What type of PR is this?**
/kind bug

fix: resource group issue in disk attach

Original behavior is using disk resource group as node resource group which is wrong, there would be issue when disk is not in node resource group, this PR fixes the issue.

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/c905fb10cb277916fb580fef75d276f328bc0d7c/pkg/provider/azure_controller_common.go#L270-L274

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: resource group issue in disk attach
```


ZeroMagic (approved):
/lgtm

PR #1625, opened 2022-11-29
[V2] fix: update PV controller's Recover function to log the error 

### LINKED ISSUE #1617: [V2] Both AzDriverNode and PV Controller's Recover function can't keep recovering if error happens
**What happened**:

In `azdrivernode.go`, the `Recover()` function will for-loop all nodes and recover them one by one. But if one node is failed on `Update()`, it will return error immediately instead of recovering the next node in the loop. Should we log the error and loop to the next node?
Code link:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/ea2d5a317b59658ee073dd025197526a022112b2/pkg/controller/azdrivernode.go#L107

In `pv.go`, the `Recover()` function will for-loop all pv-s and recover them one by one. But if one pv is failed on `GetDiskName()`, it will return error immediately instead of recovering the next pv in the loop. Should we log the error and loop to the next pv?
Code link:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/ea2d5a317b59658ee073dd025197526a022112b2/pkg/controller/pv.go#L245
 
**How to reproduce it**:

**Environment**:

- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

alice-zheyan-yu:
**What type of PR is this?**

/kind bug

Fixes #1617 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/pv.go
+++ b/pkg/controller/pv.go
@@ -21,9 +21,10 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
@@ -69,7 +70,7 @@ func (r *ReconcilePV) Reconcile(ctx context.Context, request reconcile.Request) 
 
 	// Ignore not found errors as they cannot be fixed by a requeue
 	if err := r.cachedClient.Get(ctx, request.NamespacedName, &pv); err != nil {
-		if errors.IsNotFound(err) {
+		if apierrors.IsNotFound(err) {
 			r.deletePV(request.Name)
 			return reconcileReturnOnSuccess(request.Name, r.controllerRetryInfo)
 		}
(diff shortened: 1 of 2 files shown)


hccheng72:
I think we should return a generic error like "pv recovery failed" in the end if any failures happen.

pkg/controller/pv.go
```diff
@@ -242,12 +244,20 @@ func (r *ReconcilePV) Recover(ctx context.Context) error {
 		var diskName string
 		diskName, err = azureutils.GetDiskName(pv.Spec.CSI.VolumeHandle)
 		if err != nil {
-			return err
+			w.Logger().Errorf(err, "failed to recover pv (%s)", pv.Name)
```
edreed, line 247: np: please spell out `PersistentVolume` instead of abbreviated it. Same applies below.

landreasyan: let's also log the error for debugging.

alice-zheyan-yu: Done.

pkg/controller/pv.go
```diff
@@ -242,12 +244,20 @@ func (r *ReconcilePV) Recover(ctx context.Context) error {
...
+			continue
 		}
 		azVolumeName := strings.ToLower(diskName)
 		pvClaimName := getQualifiedName(pv.Spec.ClaimRef.Namespace, pv.Spec.ClaimRef.Name)
 		r.addVolumeAndClaim(azVolumeName, pv.Name, pvClaimName)
 	}
+	if errCount > 0 {
+		pvCount := pvs.Size()
+		err = errors.New("failed to recover all pv(s)")
+		w.Logger().Errorf(err, "Total number of pv to be recovered is %d. But failed to recover (%d) of them.", pvCount, errCount)
```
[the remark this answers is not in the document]
alice-zheyan-yu: Done.

PR #1633, opened 2022-12-01
[V2] perf: changed to use channel to communicate when delete azvolume

### LINKED ISSUE #1589: [V2] azvolume.go add communication through channels instead of shared memory
We should be communicating between goroutines through channels instead of shared memory. 

https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1557#discussion_r1018462855

**Additional context**

alice-zheyan-yu:
**What type of PR is this?**

Fixes #1589

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/azvolume.go
+++ b/pkg/controller/azvolume.go
@@ -272,45 +272,40 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
 		var attachments []azdiskv1beta2.AzVolumeAttachment
 		attachments, err = r.cleanUpAzVolumeAttachmentByVolume(deleteCtx, azVolume.Name, azvolume, azureutils.AllRoles, mode)
 		if err == nil {
-			var wg sync.WaitGroup
-			errors := make([]error, len(attachments))
-			numErrors := uint32(0)
+			attachmentsCount := len(attachments)
+			errorMessageCh := make(chan string, attachmentsCount)
 
 			// start waiting for replica AzVolumeAttachment CRIs to be deleted
-			for i, attachment := range attachments {
-				var waiter *watcher.ConditionWaiter
-				waiter, err = r.conditionWatcher.NewConditionWaiter(deleteCtx, watcher.AzVolumeAttachmentType, attachment.Name, verifyObjectFailedOrDeleted)
-				if err != nil {
-					break
-				}
-
+			for _, attachment := range attachments {
 				// wait async and report error to go channel
-				wg.Add(1)
-				go func(ctx context.Context, waiter *watcher.ConditionWaiter, i int) {
+				go func(ctx context.Context, attachment azdiskv1beta2.AzVolumeAttachment) {
+					waiter, derr := r.conditionWatcher.NewConditionWaiter(deleteCtx, watcher.AzVolumeAttachmentType, attachment.Name, verifyObjectFailedOrDeleted)
 					defer waiter.Close()
-					defer wg.Done()
-					_, err := waiter.Wait(ctx)
-					if err != nil {
-						errors[i] = err
-						atomic.AddUint32(&numErrors, 1)
+					if derr != nil {
+						errorMessageCh <- fmt.Sprintf("%s: %v", attachment.Name, derr)
+						return
 					}
-				}(deleteCtx, waiter, i)
+
+					_, derr = waiter.Wait(ctx)
+					if derr != nil {
+						errorMessageCh <- fmt.Sprintf("%s: %v", attachment.Name, derr)
+					} else {
+						errorMessageCh <- ""
+					}
+				}(deleteCtx, attachment)
 			}
 
-			// start waiting for the attachment clean up to complete (outside of the (if err == nil) to avoid child workflow finishing after parent workflow finishing)
-			wg.Wait()
-
-			if err == nil {
-				// if errors have been found with the wait calls, format the error msg and report via CRI
-				if numErrors > 0 {
-					var errMsgs []string
-					for i, derr := range errors {
-						if derr != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/azvolume.go
```diff
@@ -286,31 +285,29 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
...
-						atomic.AddUint32(&numErrors, 1)
+						errorMessageCh <- fmt.Sprintf("%s: %v", attachment.Name, err)
 					}
-				}(deleteCtx, waiter, i)
+				}(deleteCtx, waiter, attachment)
 			}
 
 			// start waiting for the attachment clean up to complete (outside of the (if err == nil) to avoid child workflow finishing after parent workflow finishing)
 			wg.Wait()
+			close(errorMessageCh)
```
sunpa93, line 300: `errorMessageCh` was not created as a buffered channel, so the goroutines would never be able to exit, as it would be waiting for a recipient. That said, `wg.Wait()` would be blocked indefinitely if an error occurs.

So, you would either need to:
* make the `errorMessageCh` a buffered channel of `size = len(attachments)`
OR
* get rid of waitgroup.
* send result to `errorMessageCh` in both success and failure cases (`nil` value would be sent in case of success, and `non-nil` error would be sent in case of failure`
* loop over `len(azVolumeAttachment` and collect `non-nil` errors.
* close channel once completed looping.

pkg/controller/azvolume.go
```diff
@@ -286,31 +285,29 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
...
 
 			if err == nil {
-				// if errors have been found with the wait calls, format the error msg and report via CRI
-				if numErrors > 0 {
+				go func(errorMessageCh chan string) {
```
sunpa93, line 303: This shouldn't be done in a goroutine as below would have dependency on this operation.

https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/0c86bb7e168a42435da2211cef0535f2d48f76ea/pkg/controller/azvolume.go#L315

pkg/controller/azvolume.go
landreasyan: for the buffered `errorMessageCh` case, we would need to create another goroutine to wait and close the channel as follows: 
```go func() { wg.Wait() close(errors ) }()```
and after that we can loop over `errorMessageCh` to get all the errors: 
```for error := range errors { // update status }```

Comment (alice-zheyan-yu) [reply pkg/controller/azvolume.go]: Done. I prefer the second one since it gets rid of the waitgroup which is more straightforward.

File: pkg/controller/azvolume.go (hunk 2 above)
Comment (alice-zheyan-yu) [reply pkg/controller/azvolume.go]: Done.

File: pkg/controller/azvolume.go [hunk 3]
```diff
@@ -272,43 +272,39 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
 		var attachments []azdiskv1beta2.AzVolumeAttachment
 		attachments, err = r.cleanUpAzVolumeAttachmentByVolume(deleteCtx, azVolume.Name, azvolume, azureutils.AllRoles, mode)
 		if err == nil {
-			var wg sync.WaitGroup
-			errors := make([]error, len(attachments))
-			numErrors := uint32(0)
+			errorMessageCh := make(chan string)
```
edreed, line 275: Create a buffered channel with a buffer size equal to the number of attachments, i.e. `make(chan string, len(attachments))`.

pkg/controller/azvolume.go
```diff
@@ -272,43 +272,39 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
...
-							errMsgs = append(errMsgs, fmt.Sprintf("%s: %v", attachments[i].Name, derr))
-						}
+				var errMsgs []string
+				for i := 0; i < len(attachments); i++ {
+					v, ok := <-errorMessageCh
+					if ok && v != "" {
+						errMsgs = append(errMsgs, v)
 					}
+				}
+				close(errorMessageCh)
```
edreed, line 306: We need to close the channel unconditionally. With the current structure we can't do this safely because `NewConditionWaiter` could (theoretically) return an error after one or more `Wait` goroutines have already been dispatched. To make this safe, create the `ConditionWaiter` in the goroutine itself. You could then remove the `err == nil` check on line 297 and unconditionally close the channel after iterating through the results.

pkg/controller/azvolume.go
```diff
@@ -272,43 +272,39 @@ func (r *ReconcileAzVolume) triggerDelete(ctx context.Context, azVolume *azdiskv
...
 			if err == nil {
 				// if errors have been found with the wait calls, format the error msg and report via CRI
-				if numErrors > 0 {
-					var errMsgs []string
-					for i, derr := range errors {
-						if derr != nil {
-							errMsgs = append(errMsgs, fmt.Sprintf("%s: %v", attachments[i].Name, derr))
-						}
+				var errMsgs []string
+				for i := 0; i < len(attachments); i++ {
```
edreed, line 300: This look can be simplified using a range over the channel itself, e.g.

```go
for v := range errorMessageCh {
   ...
}
```

edreed: Oh, nevermind. We close the channel in this function and not in the goroutines.

pkg/controller/azvolume.go
alice-zheyan-yu: Oh, I see the benefit of buffered channel here! I will first consider buffered next time. Done.

pkg/controller/azvolume.go
alice-zheyan-yu: Yes! Done.

PR #1650, opened 2022-12-09
[V2] fix: update replica attachment failure retries logic

hccheng72:
**What type of PR is this?**
/kind feature

This PR is to update logic of replica attachment failure retries based on changes of cloud-provider-azure logic to handle preemption.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/controller/replica.go
```diff
@@ -108,13 +108,25 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
...
 			}
-			// if replica attachment failed, delete it and create a replacement
 		} else if azVolumeAttachment.Status.State == azdiskv1beta2.AttachmentFailed {
-			// if attachment failed for replica AzVolumeAttachment, delete the CRI so that replace replica AzVolumeAttachment can be created.
-			if err := r.cachedClient.Delete(ctx, azVolumeAttachment); err != nil {
-				return reconcile.Result{Requeue: true}, err
+			// if the failed attachment is allowed to retry, reset the state to pending
+			if azureutils.MapContains(azVolumeAttachment.Status.Annotations, consts.VolumeAttachRetryAnnotation) {
+				updateFunc := func(obj client.Object) error {
+					azva := obj.(*azdiskv1beta2.AzVolumeAttachment)
```
landreasyan, line 115: Let's remove the `consts.VolumeAttachRetryAnnotation` annotation here when we reset the state to `Pending` because we won't be needing it after that.

pkg/controller/attach_detach.go
```diff
@@ -124,11 +125,11 @@ func (r *ReconcileAttachDetach) Reconcile(ctx context.Context, request reconcile
 			return reconcileReturnOnError(ctx, azVolumeAttachment, "detach", err, r.retryInfo)
 		}
 		// attachment request
-	} else if azVolumeAttachment.Status.Detail == nil {
+	} else if azVolumeAttachment.Status.Detail == nil || (azVolumeAttachment.Status.State == azdiskv1beta2.AttachmentPending && azureutils.MapContains(azVolumeAttachment.Status.Annotations, consts.VolumeAttachRetryAnnotation)) {
```
landreasyan, line 128: If we remove the annotation and reset the state in replica.go, we won't be needing to add any new conditions here.

go.mod
```diff
@@ -17,28 +17,28 @@ require (
...
-	github.com/onsi/gomega v1.23.0
+	github.com/onsi/ginkgo/v2 v2.5.1
+	github.com/onsi/gomega v1.24.1
 	github.com/pborman/uuid v1.2.0
 	github.com/pelletier/go-toml v1.9.4
 	github.com/prometheus/client_golang v1.12.1
 	github.com/sirupsen/logrus v1.8.1
 	github.com/stretchr/testify v1.8.1
-	golang.org/x/net v0.1.0
+	golang.org/x/net v0.2.0
```
landreasyan, line 27: switch to 0.4.0 to fix the vulnerability issue.

hccheng72: Thanks for the pointer.

pkg/azureconstants/azure_constants.go
```diff
@@ -134,6 +134,8 @@ const (
 	RequestTimeFormat              = time.RFC3339Nano
 	RequesterKey                   = "disk.csi.azure.com/requester-name"
 	WorkflowKey                    = "disk.csi.azure.com/requester-name"
+	VolumeAttachRetryAnnotation    = "disk.csi.azure.com/volume-attach-retry"
+	VolumeAttachRetryCount         = "disk.csi.azure.com/volume-attach-retry-count"
```
landreasyan, line 138: Instead of keeping two annotations for retry, we could just keep the `volume-attach-retry-count`. When retry is needed, we could increment the count.

landreasyan: Actually, since we have two controllers for azvolumeattachments, having two different annotations might make it easier to coordinate the operations.

charts/latest-v2/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
```diff
@@ -16,6 +16,7 @@ data:
       leaderElectionNamespace: {{ .Release.Namespace }}
       workerThreads: {{ .Values.controller.driverWorkerThreads }}
       waitForLunEnabled: {{ .Values.controller.waitForLunEnabled}}
+      volumeAttachRetryLimit: {{ .Values.controller.volumeAttachRetryLimit }}
```
sunpa93, line 19: With this change, helm chart package needs to be regenerated and helm index needs to be updated.
`helm package charts/latest-v2/azuredisk-csi-driver -d charts/latest-v2`
`helm repo index ./charts --url https://raw.githubusercontent.com/kubernetes-sigs/azuredisk-csi-driver/main_v2/charts`

charts/latest-v2/azuredisk-csi-driver/values.yaml
```diff
@@ -147,6 +147,7 @@ controller:
   initContainers: []
   additionalContainers: []
   additionalVolumes: []
+  volumeAttachRetryLimit: 3
```
sunpa93, line 150: It would be good to create a constant for this value.

pkg/controller/replica.go
```diff
@@ -108,13 +109,48 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
...
+				retryCount := 0
+				if value, exists := azureutils.GetFromMap(azVolumeAttachment.Status.Annotations, consts.VolumeAttachRetryCount); exists {
+					if n, err := strconv.Atoi(value); err == nil {
+						retryCount = n
+					} else {
+						r.logger.Error(err, "failed to convert retry count to int")
+					}
+				}
+
+				var updateFunc func(obj client.Object) error
```
sunpa93, line 124: You can use `UpdateCRIFunc` type.

https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/c6813a5c47afefcaa13df68f9f93bc37911b4ebf/pkg/azureutils/azure_disk_utils.go#L1072

pkg/controller/attach_detach.go
```diff
@@ -27,7 +28,7 @@ import (
 	"google.golang.org/grpc/status"
 	v1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierr "k8s.io/apimachinery/pkg/api/errors"
```
sunpa93, line 31: If you plan to change the package import name, let's stick with the same naming as in other files.
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/c6813a5c47afefcaa13df68f9f93bc37911b4ebf/pkg/provisioner/crdprovisioner.go#L32

pkg/controller/replica.go
```diff
@@ -108,13 +109,48 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
...
+				if retryCount < r.config.ControllerConfig.VolumeAttachRetryLimit {
+					updateFunc = func(obj client.Object) error {
+						azva := obj.(*azdiskv1beta2.AzVolumeAttachment)
+						azureutils.RemoveFromMap(azva.Status.Annotations, consts.VolumeAttachRetryAnnotation)
+						azureutils.AddToMap(azva.Status.Annotations, consts.VolumeAttachRetryCount, strconv.Itoa(retryCount+1))
+						_, derr := updateState(azva, azdiskv1beta2.AttachmentPending, normalUpdate)
```
sunpa93, line 130: Use `forceUpdate` mode for updating state from `AttachmentFailed` to `AttachmentPending`

pkg/controller/attach_detach.go
```diff
@@ -81,7 +83,7 @@ var allowedTargetAttachmentStates = map[string][]string{
 	string(azdiskv1beta2.Detaching):          {string(azdiskv1beta2.Detached), string(azdiskv1beta2.DetachmentFailed)},
 	string(azdiskv1beta2.Attached):           {string(azdiskv1beta2.Detaching)},
 	string(azdiskv1beta2.Detached):           {},
-	string(azdiskv1beta2.AttachmentFailed):   {string(azdiskv1beta2.Detaching)},
+	string(azdiskv1beta2.AttachmentFailed):   {string(azdiskv1beta2.Detaching), string(azdiskv1beta2.AttachmentPending)},
```
sunpa93, line 86: State transition from `AttachmentFailed` to `AttachmentPending` shouldn't be allowed as this would allow `attach_detach_controller` to retry on failed attachments on its own without external triggers. When we wish to update state from `AttachmentFailed` to `AttachmentPending` to retrigger attach operation in `replica controller`, we should use force update mode instead.

hccheng72: Got it, thanks for the explanation.

charts/latest-v2/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
hccheng72: Oh yeah! Thanks for the reminder.

edreed:
@landreasyan @hccheng72 kubernetes-sigs/azuredisk-csi-driver#1654 has been merged. Do we still need a hold on this change?

pkg/azureconstants/azure_constants.go
```diff
@@ -216,6 +218,7 @@ const (
 	DefaultAzureClientAttachDetachRateLimiterQPS    = (240.0 / 180.0)                                          // Default compute QPS limit is 240 queries / 3 minutes
 	DefaultAzureClientAttachDetachRateLimiterBucket = int(DefaultAzureClientAttachDetachRateLimiterQPS * 60.0) // Allow for a burst of a minutes worth of quota
 	DefaultAzureClientAttachDetachBatchInitialDelay = 1 * time.Second                                          // Wait 1s before processing a batch of attach or detach disk requests
+	DefaultVolumeAttachRetryLimit                   = 2
```
edreed, line 221: @sunpa93 Is 2 a large enough default?

pkg/controller/replica.go
```diff
@@ -113,10 +114,47 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
 				}
 				// if replica attachment failed, delete the CRI and create a replacement
```
edreed, line 115: This is statement is no longer unconditionally true.

pkg/controller/replica.go
```diff
@@ -113,10 +114,47 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
...
+							return derr
+						}
+					}
+
+					if _, err := azureutils.UpdateCRIWithRetry(ctx, nil, r.cachedClient, r.azClient, azVolumeAttachment, updateFunc, consts.NormalUpdateMaxNetRetry, azureutils.UpdateCRIStatus); err != nil {
+						return reconcile.Result{Requeue: true}, err
+					}
+					// else, delete the CRI so that replace replica AzVolumeAttachment can be created.
+				} else {
+					if err := r.cachedClient.Delete(ctx, azVolumeAttachment); err != nil {
```
edreed, line 153: I find it easier to see the flow of code if the smaller of the `if..else` clauses is first, e.g.

```go
if !azureutils.MapContains(...) {
    // delete AzVolumeAttachment CRI clause
} else {
    // much larger retry handling clause}
```

This is especially true when one clause is *MUCH* larger than the other.

hccheng72: Nice suggestion! Thank you.

pkg/azureconstants/azure_constants.go
sunpa93: It occurs to me that we should have different retry limits based on what kind of errors we have retrieved. In case of attach failure due to `contextTimeout`, 2 retries would add up to 15 minutes of waiting for attach operation to complete as that would be total 3 attach attempts with 5 minutes timeout each. However, in case of attach failures due to `operationPreemption`, where the operation usually fails early, 2 retries may not give sufficient amount of time.

hccheng72: I did two experiments of 500 pods scale test with the retry limit of **3**. Each of them has 40-50 times of retriable error, totally only one CRI retried twice while others just used one retry. I'll keep the retry limit to 2 as discussed and we can improve it in the future when we get more ideas about how it works.

hccheng72:
The code has been updated and the scale test with replicas of 500 pods succeeded.

sunpa93 (approved):
/lgtm

PR #1653, opened 2022-12-09
[V2] fix: make recover annotation unique per recovery

### LINKED ISSUE #1648: [V2] CRI Recovery Enhancement / Bug Fix
**What happened**:
1. The driver relies on adding `RecoverAnnotation` to force register an UPDATE event for reconciliation to retrigger any operations that failed to complete due to driver restart or upgrade. https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/b084d41779abcfb6a026949dad092d90c314b1ec/pkg/controller/azvolume.go#L550-L551
However, if driver restarts more than once, any CRI that were recovered from previous restart will already have the `RecoverAnnotation`, making the status update operation a noop. Then, no UPDATE event will be registered, and no recovery reconciliation will be triggered for that CRI.
> Therefore, we need to create a unique UUID per recovery to set as a value for the `RecoverAnnotation`.

2. With v2's asynchronous attach feature, `VolumeAttachment.Status.Attached` no longer truly reflects on disk's attach state. As `ControllerPublishVolume` no longer waits for attachment to complete and instead returns an early success, we cannot really tell if the disk is attached to a node or not by looking at `VolumeAttachment.Status.Attached`. And this becomes a problem when we try to recover states for AzVolumeAttachment CRIs on driver restart or upgrade because the driver is clueless on what state the `AzVolumeAttachment.Status.State` should be at. 
> Therefore, we should fetch VMs from cloud-provider-azure and check disks attached to properly recover the `Status.State` field.

* CRI Recovery should properly trigger a controller reconciliation with a unique RecoverAnnotation update.
* AzVolumeAttachment CRI Recovery should have a correct `Status.State`.

**How to reproduce it**:

**Environment**:

- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

alice-zheyan-yu:
**What type of PR is this?**

/kind bug

Fixes #1648

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azuredisk_v2.go
+++ b/pkg/azuredisk/azuredisk_v2.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/go-logr/logr"
+	"github.com/pborman/uuid"
 
 	v1 "k8s.io/api/core/v1"
 	crdClientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
@@ -421,15 +422,20 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
 		var errors []error
 		ctx, w := workflow.New(ctx)
 		defer func() { w.Finish(errors...) }()
+
+		// Use unique recoveryID as the Recover Annotation, so if it recovers more than once(for example: driver restarts or upgrades)
+		// the objects can still get a new Recover Annotation to force register an UPDATE event for reconciliation.
+		recoveryID := uuid.NewUUID().String()
+
 		// recover lost states if necessary
-		w.Logger().Infof("Elected as leader; initiating CRI deperecation / recovery...")
-		if err := nodeReconciler.Recover(ctx); err != nil {
+		w.Logger().Infof("Elected as leader; initiating CRI deperecation / recovery with recoveryID: %s ...", recoveryID)
+		if err := nodeReconciler.Recover(ctx, recoveryID); err != nil {
 			errors = append(errors, err)
 		}
-		if err := azvReconciler.Recover(ctx); err != nil {
+		if err := azvReconciler.Recover(ctx, recoveryID); err != nil {
 			errors = append(errors, err)
 		}
-		if err := attachReconciler.Recover(ctx); err != nil {
+		if err := attachReconciler.Recover(ctx, recoveryID); err != nil {
 			errors = append(errors, err)
 		}
 		if err := pvReconciler.Recover(ctx); err != nil {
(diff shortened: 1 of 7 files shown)


pkg/azureutils/azure_disk_utils.go
```diff
@@ -1386,3 +1386,8 @@ func GetDefaultDiskMBPSReadWrite(requestGiB int) int {
 	}
 	return bandwidth
 }
+
+func GenerateUniqueRecoverAnnotation() string {
+	uuid := uuid.NewUUID().String()
+	return consts.RecoverAnnotation + uuid
```
sunpa93, line 1392: Instead of generating a unique `recoverAnnotation` key by concatenating the annotation prefix with an uuid, we should use `recoverAnnotation` as the key and generated uuid as the value.

alice-zheyan-yu: Yes. Done.

pkg/controller/attach_detach.go
```diff
@@ -496,8 +497,11 @@ func (r *ReconcileAttachDetach) Recover(ctx context.Context) error {
 	}
 	// retrigger any aborted operation from possible previous controller crash
 	recovered := &sync.Map{}
+	// Make Recover Annotation unique per Recovery, so if driver restarts more than once
+	// the AzVolumeAttachment can still get a new Recover Annotation to force register an UPDATE event for reconciliation
+	uniqueRecoverAnnotation := uuid.NewUUID().String()
```
sunpa93, line 502: Consider creating the recovery uuid in `azuredisk_v2.go` and passing it in as an argument to `attachdetach.Recover`, `azvolume.Recover`, and `azdrivernode.Recover`.

sunpa93: np: Consider naming the variable `recoveryId` maybe?

alice-zheyan-yu: Like the idea to name it `recoveryId `! Also if we create the uuid in `azuredisk_v2.go`, we can use the same uuid for 3 controllers. Nice!

pkg/azuredisk/azuredisk_v2.go
```diff
@@ -423,13 +424,16 @@ func (d *DriverV2) StartControllersAndDieOnExit(ctx context.Context) {
 		defer func() { w.Finish(errors...) }()
 		// recover lost states if necessary
 		w.Logger().Infof("Elected as leader; initiating CRI deperecation / recovery...")
-		if err := nodeReconciler.Recover(ctx); err != nil {
+		// Use unique recoveryID as the Recover Annotation, so if it recovers more than once(for example: driver restarts or upgrades)
+		// the objects can still get a new Recover Annotation to force register an UPDATE event for reconciliation.
+		recoveryID := uuid.NewUUID().String()
```
edreed, line 429: It would be good to include the recovery ID in the log trace above so that we can match the recovered CRIs to the driver logs.

alice-zheyan-yu: Done.

PR #1658, opened 2022-12-13
[V2] feat: toggle for waitforlun

connorbradshaw10:
**What type of PR is this?**

/kind feature

- adding ability to toggle on/off waitforlun
- setting waitForLunEnabled=false will make publishvolume faster but will break the driver if you downgrade to v1

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/provisioner/crdprovisioner.go
```diff
@@ -177,12 +178,13 @@ func (a CachedReader) List(_ context.Context, objectList client.ObjectList, opts
 	return err
 }
 
-func NewCrdProvisioner(azdiskClient azdisk.Interface, objNamespace string, cw *watcher.ConditionWatcher, azCachedReader CachedReader, informer azureutils.GenericInformer) *CrdProvisioner {
+func NewCrdProvisioner(azdiskClient azdisk.Interface, objNamespace string, cw *watcher.ConditionWatcher, azCachedReader CachedReader, informer azureutils.GenericInformer, waitForLunEnabled bool) *CrdProvisioner {
```
edreed, line 181: Instead of adding a new parameter, let's replace the `objNamespace string` parameter with a `config *azdiskv1beta2.AzDiskDriverConfiguration` parameter.

edreed:
The `translations/en_US/LC_MESSAGES/ascengine.*` files need to be removed from this PR.

connorbradshaw10:
> 

can I add this to the .gitignore for future cases?

connorbradshaw10:
> The `translations/en_US/LC_MESSAGES/ascengine.*` files need to be removed from this PR.
can I add this to the .gitignore for future cases?

edreed:
> 

Yeah, that wouldn't be a bad idea.

pkg/provisioner/crdprovisioner.go
connorbradshaw10: fixed

charts/latest-v2/azuredisk-csi-driver/values.yaml
```diff
@@ -58,6 +58,7 @@ controller:
   cloudConfigSecretName: azure-cloud-provider
   cloudConfigSecretNamespace: kube-system
   allowEmptyCloudConfig: false
+  waitForLunEnabled: true
```
sunpa93, line 61: `helm package charts/latest-v2/azuredisk-csi-driver -d charts/latest-v2`

sunpa93 (approved):
/lgtm

landreasyan (approved):

PR #1663, opened 2022-12-16
[V2] fix: clear status error and detail in AzVolumeAttachment recovery if in in Attaching state 🐞

jmclong:
**What type of PR is this?**

/kind bug

This PR fixes an issue in 1AzVolumeAttachment1 recovery. If the 1AzVolumeAttachment1 is in the `Attaching` state when the driver crashes, we recover the state by setting the status to `AttachmentPending`, as we do not know where in the `Attaching` state the controller was in when it crashed.

However, the code is checking if `azVolumeAttachment.Status.Detail` is `nil` prior to calling `triggerAttach`. Setting the `Status.Detail` and `Status.Error` to `nil` will ensure that the driver completes the attachment upon recovery. 

This issue was exposed because we are seeing driver OOM issues.

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/attach_detach.go
+++ b/pkg/controller/attach_detach.go
@@ -765,6 +765,16 @@ func (r *ReconcileAttachDetach) recoverAzVolumeAttachment(ctx context.Context, r
 			case azdiskv1beta2.Attaching:
 				// reset state to Pending so Attach operation can be redone
 				targetState = azdiskv1beta2.AttachmentPending
+				innerUpdateFunc := updateFunc
+				updateFunc = func(obj client.Object) error {
+					if err := innerUpdateFunc(obj); err != nil {
+						return err
+					}
+					azv := obj.(*azdiskv1beta2.AzVolumeAttachment)
+					azv.Status.Detail = nil
+					azv.Status.Error = nil
+					return nil
+				}
 			case azdiskv1beta2.Detaching:
 				// reset state to Attached so Detach operation can be redone
 				targetState = azdiskv1beta2.Attached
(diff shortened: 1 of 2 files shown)


pkg/controller/attach_detach.go
```diff
@@ -742,6 +742,7 @@ func (r *ReconcileAttachDetach) recoverAzVolumeAttachment(ctx context.Context, r
 		go func(azv azdiskv1beta2.AzVolumeAttachment, azvMap *sync.Map) {
 			defer wg.Done()
 			var targetState azdiskv1beta2.AzVolumeAttachmentAttachmentState
+			clearState := false
```
edreed, line 745: Consider `clearDetail`

pkg/controller/attach_detach.go
```diff
@@ -759,12 +760,18 @@ func (r *ReconcileAttachDetach) recoverAzVolumeAttachment(ctx context.Context, r
...
+					azv.Status.Detail = nil
+					azv.Status.Error = nil
+				}
 				return err
 			}
 			switch azv.Status.State {
 			case azdiskv1beta2.Attaching:
 				// reset state to Pending so Attach operation can be redone
 				targetState = azdiskv1beta2.AttachmentPending
+				clearDetail = true
```
edreed, line 774: Rather than use a Boolean flag referenced by the update function to determine whether or not to clear the `Status.Detail` field, it would be cleaner to wrap `updateFunc` in another function that clears the detail, e.g.

```go
innerUpdateFunc := updateFunc
updateFunc := func(obj client.Object) error {
    if err := innerUpdateFunc(obj); err != nil {
        return err
    }

    azv := obj.(*azdiskv1beta2.AzVolumeAttachment)
    azv.Status.Detail = nil
    azv.Status.Error = nil
}
```

jmclong:
@edreed, thank you. There were some recent changes that @landreasyan and I were trying to track down yesterday that broken test-infra. It ended up being reverted:

https://github.com/kubernetes/test-infra/pull/28274

PR #1666, opened 2022-12-20
[V2] feat: fetch VMs to recover AzVolumeAttachment 

### LINKED ISSUE #1648: [V2] CRI Recovery Enhancement / Bug Fix

alice-zheyan-yu:
**What type of PR is this?**

/kind bug

Without this change, only Primary role can be recreated and the Status.State and Status.Detail are missing. This change ensures that both Primary and Replica role can be recreated and if they are attached their Status.state can be recovered to "Attached". And the Status.Detail can be recover.

This change will do 3 steps to Recover AzVolumeAttachment:

1, recreate AzVolumeAttachments with Primary role from VolumeAttachemnts

2, recreate AzVolumeAttachments with Replica role from VMs

3, recover the  AzVolumeAttachments.Status.State and AzVolumeAttachments.Status.Detail

**How to test it?**
1, install driver 
2, create AzVolumeAttachments with Primary role and Replica role
3, update cleanUpAzVolumeAttachments() so when uninstall driver the replica role won't be detached
4, uninstall driver
5, install driver
6, it can recover AzVolumeAttachments with Primary role and Replica role. And both of their States are "Attached". And both of them have Status.Detail.

Fixes #1648

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azureconstants/azure_constants.go
+++ b/pkg/azureconstants/azure_constants.go
@@ -131,7 +131,7 @@ const (
 	PodNameKey                         = "disk.csi/azure.com/pod-name"
 	PreProvisionedVolumeAnnotation     = "disk.csi.azure.com/pre-provisioned"
 	RequestIDKey                       = "disk.csi.azure.com/request-id"
-	RequestStartimeKey                 = "disk.csi.azure.com/request-starttime"
+	RequestStartTimeKey                = "disk.csi.azure.com/request-starttime"
 	RequestTimeFormat                  = time.RFC3339Nano
 	RequesterKey                       = "disk.csi.azure.com/requester-name"
 	WorkflowKey                        = "disk.csi.azure.com/requester-name"
--- a/pkg/controller/attach_detach.go
+++ b/pkg/controller/attach_detach.go
@@ -19,10 +19,13 @@ package controller
 import (
 	"context"
 	"errors"
+	"fmt"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
 
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2022-03-01/compute"
 	"github.com/go-logr/logr"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -31,9 +34,8 @@ import (
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	utilfeature "k8s.io/apiserver/pkg/util/feature"
-	"k8s.io/kubernetes/pkg/features"
 	azdiskv1beta2 "sigs.k8s.io/azuredisk-csi-driver/pkg/apis/azuredisk/v1beta2"
+	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 	consts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/provisioner"
@@ -53,11 +55,21 @@ import (
 type CloudDiskAttachDetacher interface {
 	PublishVolume(ctx context.Context, volumeID string, nodeID string, volumeContext map[string]string) provisioner.CloudAttachResult
 	UnpublishVolume(ctx context.Context, volumeID string, nodeID string) error
+	GetNodeDataDisks(nodeName types.NodeName) ([]compute.DataDisk, *string, error)
+	GetDiskLun(diskName, diskURI string, nodeName types.NodeName) (int32, *string, error)
+	CheckDiskExists(ctx context.Context, diskURI string) (*compute.Disk, error)
+	GetNodeNameByProviderID(providerID string) (types.NodeName, error)
 }
 
 type CrdDetacher interface {
 	UnpublishVolume(ctx context.Context, volumeID string, nodeID string, secrets map[string]string, mode consts.UnpublishMode) error
 	WaitForDetach(ctx context.Context, volumeID, nodeID string) error
+}
+
+type volumeInfo struct {
+	syncedAzVolumeAttachments map[string]bool
+	handle                    string
+	attributes                map[string]string
 }
 
 /*
@@ -73,6 +85,7 @@ type ReconcileAttachDetach struct {
 	cloudDiskAttacher CloudDiskAttachDetacher
 	stateLock         *sync.Map
 	retryInfo         *retryInfo
+	volumeInfos       map[string]*volumeInfo
 }
 
 var _ reconcile.Reconciler = &ReconcileAttachDetach{}
@@ -181,7 +194,7 @@ func (r *ReconcileAttachDetach) triggerAttach(ctx context.Context, azVolumeAttac
 	// update status block
 	updateFunc := func(obj client.Object) error {
 		azv := obj.(*azdiskv1beta2.AzVolumeAttachment)
-		// Update state to attaching, Initialize finalizer and add label to the object
+		// Update state to attaching
 		_, derr := updateState(azv, azdiskv1beta2.Attaching, normalUpdate)
 		return derr
 	}
@@ -493,33 +506,6 @@ func (r *ReconcileAttachDetach) detachVolume(ctx context.Context, volumeID, node
 	return r.cloudDiskAttacher.UnpublishVolume(ctx, volumeID, node)
 }
 
-func (r *ReconcileAttachDetach) Recover(ctx context.Context, recoveryID string) error {
-	var err error
-	ctx, w := workflow.New(ctx)
-	defer func() { w.Finish(err) }()
-
-	w.Logger().V(5).Info("Recovering AzVolumeAttachment CRIs...")
-	// try to recreate missing AzVolumeAttachment CRI
-	var syncedVolumeAttachments, volumesToSync map[string]bool
-
-	for i := 0; i < maxRetry; i++ {
-		if syncedVolumeAttachments, volumesToSync, err = r.recreateAzVolumeAttachment(ctx, syncedVolumeAttachments, volumesToSync); err == nil {
-			break
-		}
-		w.Logger().Error(err, "failed to recreate missing AzVolumeAttachment CRI")
-	}
-	// retrigger any aborted operation from possible previous controller crash
-	recovered := &sync.Map{}
-	for i := 0; i < maxRetry; i++ {
-		if err = r.recoverAzVolumeAttachment(ctx, recovered, recoveryID); err == nil {
-			break
-		}
-		w.Logger().Error(err, "failed to recover AzVolumeAttachment state")
-	}
-
-	return err
-}
-
 func updateRole(azVolumeAttachment *azdiskv1beta2.AzVolumeAttachment, role azdiskv1beta2.Role) *azdiskv1beta2.AzVolumeAttachment {
 	if azVolumeAttachment == nil {
 		return nil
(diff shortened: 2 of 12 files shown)


vendor/sigs.k8s.io/cloud-provider-azure/pkg/provider/azure_controller_common.go
```diff
@@ -476,8 +476,8 @@ func (c *controllerCommon) UpdateVM(ctx context.Context, nodeName types.NodeName
 	return vmset.UpdateVM(ctx, nodeName)
 }
 
-// getNodeDataDisks invokes vmSet interfaces to get data disks for the node.
-func (c *controllerCommon) getNodeDataDisks(nodeName types.NodeName, crt azcache.AzureCacheReadType) ([]compute.DataDisk, *string, error) {
+// GetNodeDataDisks invokes vmSet interfaces to get data disks for the node.
```
sunpa93, line 479: Please submit a PR with following changes in `cloud-provider-azure` to https://github.com/edreed/cloud-provider-azure

alice-zheyan-yu: Done. I submitted the PR for it: https://github.com/edreed/cloud-provider-azure/pull/172

alice-zheyan-yu:
/hold
This PR currently contains unmerged changes from cloud-provider-azure.
https://github.com/edreed/cloud-provider-azure/pull/172

alice-zheyan-yu:
/hold
I will add the changes of function `recoverAzVolumeAttachment` within this PR to have a correct recover logic.

edreed:
@alice-zheyan-yu I merged edreed/cloud-provider-azure#172. You will need to update the [cloud-provider-azure library version](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/13b204b5dc205f08bdc7f67449e01b25b401c209/go.mod#L194) in `go.mod`. You can get the version suffix using the instructions in [How to update Azure cloud provider library](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/docs/csi-dev.md#how-to-update-azure-cloud-provider-library).

alice-zheyan-yu:
> @alice-zheyan-yu I merged [edreed/cloud-provider-azure#172](https://github.com/edreed/cloud-provider-azure/pull/172). You will need to update the [cloud-provider-azure library version](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/13b204b5dc205f08bdc7f67449e01b25b401c209/go.mod#L194) in `go.mod`. You can get the version suffix using the instructions in [How to update Azure cloud provider library](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/docs/csi-dev.md#how-to-update-azure-cloud-provider-library).
> 
> (Ignore the reference to https://github.com/kubernetes/legacy-cloud-providers. It needs to be updated to https://github.com/kubernetes-sigs/cloud-provider-azure, but you'll need to use https://github.com/edreed/cloud-provider-azure.)

Done. But the wiki couldn't generate the correct suffix. I used `$go get github.com/edreed/cloud-provider-azure@61f1e6ab985444285cfac0313de66e4d528c5f4f` to find the version number.

sunpa93 (changes requested):

pkg/controller/attach_detach.go
```diff
@@ -741,45 +842,50 @@ func (r *ReconcileAttachDetach) recoverAzVolumeAttachment(ctx context.Context, r
 		wg.Add(1)
 		go func(azv azdiskv1beta2.AzVolumeAttachment, azvMap *sync.Map) {
 			defer wg.Done()
+			var publishCtx map[string]string
+			if lun, _, err := r.cloudDiskAttacher.GetDiskLun(azv.Spec.VolumeName, azv.Spec.VolumeID, types.NodeName(azv.Spec.NodeName)); err == nil {
```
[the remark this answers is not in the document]
alice-zheyan-yu: Hi @sunpa93 Could you share the code links of `managedDiskController` and `disk.DiskProperties.ShareInfo` ? Thanks.

pkg/controller/attach_detach.go
```diff
@@ -603,122 +627,199 @@ func (r *ReconcileAttachDetach) recreateAzVolumeAttachment(ctx context.Context,
...
+			}
+			_, ok := r.azVolumeAttachmentToVaMap.Load(azVolumeAttachmentName)
+			if ok {
+				// If the AzVolumeAttachment has a VolumeAttachment mapping to it, it means it's a Primary role which should be recreated already. Skip it.
+				continue
+			}
 
-				// for now, we don't empty the meta annotatinos after migrating them to status annotation for safety.
-				// note that this will leave some remnant garbage entries in meta annotations
+			volumeHandle, hasVolumeHandles := volumeHandles[*disk.Name]
```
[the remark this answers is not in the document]
alice-zheyan-yu: Yes, you are right. I may need to get volumeHandle from VM instead.

pkg/controller/attach_detach.go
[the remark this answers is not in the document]
sunpa93: This is code block on getting the disk with the disk client.
https://github.com/edreed/cloud-provider-azure/blob/9397319d763831f7e9a1d8aa8253f23d15d558cc/pkg/provider/azure_managedDiskController.go#L324

And this is what `compute.Disk` struct looks like:
https://github.com/edreed/cloud-provider-azure/blob/9397319d763831f7e9a1d8aa8253f23d15d558cc/vendor/github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2022-03-01/compute/models.go#L4078-:4100

`disk.ManagedByExtended` and `disk.Properties.ShareInfos` reflect on the vms the disk is attached to. Now that I think about it, it makes more sense to rely on `disk.ManagedByExtended` instead.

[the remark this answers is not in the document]
alice-zheyan-yu: Hi @sunpa93 I found the description of `disk.ManagedByExtended` here:  https://learn.microsoft.com/en-us/rest/api/compute/disks/get?tabs=HTTP#disk. So I reached out to Chetan Maddula from Disk RP team and confirmed that we can either use `disk.ManagedByExtended` or `disk.DiskProperties.DiskState` since they are linked data as a source of truth (unless there are bugs between crp and diskrp).

Next step: I'll run some examples to see how the property would change from shared disks vs non shared disks to confirm that we can rely on that to recover our AzVolumeAttachment.

pkg/controller/attach_detach.go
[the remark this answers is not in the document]
alice-zheyan-yu: Wait. If there is no Primary attachment, why do we need the replica? Maybe for safety reason?

It seems has 3 chooses:
1, recreate the replica AzVolumeAttachment from disk (not sure do we have enough data to do the recreation)
2, detach the disk
3, promote it to Primary Attachment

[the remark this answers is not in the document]
sunpa93: When we have `maxShares > 1`, our strategy for `ControllerUnpublishVolume` is to demote `Primary` to `Replica`. So, in those cases, if uninstallation fails to detach the replica attachment, we could end up with a replica `AzVolumeAttachment` without primary `AzVolumeAttachment`

pkg/controller/attach_detach.go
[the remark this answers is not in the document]
sunpa93: Awesome! Great to hear that there is a source of truth we can safely rely on. Thanks for checking with the DiskRP team to confirm this. Good progress indeed!

pkg/controller/attach_detach.go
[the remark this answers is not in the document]
alice-zheyan-yu: Got it. Then I will try to fetch the `AzVolume` (which should have `volumeId`) to generate my `volumeHandle`. If there is no `AzVolume`, we don't need to recreate any `AzVolumeAttachment`.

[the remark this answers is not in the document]
alice-zheyan-yu: I was wrong. If we end up with a replica `AzVolumeAttachment `without primary `AzVolumeAttachment`, we also have no `AzVolume`. So we can't fetch the `AzVolume `to get the `volumeId`. I'll keep looking for other ways to get the `volumeId` to recover replicas.

pkg/controller/attach_detach.go
```diff
@@ -603,122 +627,199 @@ func (r *ReconcileAttachDetach) recreateAzVolumeAttachment(ctx context.Context,
...
-					VolumeName:    volumeName,
-					VolumeID:      pv.Spec.CSI.VolumeHandle,
-					NodeName:      nodeName,
-					RequestedRole: azdiskv1beta2.PrimaryRole,
-					VolumeContext: pv.Spec.CSI.VolumeAttributes,
-				},
+			desiredAzVolumeAttachment := r.generateDesiredAzVolumeAttachment(azVolumeAttachmentName, nodeName, volumeName, pv.Spec.CSI.VolumeHandle, azdiskv1beta2.PrimaryRole, pv.Spec.CSI.VolumeAttributes)
+
+			err = r.recreateAzVolumeAttachment(ctx, azVolumeAttachmentName, desiredAzVolumeAttachment)
+			if err == nil {
```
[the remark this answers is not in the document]
alice-zheyan-yu: Done

pkg/controller/attach_detach.go
```diff
@@ -603,122 +627,199 @@ func (r *ReconcileAttachDetach) recreateAzVolumeAttachment(ctx context.Context,
...
-					azVolumeAttachment.Status.Annotations = azureutils.AddToMap(azVolumeAttachment.Status.Annotations, k, v)
-				}
+		for _, disk := range disks {
+			azVolumeAttachmentName := azureutils.GetAzVolumeAttachmentName(*disk.Name, nodeName)
+			// store node name and disk name pair which is attached
+			nodeDiskAttachedPairs[azVolumeAttachmentName] = true
+			if syncedAzVolumeAttachments[azVolumeAttachmentName] {
+				continue
+			}
+			_, ok := r.azVolumeAttachmentToVaMap.Load(azVolumeAttachmentName)
```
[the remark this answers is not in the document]
alice-zheyan-yu: Done.

alice-zheyan-yu:
I will move some small helper functions to `shared_state.go` to maintain `attach_detach.go` clean.

pkg/azureconstants/azure_constants.go
```diff
@@ -130,7 +130,7 @@ const (
 	PodNameKey                     = "disk.csi/azure.com/pod-name"
 	PreProvisionedVolumeAnnotation = "disk.csi.azure.com/pre-provisioned"
 	RequestIDKey                   = "disk.csi.azure.com/request-id"
-	RequestStartimeKey             = "disk.csi.azure.com/request-starttime"
```
edreed, line 133: np: `request-start-time`

pkg/controller/attach_detach.go
```diff
@@ -50,6 +52,9 @@ import (
 type CloudDiskAttachDetacher interface {
 	PublishVolume(ctx context.Context, volumeID string, nodeID string, volumeContext map[string]string) provisioner.CloudAttachResult
 	UnpublishVolume(ctx context.Context, volumeID string, nodeID string) error
+	GetNodeDataDisks(nodeName types.NodeName, crt azcache.AzureCacheReadType) ([]compute.DataDisk, *string, error)
```
edreed, line 55: Let's avoid exposing the `cloud-provider-azure` types via this interface and drop the `crt` parameter. I doubt that we would ever need to use anything by the default.

pkg/controller/attach_detach.go
```diff
@@ -599,128 +585,130 @@ func (r *ReconcileAttachDetach) recreateAzVolumeAttachment(ctx context.Context,
...
-					pv.Spec.AzureDisk != nil {
-					if pv, err = r.translateInTreePVToCSI(pv); err != nil {
-						w.Logger().V(5).Errorf(err, "skipping azVolumeAttachment creation for volumeAttachment (%s)", volumeAttachment.Name)
-					}
-				}
-			}
-
-			if pv.Spec.CSI == nil || pv.Spec.CSI.Driver != r.config.DriverName {
+			// get PV and retrieve volume info
+			volumeHandle, volumeContext, err := r.getVolumeInfoFromPV(ctx, *pvName)
```
edreed, line 590: Could return `volumeInfo`

pkg/controller/attach_detach.go
```diff
@@ -599,128 +585,130 @@ func (r *ReconcileAttachDetach) recreateAzVolumeAttachment(ctx context.Context,
...
 
-				for k, v := range azVolumeAttachment.Annotations {
-					azVolumeAttachment.Status.Annotations = azureutils.AddToMap(azVolumeAttachment.Status.Annotations, k, v)
-				}
+// recreateAzVolumeAttachmentWithReplicaRole fetches attachment information from VM to properly recreate replica AzVolumeAttachments
+func (r *ReconcileAttachDetach) recreateAzVolumeAttachmentWithReplicaRole(
+	ctx context.Context,
+	syncedAzVolumeAttachments map[string]bool,
+	volumeHandles map[string]string,
+	volumeAttributes map[string]map[string]string) (map[string]bool, error) {
```
edreed, line 624: Use `volumeInfo`

pkg/controller/attach_detach.go
alice-zheyan-yu: I'm wondering what's the reason that we need to avoid exposing `types.NodeName`?

pkg/controller/attach_detach.go
```diff
@@ -573,19 +551,27 @@ func reportError(azVolumeAttachment *azdiskv1beta2.AzVolumeAttachment, state azd
 	return updateState(azVolumeAttachment, state, forceUpdate)
 }
 
-func (r *ReconcileAttachDetach) recreateAzVolumeAttachment(ctx context.Context, syncedVolumeAttachments map[string]bool, volumesToSync map[string]bool) (map[string]bool, map[string]bool, error) {
+func (r *ReconcileAttachDetach) recreateAzVolumeAttachmentWithPrimaryRole(
+	ctx context.Context,
+	syncedVolumeAttachments map[string]bool,
+	volumeHandles map[string]string,
+	volumeAttributes map[string]map[string]string) (map[string]bool, map[string]string, map[string]map[string]string, error) {
```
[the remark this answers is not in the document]
edreed: Instead of passing and returning `map[string]volumeInfo`, consider adding this as a field of `ReconcileAttachDetach`. This would clean this up a bit and make it more efficient.

This would also enable us to track synced state of the volume by adding a `synced bool` field to the `volumeInfo` struct instead of maintaining a separate map.

pkg/controller/attach_detach.go
```diff
@@ -599,128 +585,130 @@ func (r *ReconcileAttachDetach) recreateAzVolumeAttachment(ctx context.Context,
...
-					w.Logger().Errorf(err, "failed to get AzVolumeAttachment (%s): %v", azVolumeAttachmentName, err)
-					return syncedVolumeAttachments, volumesToSync, err
-				}
-			} else if apiVersion, ok := azureutils.GetFromMap(azVolumeAttachment.Annotations, consts.APIVersion); !ok || apiVersion != azdiskv1beta2.APIVersion {
-				w.Logger().Infof("Found AzVolumeAttachment (%s) with older api version. Converting to apiVersion(%s)", azVolumeAttachmentName, azdiskv1beta2.APIVersion)
-
-				for k, v := range desiredAzVolumeAttachment.Labels {
-					azVolumeAttachment.Labels = azureutils.AddToMap(azVolumeAttachment.Labels, k, v)
-				}
+				return syncedVolumeAttachments, volumeHandles, volumeAttributes, err
```
edreed, line 611: If there is an error processing one volume, we should log the error and continue rather than terminate the loop.

pkg/controller/attach_detach.go
```diff
@@ -599,128 +585,130 @@ func (r *ReconcileAttachDetach) recreateAzVolumeAttachment(ctx context.Context,
...
+		}
 
-				azVolumeAttachment, err = r.azClient.DiskV1beta2().AzVolumeAttachments(r.config.ObjectNamespace).Update(ctx, azVolumeAttachment, metav1.UpdateOptions{})
+		for _, dataDisk := range disks {
+			azVolumeAttachmentName := azureutils.GetAzVolumeAttachmentName(*dataDisk.Name, nodeName)
+			if syncedAzVolumeAttachments[azVolumeAttachmentName] {
+				continue
+			}
+			if _, ok := r.azVolumeAttachmentToVaMap.Load(azVolumeAttachmentName); ok {
+				// If the AzVolumeAttachment has a VolumeAttachment mapping to it, it means it's a Primary role which should be recreated already. Skip it.
```
edreed, line 654: Should we mark this volume as synced in this case?

pkg/controller/attach_detach.go
edreed: I was referring to `azcache.AzureCacheReadType`. `types.NodeName` is a Kubernetes type.

pkg/controller/attach_detach.go
```diff
@@ -741,45 +729,47 @@ func (r *ReconcileAttachDetach) recoverAzVolumeAttachment(ctx context.Context, r
...
 			defer wg.Done()
+			var publishCtx map[string]string
+			if lun, _, err := r.cloudDiskAttacher.GetDiskLun(azv.Spec.VolumeName, azv.Spec.VolumeID, types.NodeName(azv.Spec.NodeName)); err == nil {
+				publishCtx = map[string]string{"LUN": strconv.Itoa(int(lun))}
+			}
 			var targetState azdiskv1beta2.AzVolumeAttachmentAttachmentState
-			updateMode := azureutils.UpdateCRIStatus
-			if azv.Spec.RequestedRole == azdiskv1beta2.ReplicaRole {
-				updateMode = azureutils.UpdateAll
+			if r.isAttached(ctx, azv.Spec.VolumeID, azv.Spec.NodeName) {
```
edreed, line 737: Instead of checking if the disk is attached here, replace the `if (targetState == "")` statement on line 752 with an `if..else` statement, i.e.

```go
if (r.isAttached(...)) {
    targetState = azdiskv1beta2.Attached
} else {
    switch azv.Status.State {
        ...
    }
}
```

Splitting the initialization of `targetState` around the definition of the CRI update function makes it difficult to see the dependencies and could lead to errors in subsequent changes to this code.

pkg/controller/attach_detach.go
```diff
@@ -800,6 +790,45 @@ func (r *ReconcileAttachDetach) recoverAzVolumeAttachment(ctx context.Context, r
...
+func (r *ReconcileAttachDetach) isAttached(ctx context.Context, volumeHandle, nodeName string) bool {
+	w, _ := workflow.GetWorkflowFromContext(ctx)
+	disk, err := r.cloudDiskAttacher.CheckDiskExists(ctx, volumeHandle)
+	if err != nil {
+		w.Logger().Errorf(err, "Volume (%s) not found", volumeHandle)
+		return false
+	}
+	for _, managedByNodeID := range *disk.ManagedByExtended {
+		if strings.Contains(strings.ToLower(managedByNodeID), strings.ToLower(nodeName)) {
+			w.Logger().Infof("disk(%s) is attached to node(%s)", volumeHandle, nodeName)
```
edreed, line 802: Noisy trace - log at verbosity 5 or greater.

pkg/controller/shared_state.go
```diff
@@ -27,6 +27,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	er "github.com/pkg/errors"
```
edreed, line 30: Just use the std errors package. This one has been archived.

pkg/controller/shared_state.go
```diff
@@ -1818,3 +1819,121 @@ func (c *SharedState) incrementAttachmentCount(ctx context.Context, node string)
...
+		utilfeature.DefaultFeatureGate.Enabled(features.CSIMigrationAzureDisk) &&
+		pv.Spec.AzureDisk != nil {
+		if pv, err = c.translateInTreePVToCSI(pv); err != nil {
+			w.Logger().Errorf(err, "failed to translate intree pv to csi pv (%s)", volumeName)
+			return volumeHandle, volumeContext, err
+		}
+	}
+
+	if pv.Spec.CSI == nil || pv.Spec.CSI.Driver != c.config.DriverName {
+		return volumeHandle, volumeContext, er.New("PersistentVolume is not controlled by " + c.config.DriverName)
```
edreed, line 1846: Consider using `fmt.Errorf()` rather than `New()` with string appends.

pkg/azureconstants/azure_constants.go
edreed: Ah, I see this annotation has already been in use. Let's just leave the value as-is, then.

pkg/controller/attach_detach.go
alice-zheyan-yu: Done.

pkg/controller/attach_detach.go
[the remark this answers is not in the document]
alice-zheyan-yu: Done.

pkg/controller/attach_detach.go
alice-zheyan-yu: Done.

pkg/controller/attach_detach.go
alice-zheyan-yu: Done.

pkg/controller/attach_detach.go
alice-zheyan-yu: Done.

pkg/controller/attach_detach.go
alice-zheyan-yu: Done.

pkg/controller/attach_detach.go
alice-zheyan-yu: Done.

pkg/controller/attach_detach.go
alice-zheyan-yu: Done.

pkg/controller/shared_state.go
alice-zheyan-yu: Done.

pkg/controller/shared_state.go
alice-zheyan-yu: Done.

pkg/azureconstants/azure_constants.go
alice-zheyan-yu: Okay. I will leave the value as `"disk.csi.azure.com/request-starttime"`. My change is only fixing the typo from `RequestStartimeKey `to `RequestStartTimeKey`

pkg/controller/attach_detach.go
```diff
@@ -800,6 +783,45 @@ func (r *ReconcileAttachDetach) recoverAzVolumeAttachment(ctx context.Context, r
 	return nil
 }
 
+func (r *ReconcileAttachDetach) isAttached(ctx context.Context, volumeHandle, nodeName string) bool {
+	w, _ := workflow.GetWorkflowFromContext(ctx)
+	disk, err := r.cloudDiskAttacher.CheckDiskExists(ctx, volumeHandle)
```
hccheng72, line 788: [`CheckDiskExists`](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/509d944d1b21e66112bc0aaf38047131856f0397/pkg/provisioner/cloudprovisioner.go#L776) will skip the check when in throttling and returns (nil, nil). Should we also check if `disk != nil` to ensure disk's existence?

pkg/controller/attach_detach.go
```diff
@@ -741,45 +723,46 @@ func (r *ReconcileAttachDetach) recoverAzVolumeAttachment(ctx context.Context, r
...
-				targetState = azdiskv1beta2.AttachmentPending
-				innerUpdateFunc := updateFunc
-				updateFunc = func(obj client.Object) error {
-					if err := innerUpdateFunc(obj); err != nil {
-						return err
+			if r.isAttached(ctx, azv.Spec.VolumeID, azv.Spec.NodeName) {
+				targetState = azdiskv1beta2.Attached
+			} else {
+				switch azv.Status.State {
+				case azdiskv1beta2.Attaching:
```
hccheng72, line 747: Sorry, I'm not sure where the `azv.Status.State` has been set when recreating the azvolumeattachment. But is it possible that a primary azvolumeattachment's state is Attached, however, `isAttached` function returns false due to async attach feature? In this case, should we change its state to AttachmentPending as well to redo the attach operation?

alice-zheyan-yu: When we call `recreateAzVolumeAttachment()`, a create event will be triggered and `Reconcile()` can catch the event in this line:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/448acb3466b2ae888cdf3aa271bab9f911c18aa2/pkg/controller/attach_detach.go#L138
Then `triggerAttach()` will update status.

alice-zheyan-yu: Thanks for bringing up this corner case. To fix it, I will add the check `isAttached()` before `recreateAzVolumeAttachment()`. If `isAttached()` returns false, I will not recreate it.

pkg/controller/attach_detach.go
alice-zheyan-yu: Done.

pkg/controller/attach_detach.go
sunpa93: It's a possible scenario for the driver to crash in the middle of an attach operation, in which case it's very possible for the `AzVolumeAttachment.Status.State` and the actual attach state of the disk to be inconsistent (i.e. `AzVolumeAttachment.Status.State` would be `Attaching` when the disk is, in fact, attached.

The driver currently handles this scenario by updating the state from `Attaching` to `AttachmentPending` to restart the attach operation to update the state but given the new feature this PR adds to check the attach state of the disk, we can directly infer and update the state of the `AzVolumeAttachment` CRI at the recovery stage instead of relying on an additional + unnecessary attach operation on an already attached disk.

pkg/controller/attach_detach.go
```diff
@@ -573,154 +559,160 @@ func reportError(azVolumeAttachment *azdiskv1beta2.AzVolumeAttachment, state azd
...
+	volumeHandle, volumeContext, volumeInfo, hasSynced, err := r.loadOrStoreVolumeInfo(ctx, volumeName, azVolumeAttachmentName)
+	if hasSynced {
+		// this azVolumeAttachment has already been recreated
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	attached, err := r.isAttached(ctx, volumeHandle, nodeName)
+	if err != nil {
```
alice-zheyan-yu, line 667: If the err is `getDisk` throttled, current code will return err. No azVolumeAttachment will be created. Then we have retry logic which will pick up the failed cases. Disadvantage: if all retries get throttled, we may miss the `azVolumeAttachment`. The recovery process can't recover all `azVolumeAttachment`s. I'm not sure how bad will it affect our driver's job.

Another way:
If the err is `getDisk` throttled, we still create the `azVolumeAttachment`. In the `recoverAzVolumeAttachment()` we re-check `isAttached()`. If it's not attached, we detach and delete the `azVolumeAttachment`. If it's still throttled, we just run normal recover process. Disadvantage: we may end up with creating an extra `azVolumeAttachment`. Advantage: we will not miss any `azVolumeAttachment`.

@edreed @sunpa93 @hccheng72 Please feel free to provide your input here.

pkg/controller/attach_detach.go
```diff
@@ -573,154 +559,162 @@ func reportError(azVolumeAttachment *azdiskv1beta2.AzVolumeAttachment, state azd
...
+	}
+	attached, err := r.isAttached(ctx, volumeHandle, nodeName)
+	if !attached && err == nil {
+		// disk is not attached to node. no need to recreate AzVolumeAttachment.
+		return nil
+	}
+	if err != nil {
+		// when it failed to check if the disk is attached, we can still recreate the AzVolumeAttachment based on the previous information we got.
+		// so we will not miss any AzVolumeAttachment
+		w.Logger().Errorf(err, "failed to check if disk (%s) is attached to node (%s).", volumeHandle, nodeName)
```
hccheng72, line 674: Besides failed check, I think the error also could be "the volume is not found" that we don't want to create AzVolumeAttachment.

pkg/controller/attach_detach.go
```diff
@@ -800,6 +796,56 @@ func (r *ReconcileAttachDetach) recoverAzVolumeAttachment(ctx context.Context, r
...
+		// The rate limit of Get Disk call is 15000 times per 3 mins. We shouldn't get throttled often.
+		return false, fmt.Errorf("getting disk is throttled. %s", azureconstants.RateLimited)
+	}
+	if err != nil {
+		w.Logger().Errorf(err, "Volume (%s) not found", volumeHandle)
+		return false, err
+	}
+	// there is possibility that disk is nil when it is throttled
+	// don't check disk state when GetDisk is throttled
+	if disk != nil {
```
hccheng72, line 812: When it's throttled, the function will return in line 804, so here `disk` won't be `nil`. We may keep the if statement for safe but update the comment to avoid confusion.

pkg/controller/attach_detach.go
```diff
@@ -573,154 +559,162 @@ func reportError(azVolumeAttachment *azdiskv1beta2.AzVolumeAttachment, state azd
...
+		}
 
-				for k, v := range azVolumeAttachment.Annotations {
-					azVolumeAttachment.Status.Annotations = azureutils.AddToMap(azVolumeAttachment.Status.Annotations, k, v)
-				}
+		for _, dataDisk := range disks {
+			volumeName := strings.ToLower(*dataDisk.Name)
+			azVolumeAttachmentName := azureutils.GetAzVolumeAttachmentName(volumeName, nodeName)
+			err = r.recreateAzVolumeAttachmentFromVolumeInfo(ctx, "", azVolumeAttachmentName, volumeName, nodeName)
+			w.Logger().Errorf(err, "failed to recreate replica AzVolumeAttachment (%s) from volumeInfo.", azVolumeAttachmentName)
```
edreed, line 614: We shouldn't log an error if `err == nil`.

pkg/controller/attach_detach.go
```diff
@@ -573,154 +559,162 @@ func reportError(azVolumeAttachment *azdiskv1beta2.AzVolumeAttachment, state azd
...
+		}
+	}
+	return err
+}
 
-				// for now, we don't empty the meta annotatinos after migrating them to status annotation for safety.
-				// note that this will leave some remnant garbage entries in meta annotations
+func generateVolumeInfo(volumeHandle string, volumeContext map[string]string) volumeInfo {
```
edreed, line 620: `newVolumeInfo`

pkg/controller/attach_detach.go
```diff
@@ -70,6 +82,7 @@ type ReconcileAttachDetach struct {
 	cloudDiskAttacher CloudDiskAttachDetacher
 	stateLock         *sync.Map
 	retryInfo         *retryInfo
+	volumeInfos       *sync.Map
```
edreed, line 85: If this is only used during `Recover`, do we really need to use `sync.Map`? It doesn't seem to be accessed by multiple goroutines, so it seems like overkill.

pkg/controller/attach_detach.go
```diff
@@ -573,154 +559,162 @@ func reportError(azVolumeAttachment *azdiskv1beta2.AzVolumeAttachment, state azd
...
+		volumeContext = vInfo.attributes
+	} else {
+		w.Logger().Infof("getting volumeHandle and volumeContext from PersistentVolume for volume (%s)", volumeName)
+		volumeHandle, volumeContext, err = r.getVolumeInfoFromPV(ctx, volumeName)
+		if err != nil {
+			w.Logger().Errorf(err, "failed to find volume info from PersistentVolume for volume (%s)", volumeName)
+			return "", nil, vInfo, hasSynced, err
+		}
+		vInfo = generateVolumeInfo(volumeHandle, volumeContext)
+		r.volumeInfos.Store(volumeName, vInfo)
```
edreed, line 651: In general when using `sync.Map`, you need to store using `LoadOrStore`. Otherwise, you risk overwriting the data another goroutine has added.

The general usage pattern for `sync.Map` is:

```go
entry, ok := syncMap.Load(key)
if !ok {
    entry, _ := syncMap.LoadOfStore(key, newEntry())
}
```

pkg/controller/attach_detach.go
edreed: Since the error from `isAttached` is ignored, it should be logged at `Info` level.

pkg/controller/attach_detach.go
```diff
@@ -573,154 +559,162 @@ func reportError(azVolumeAttachment *azdiskv1beta2.AzVolumeAttachment, state azd
...
+		role = azdiskv1beta2.PrimaryRole
+	}
+	desiredAzVolumeAttachment := r.generateDesiredAzVolumeAttachment(azVolumeAttachmentName, nodeName, volumeName, volumeHandle, role, volumeContext)
+	err = r.recreateAzVolumeAttachment(ctx, azVolumeAttachmentName, desiredAzVolumeAttachment)
+	if err != nil {
+		w.Logger().Errorf(err, "failed to recreate AzVolumeAttachment (%s)", azVolumeAttachmentName)
+		return err
+	}
+	volumeInfo.syncedAzVolumeAttachments[azVolumeAttachmentName] = true
+	r.volumeInfos.Store(volumeName, volumeInfo)
```
edreed, line 689: If you store the `volumeInfo` in the map by pointer rather than by value, you wouldn't have to do an update here.

pkg/controller/attach_detach.go
```diff
@@ -741,45 +735,47 @@ func (r *ReconcileAttachDetach) recoverAzVolumeAttachment(ctx context.Context, r
 		wg.Add(1)
 		go func(azv azdiskv1beta2.AzVolumeAttachment, azvMap *sync.Map) {
 			defer wg.Done()
+			var publishCtx map[string]string
+			if lun, _, err := r.cloudDiskAttacher.GetDiskLun(azv.Spec.VolumeName, azv.Spec.VolumeID, types.NodeName(azv.Spec.NodeName)); err == nil {
+				publishCtx = map[string]string{"LUN": strconv.Itoa(int(lun))}
```
edreed, line 740: Use `azureconstants.LUN` instead of `"LUN"` here.

hccheng72: Then here should be updated as well.
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/509d944d1b21e66112bc0aaf38047131856f0397/pkg/provisioner/cloudprovisioner.go#L475

pkg/controller/attach_detach.go
```diff
@@ -741,53 +734,64 @@ func (r *ReconcileAttachDetach) recoverAzVolumeAttachment(ctx context.Context, r
...
-
-			if _, err := azureutils.UpdateCRIWithRetry(ctx, nil, r.cachedClient, r.azClient, &azv, updateFunc, consts.ForcedUpdateMaxNetRetry, updateMode); err != nil {
-				w.Logger().Errorf(err, "failed to update AzVolumeAttachment (%s) for recovery", azv.Name)
-			} else {
-				// if update succeeded, add the CRI to the recoveryComplete list
+			} else if !attached && err == nil {
+				// disk is not attached. Delete the azVolumeAttachment
+				shouldDelete = true
+				w.Logger().Infof("need to detach AzVolumeAttachment (%s)", azv.Name)
+				markDetachRequest(&azv, attachdetach)
```
alice-zheyan-yu, line 761: I'm wondering should I use `updateFunc` in `UpdateCRIWithRetry()` to `markDetachRequest` instead of updating the object directly here.

sunpa93: Modifying the object here likely will not result in a expected object update. As the update call actually happens in 788, where `UpdateCRIWithRetry` is called with the defined update func. Consider using below function to append to existing update func:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/38590912a5baddba38b40cb94922346ff73592f3/pkg/azureutils/azure_disk_utils.go#L1254

pkg/controller/attach_detach.go
```diff
@@ -587,154 +572,160 @@ func reportError(azVolumeAttachment *azdiskv1beta2.AzVolumeAttachment, state azd
...
 
-				// update status
-				_, err = r.azClient.DiskV1beta2().AzVolumeAttachments(r.config.ObjectNamespace).UpdateStatus(ctx, azVolumeAttachment, metav1.UpdateOptions{})
-				if err != nil {
-					w.Logger().Errorf(err, "failed to update status of AzVolumeAttachment (%s) for volume (%s) and node (%s): %v", azVolumeAttachmentName, *pvName, nodeName, err)
-					return syncedVolumeAttachments, volumesToSync, err
-				}
-			}
+	// try to recreate missing AzVolumeAttachment CRI with Primary role from VolumeAttachment
+	for i := 0; i < maxRetry; i++ {
```
edreed, line 709: Retrying in tight loops is problematic because:

1. it's like to fail if the issue is transient and the retry immediate
2. it floods the API server with requests

I'm honestly not sure we should retry and instead fail the recovery to let another replica take over. If we do retry, we need to queue the failed replicas and retry with a delay.

alice-zheyan-yu: I totally agree that we could have a better retry strategy. To do that I need more test data to support our solution. Since this PR was created last year and its main purpose is to recreate replica attachments. I'd like to merge it first so we can test it in a large scale and receive more feedback. In the meanwhile, I can add a TODO or create a GitHub issue to track task of improving the retry policy instead of blocking this PR.

However, for the worst case, the replica AzVolumeAttachment can't be recovered. When pod fails, it doesn't have replica attachment which may slow down our failover time. Then it seems like it may block us from launch which is not acceptable. If that's the case, l would agree we need a better retry strategy before merging this PR.

Let me know your thoughts.

pkg/controller/shared_state.go
```diff
@@ -1820,3 +1820,122 @@ func (c *SharedState) incrementAttachmentCount(ctx context.Context, node string)
...
+			azVolumeAttachment.Status.Annotations = azureutils.AddToMap(azVolumeAttachment.Status.Annotations, k, v)
+		}
+
+		// for now, we don't empty the meta annotations after migrating them to status annotation for safety.
+		// note that this will leave some remnant garbage entries in meta annotations
+		for k, v := range desiredAzVolumeAttachment.Annotations {
+			azVolumeAttachment.Annotations = azureutils.AddToMap(azVolumeAttachment.Status.Annotations, k, v)
+		}
+
+		azVolumeAttachment, err = c.azClient.DiskV1beta2().AzVolumeAttachments(c.config.ObjectNamespace).Update(ctx, azVolumeAttachment, metav1.UpdateOptions{})
```
alice-zheyan-yu, line 1921: Here we are calling Update() directly. Is it better to use UpdateCRIWithRetry() instead? So, we can retry if it fails.

pkg/controller/shared_state.go
```diff
@@ -1820,3 +1820,122 @@ func (c *SharedState) incrementAttachmentCount(ctx context.Context, node string)
...
+		}
+	} else {
+		statusUpdateRequired = false
+	}
+
+	if volumeAttachment, ok := c.azVolumeAttachmentToVaMap.Load(azVolumeAttachmentName); ok && statusUpdateRequired {
+		// need to add consts.VolumeAttachmentKey annotation for newly created azVolumeAttachment and migrate ObjectMeta.Annotations to Status.Annotations if version just got updated to v1beta2
+		azVolumeAttachment.Status.Annotations = azureutils.AddToMap(azVolumeAttachment.Status.Annotations, consts.VolumeAttachmentKey, volumeAttachment.(string))
+		// update status
+		_, err = c.azClient.DiskV1beta2().AzVolumeAttachments(c.config.ObjectNamespace).UpdateStatus(ctx, azVolumeAttachment, metav1.UpdateOptions{})
```
alice-zheyan-yu, line 1934: Same. Use UpdateCRIWithRetry()?

pkg/controller/attach_detach.go
edreed: I would recommend dropping the retries in this PR. As I mentioned, retrying immediately is likely to fail anyway and I'd rather avoid the potential for a tight retry loop to overwhelm API server. Let's continue to discuss how to handle failures on recovery to come up with a good solution.

pkg/controller/shared_state.go
edreed: Yes.

pkg/controller/shared_state.go
edreed: Yes.

pkg/controller/attach_detach.go
```diff
@@ -587,154 +572,160 @@ func reportError(azVolumeAttachment *azdiskv1beta2.AzVolumeAttachment, state azd
...
+		if err = r.recreateAzVolumeAttachmentWithPrimaryRole(ctx); err == nil {
+			break
+		}
+		w.Logger().Errorf(err, "failed to recreate missing AzVolumeAttachment CRI with primary role from VolumeAttachment with retry number (%d).", i)
+	}
 
-			syncedVolumeAttachments[volumeAttachment.Name] = true
+	// try to recreate replica AzVolumeAttachment that had failed to detach for any reason during driver uninstall
+	for i := 0; i < maxRetry; i++ {
+		if err = r.recreateAzVolumeAttachmentWithReplicaRole(ctx); err == nil {
```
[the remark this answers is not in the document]
alice-zheyan-yu: Maybe it's better to call tryCreateFailedReplicas()?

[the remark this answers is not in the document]
edreed: I don't believe so. The `manageReplicas` relies on the current `AzVolumeAttachment` instances to determine how many replicas need to be attached. If we fail to recover an existing replica, `manageReplicas` would not be able to create new replica attachments because the `maxShares` count will be exhausted.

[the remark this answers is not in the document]
edreed: `tryCreateFailedReplicas` invokes `manageReplicas` and so also relies on the `AzVolumeAttachments` being in the correct state.

pkg/controller/attach_detach.go
alice-zheyan-yu: How about use the same way as we do for `UpdateCRIWithRetry`?
```
err = retry.OnError(
		wait.Backoff{
			Duration: consts.CRIUpdateRetryDuration,
			Factor:   consts.CRIUpdateRetryFactor,
			Steps:    consts.CRIUpdateRetryStep,
			Cap:      consts.DefaultBackoffCap,
		},
		isRetriable,
		conditionFunc,
	)
```

edreed: I believe that to be a workable solution.

pkg/controller/shared_state.go
```diff
@@ -1820,3 +1820,122 @@ func (c *SharedState) incrementAttachmentCount(ctx context.Context, node string)
...
+		}
+
+		for k, v := range azVolumeAttachment.Annotations {
+			azVolumeAttachment.Status.Annotations = azureutils.AddToMap(azVolumeAttachment.Status.Annotations, k, v)
+		}
+
+		// for now, we don't empty the meta annotations after migrating them to status annotation for safety.
+		// note that this will leave some remnant garbage entries in meta annotations
+		for k, v := range desiredAzVolumeAttachment.Annotations {
+			azVolumeAttachment.Annotations = azureutils.AddToMap(azVolumeAttachment.Status.Annotations, k, v)
```
alice-zheyan-yu, line 1918: Seems like we need to change it to be `azVolumeAttachment.Annotations = azureutils.AddToMap(azVolumeAttachment.Annotations, k, v)`

alice-zheyan-yu:
I'll add unit tests for `RecreateAzVolumeAttachmentsWithRetry`.

pkg/controller/attach_detach.go
```diff
@@ -587,154 +572,157 @@ func reportError(azVolumeAttachment *azdiskv1beta2.AzVolumeAttachment, state azd
...
-					VolumeName:    volumeName,
-					VolumeID:      pv.Spec.CSI.VolumeHandle,
-					NodeName:      nodeName,
-					RequestedRole: azdiskv1beta2.PrimaryRole,
-					VolumeContext: pv.Spec.CSI.VolumeAttributes,
-				},
+			volumeName := strings.ToLower(*pvName)
+			azVolumeAttachmentName := azureutils.GetAzVolumeAttachmentName(volumeName, nodeName)
+			if err = r.recreateAzVolumeAttachmentFromVolumeInfo(ctx, volumeAttachment.Name, azVolumeAttachmentName, volumeName, nodeName); err != nil {
+				w.Logger().Errorf(err, "failed to recreate primary AzVolumeAttachment (%s) from volumeInfo.", azVolumeAttachmentName)
```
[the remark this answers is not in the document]
edreed: Let's defer retry for a subsequent PR when we can come up with a workable strategy.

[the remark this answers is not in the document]
alice-zheyan-yu: Sure. I created this GitHub issue: https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/1773 to track. And I'll update this PR to drop the retry.

PR #1667, opened 2022-12-21
[V2] [WIP] chore: add PartialUpdateError logic to cloudprovisioner

sunpa93:
**What type of PR is this?**
/kind cleanup

* For proper handling of `PartialUpdateError`, `cloudprovisioner` should also be able to emit `PartialUpdateError`

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/provisioner/cloudprovisioner.go
+++ b/pkg/provisioner/cloudprovisioner.go
@@ -44,6 +44,7 @@ import (
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/workflow"
 	azcache "sigs.k8s.io/cloud-provider-azure/pkg/cache"
 	azure "sigs.k8s.io/cloud-provider-azure/pkg/provider"
+	"sigs.k8s.io/cloud-provider-azure/pkg/retry"
 )
 
 var (
@@ -408,7 +409,9 @@ func (c *CloudProvisioner) PublishVolume(
 		if vmState != nil && strings.ToLower(*vmState) == "failed" {
 			w.Logger().Infof("VM(%q) is in failed state, update VM first", nodeName)
 			if err = c.cloud.UpdateVM(ctx, nodeName); err != nil {
-				err = status.Errorf(codes.Internal, "update instance %q failed with %v", nodeName, err)
+				if _, ok := err.(*retry.PartialUpdateError); !ok {
+					err = status.Errorf(codes.Internal, "update instance %q failed with %v", nodeName, err)
+				}
 				return
 			}
 		}
--- a/go.mod
+++ b/go.mod
@@ -186,5 +186,5 @@ replace (
 	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.26.0
 	k8s.io/sample-controller => k8s.io/sample-controller v0.26.0
 
-	sigs.k8s.io/cloud-provider-azure => github.com/edreed/cloud-provider-azure v0.7.3-0.20230109174017-9397319d7638
+	sigs.k8s.io/cloud-provider-azure => github.com/sunpa93/cloud-provider-azure v0.7.3-0.20230208121346-f6f84d13f3fd
 )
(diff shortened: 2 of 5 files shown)


sunpa93:
/hold
This PR currently contains unmerged changes from `cloud-provider-azure` for testing purposes.

pkg/provisioner/cloudprovisioner.go
```diff
@@ -406,9 +408,22 @@ func (c *CloudProvisioner) PublishVolume(
 
 	if err == nil {
 		if vmState != nil && strings.ToLower(*vmState) == "failed" {
+			errWrapperFunc := func(future *azureautorest.Future) {
```
landreasyan, line 411: I am not sure we should be exposing all this logic here. Could we create a wrapper around UpdateVMAsync in the cloud provider azure and call that method instead?

landreasyan: I think it makes sense to have an Update method in the cloud provider azure that returns a compatible output to Attach.

landreasyan: We could parse and return the error here: https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/1ef4a7d7b5a5f3148acf37afb15b014376706390/vendor/sigs.k8s.io/cloud-provider-azure/pkg/provider/azure_controller_vmss.go#L129

sunpa93: We already have a generic logic for this implemented in https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/1ef4a7d7b5a5f3148acf37afb15b014376706390/vendor/sigs.k8s.io/cloud-provider-azure/pkg/provider/azure_controller_common.go#L379 to handle/return partial update errors. If your ask to is to move this logic a layer down to vmss controller's `WaitForUpdateResult` implementation, I do understand the motivation as we have only seen this Preemption issue occur with VMSS only but at the same time, I do feel that it would be good to observe/track `PartialUpdateError` occurring in update operations for other VMSets, such as VMSS Flex and AvSets as well, and having this logic in `controllerCommon` would allow us to do so.

landreasyan: I am more worried about the inconsistent error processing for the API calls. I the Attach call returns `Partial Update` then Update should too as behind the scenes they are the same operation. So, in the code for both the driver and cloud provider there will be confusion on why is this handled only for Attach or Update and not for all? This adds an indirect dependency between two packages which makes maintaining the code confusing and error prone. We should either do the error parsing for `PartialError` in the driver or cloud provider.

sunpa93: I see where your concern leads. Nonetheless, API calls exposed by `vmset` interface do not have inconsistent error handling for `AttachDisk` and `UpdateVM`. That's because `vmset.AttachDisk` is an asynchronous call that requires the caller to call `WaitForUpdateResult` to poll for update result. This means the caller of `vmset.AttachDisk` is expected to call `vmset.WaitForUpdateResult` and handle/parse the raw error returned by it. In our case, the caller of `vmset.AttachDisk` is `controllerCommon`, and `controllerCommon` has error handling implemented in `controllerCommon.WaitForUpdateResult`.

A more appropriate change, I believe, would be to create and expose `UpdateVM` in `controllerCommon` that internally calls `controllerCommon.WaitForUpdateResult`. And those that require this level of abstraction + error handling should consume `controllerCommon.UpdateVM` just like how they call `controllerCommon.AttachDisk` instead of directly accessing the raw vmset interface.

sunpa93: I see where your concern leads. Nonetheless, API calls exposed by `vmset` interface do not have inconsistent error handling for `AttachDisk` and `UpdateVM`. That's because `vmset.AttachDisk` is an asynchronous call that requires the caller to call `WaitForUpdateResult` to poll for update result. This means the caller of `vmset.AttachDisk` is expected to call `vmset.WaitForUpdateResult` and handle/parse the raw error returned by it. In our case, the caller of `vmset.AttachDisk` is `controllerCommon`, and `controllerCommon` has error handling implemented in `controllerCommon.WaitForUpdateResult`.

A more appropriate change, I believe, would be to create and expose `UpdateVM` in `controllerCommon` that internally calls `controllerCommon.WaitForUpdateResult`. And those that require this level of abstraction + error handling should consume `controllerCommon.UpdateVM` just like how they call `controllerCommon.AttachDisk` instead of directly accessing the raw vmset interface.

landreasyan: Yep, that was exactly my point in the top comment.

sunpa93:
Official dependency update in https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1734

PR #1671, opened 2022-12-21
[V2] Fix: update uninstall path in crdProvisioner.DeleteVolume

### LINKED ISSUE #1664: [V2] driver uninstallation timeout error caused by failing to delete azvolume CRIs
**What happened**:
If the AzVolume CRI deletion failed (volumeDeleteRequested) and then we try to uninstall the driver, we will get a timeout error.
One reason is `azVolume.Status.State` can't be updated by crd provisioner because of the code [here](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/71dc9d2afad507276f69283410f0bd1136dd7047/pkg/provisioner/crdprovisioner.go#L390).
On the other hand, I guess no update event will be triggered for azvolume controller when uninstalling the driver since the deletion timestamp is already there. Even if an update event is triggered, an error will be returned because [changing state](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/be0734b56150b05060a9ff17ff27c5f6de27ce99/pkg/controller/azvolume.go#L238) from `DeletionFailed` to `Deleting` isn't allowed.

**How to reproduce it**:

**Environment**:

- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hccheng72:
**What type of PR is this?**
/kind bug

1. update uninstall path in `crdProvisioner.DeleteVolume`
2. remove a non-used parameter from `shared_stated.getNodesForReplica`
3. fix a log typo

Fixes #1664

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/shared_state.go
+++ b/pkg/controller/shared_state.go
@@ -1251,7 +1251,7 @@ func (c *SharedState) manageReplicas(ctx context.Context, volumeName string) err
 	}
 
 	desiredReplicaCount, currentReplicaCount := azVolume.Spec.MaxMountReplicaCount, len(azVolumeAttachments)
-	w.Logger().Infof("Control number of replicas for volume (%s): desired=%d,\tcurrent:%d", azVolume.Spec.VolumeName, desiredReplicaCount, currentReplicaCount)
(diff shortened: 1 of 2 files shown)


sunpa93 (approved):
/lgtm

landreasyan (approved):

PR #1677, opened 2022-12-30
[V2] fix: make uninstallation compatible with backload DELETE change

sunpa93:
**What type of PR is this?**
/kind bug

* The current uninstallation logic has conflicts with changes introduced in https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1553.
* This PR addresses this issue and makes some refactoring to cleanup logic.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/shared_state.go
+++ b/pkg/controller/shared_state.go
@@ -20,6 +20,7 @@ import (
 	"container/list"
 	"context"
 	"fmt"
+	"reflect"
 	"sort"
 	"strings"
 	"sync"
@@ -1132,27 +1133,34 @@ func (c *SharedState) cleanUpAzVolumeAttachmentByNode(ctx context.Context, azDri
 	return attachments.Items, nil
 }
 
-func (c *SharedState) cleanUpAzVolumeAttachments(ctx context.Context, attachments []azdiskv1beta2.AzVolumeAttachment, cleanUp attachmentCleanUpMode, caller operationRequester) error {
+func (c *SharedState) cleanUpAzVolumeAttachments(ctx context.Context, attachments []azdiskv1beta2.AzVolumeAttachment, cleanUpMode attachmentCleanUpMode, caller operationRequester) error {
 	var err error
 
 	for _, attachment := range attachments {
-		var patchRequired bool
 		patched := attachment.DeepCopy()
 
-		// if caller is azdrivernode, don't append cleanup annotation
-		if (caller != azdrivernode && !metav1.HasAnnotation(patched.ObjectMeta, consts.CleanUpAnnotation)) ||
-			// replica attachments should always be detached regardless of the cleanup mode
-			((cleanUp == cleanUpAttachment || patched.Spec.RequestedRole == azdiskv1beta2.ReplicaRole) && !metav1.HasAnnotation(patched.ObjectMeta, consts.VolumeDetachRequestAnnotation)) {
-			patchRequired = true
-			if caller != azdrivernode {
-				patched.Status.Annotations = azureutils.AddToMap(patched.Status.Annotations, consts.CleanUpAnnotation, string(caller))
+		if attachment.Spec.RequestedRole == azdiskv1beta2.PrimaryRole {
+			if cleanUpMode == cleanUpAttachment && !volumeDetachRequested(patched) {
+				markDetachRequest(patched, caller)
+			} else if deleteRequested, _ := objectDeletionRequested(&attachment); !deleteRequested {
+				// if primary azvolumeattachments are being cleaned up for driver uninstall, issue a DELETE call and continue
+				// note that this DELETE request will remove AzVolumeAttachment CRI without detaching the volume from node
+				if err = c.cachedClient.Delete(ctx, patched); err != nil {
+					return err
+				}
 			}
-			if cleanUp == cleanUpAttachment || patched.Spec.RequestedRole == azdiskv1beta2.ReplicaRole {
-				patched.Status.Annotations = azureutils.AddToMap(patched.Status.Annotations, consts.VolumeDetachRequestAnnotation, string(caller))
+		} else {
+			// replica mount should always be detached in cleanup regardless to the cleanup mode
+			if !volumeDetachRequested(patched) {
+				markDetachRequest(patched, caller)
+			}
+			// append cleanup annotation to prevent replica recreations except for when the clean up was triggered by node controller due to node failure.
+			if caller != azdrivernode && !metav1.HasAnnotation(patched.ObjectMeta, consts.CleanUpAnnotation) {
+				markCleanUp(patched, caller)
 			}
 		}
 
-		if patchRequired {
+		if !reflect.DeepEqual(attachment.Status, patched.Status) {
 			if err = c.cachedClient.Status().Patch(ctx, patched, client.MergeFrom(&attachment)); err != nil && apiErrors.IsNotFound(err) {
 				err = status.Errorf(codes.Internal, "failed to patch AzVolumeAttachment (%s)", attachment.Name)
 				return err
--- a/pkg/controller/common.go
+++ b/pkg/controller/common.go
@@ -923,6 +923,14 @@ func getSupportedZones(nodeSelectorTerms []v1.NodeSelectorTerm, topologyKey stri
 	return supportedZones
 }
 
+func markDetachRequest(attachment *azdiskv1beta2.AzVolumeAttachment, caller operationRequester) {
+	attachment.Status.Annotations = azureutils.AddToMap(attachment.Status.Annotations, consts.VolumeDetachRequestAnnotation, string(caller))
+}
+
+func markCleanUp(attachment *azdiskv1beta2.AzVolumeAttachment, caller operationRequester) {
+	attachment.Status.Annotations = azureutils.AddToMap(attachment.Status.Annotations, consts.CleanUpAnnotation, string(caller))
+}
+
 func shouldCleanUp(attachment azdiskv1beta2.AzVolumeAttachment, mode azureutils.AttachmentRoleMode) bool {
 	return mode == azureutils.AllRoles || (attachment.Spec.RequestedRole == azdiskv1beta2.PrimaryRole && mode == azureutils.PrimaryOnly) || (attachment.Spec.RequestedRole == azdiskv1beta2.ReplicaRole && mode == azureutils.ReplicaOnly)
 }


pkg/controller/shared_state.go
```diff
@@ -1135,20 +1135,37 @@ func (c *SharedState) cleanUpAzVolumeAttachmentByNode(ctx context.Context, azDri
 func (c *SharedState) cleanUpAzVolumeAttachments(ctx context.Context, attachments []azdiskv1beta2.AzVolumeAttachment, cleanUp attachmentCleanUpMode, caller operationRequester) error {
 	var err error
 
+	var requestDetach = func(attachment *azdiskv1beta2.AzVolumeAttachment, patchRequired *bool) {
```
alice-zheyan-yu, line 1138: How about name: `markDetachRequest` to align with the variable name `markCleanUp` below?

pkg/controller/shared_state.go
```diff
@@ -1135,20 +1135,37 @@ func (c *SharedState) cleanUpAzVolumeAttachmentByNode(ctx context.Context, azDri
...
-			((cleanUp == cleanUpAttachment || patched.Spec.RequestedRole == azdiskv1beta2.ReplicaRole) && !metav1.HasAnnotation(patched.ObjectMeta, consts.VolumeDetachRequestAnnotation)) {
-			patchRequired = true
-			if caller != azdrivernode {
-				patched.Status.Annotations = azureutils.AddToMap(patched.Status.Annotations, consts.CleanUpAnnotation, string(caller))
+		if attachment.Spec.RequestedRole == azdiskv1beta2.PrimaryRole {
+			if cleanUp == cleanUpAttachment && !volumeDetachRequested(patched) {
+				requestDetach(patched, &patchRequired)
+			} else if deleteRequested, _ := objectDeletionRequested(&attachment); !deleteRequested {
+				// if primary azvolumeattachments are being cleaned up for driver uninstall, issue a DELETE call and continue
+				if err = c.cachedClient.Delete(ctx, patched); err != nil {
```
alice-zheyan-yu, line 1157: What's the reason that we are calling`c.cachedClient.Delete()` instead of `markCleanUp()` like what we do in line1168 for replica attachment?

sunpa93: That's due to the backload DELETE change introduced in https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1553. Detachment of `AzVolumeAttachment` is no longer triggered by deleting the CRI but by annotating the CRI with `VolumeDetachRequestAnnotation`. Then, DELETE call is made in the `attach_detach` controller after detachment is complete.

However, in case of cleanup for driver uninstall, we don't want to detach the primary `AzVolumeAttachment` but shallowly delete the CRI only, for which we need to explicitly make the DELETE call.

pkg/controller/shared_state.go
```diff
@@ -1135,20 +1135,37 @@ func (c *SharedState) cleanUpAzVolumeAttachmentByNode(ctx context.Context, azDri
...
+	var markDetachRequest = func(attachment *azdiskv1beta2.AzVolumeAttachment, patchRequired *bool) {
+		attachment.Status.Annotations = azureutils.AddToMap(attachment.Status.Annotations, consts.VolumeDetachRequestAnnotation, string(caller))
+		*patchRequired = true
+	}
+
+	var markCleanUp = func(attachment *azdiskv1beta2.AzVolumeAttachment, patchRequired *bool) {
+		attachment.Status.Annotations = azureutils.AddToMap(attachment.Status.Annotations, consts.CleanUpAnnotation, string(caller))
+		*patchRequired = true
+	}
+
```
edreed, line 1147: Let's make `markDetachRequest` and `markCleanup` standalone functions.

pkg/controller/shared_state.go
```diff
@@ -1135,20 +1135,37 @@ func (c *SharedState) cleanUpAzVolumeAttachmentByNode(ctx context.Context, azDri
 func (c *SharedState) cleanUpAzVolumeAttachments(ctx context.Context, attachments []azdiskv1beta2.AzVolumeAttachment, cleanUp attachmentCleanUpMode, caller operationRequester) error {
 	var err error
 
+	var markDetachRequest = func(attachment *azdiskv1beta2.AzVolumeAttachment, patchRequired *bool) {
+		attachment.Status.Annotations = azureutils.AddToMap(attachment.Status.Annotations, consts.VolumeDetachRequestAnnotation, string(caller))
+		*patchRequired = true
```
landreasyan, line 1140: `patchRequired` seems redundant here. We are unconditionally updating the annotation.

pkg/controller/shared_state.go
landreasyan: I believe we should create a standalone `addAnnotation` function instead.

pkg/controller/shared_state.go
```diff
@@ -1135,20 +1135,37 @@ func (c *SharedState) cleanUpAzVolumeAttachmentByNode(ctx context.Context, azDri
...
 
-		// if caller is azdrivernode, don't append cleanup annotation
-		if (caller != azdrivernode && !metav1.HasAnnotation(patched.ObjectMeta, consts.CleanUpAnnotation)) ||
-			// replica attachments should always be detached regardless of the cleanup mode
-			((cleanUp == cleanUpAttachment || patched.Spec.RequestedRole == azdiskv1beta2.ReplicaRole) && !metav1.HasAnnotation(patched.ObjectMeta, consts.VolumeDetachRequestAnnotation)) {
-			patchRequired = true
-			if caller != azdrivernode {
-				patched.Status.Annotations = azureutils.AddToMap(patched.Status.Annotations, consts.CleanUpAnnotation, string(caller))
+		if attachment.Spec.RequestedRole == azdiskv1beta2.PrimaryRole {
+			if cleanUp == cleanUpAttachment && !volumeDetachRequested(patched) {
```
landreasyan, line 1153: np: `cleanUp` -> `mode`

pkg/controller/shared_state.go
```diff
@@ -1135,20 +1135,37 @@ func (c *SharedState) cleanUpAzVolumeAttachmentByNode(ctx context.Context, azDri
...
+				markDetachRequest(patched, &patchRequired)
+			} else if deleteRequested, _ := objectDeletionRequested(&attachment); !deleteRequested {
+				// if primary azvolumeattachments are being cleaned up for driver uninstall, issue a DELETE call and continue
```
landreasyan, line 1156: np: add an explanation in the comment that the DELTE only deletes the CRI and doesn't issue detach of the disk first.

pkg/controller/shared_state.go
```diff
@@ -1135,20 +1135,37 @@ func (c *SharedState) cleanUpAzVolumeAttachmentByNode(ctx context.Context, azDri
...
+				}
+			}
+		} else {
+			// replica mount should always be detached in cleanup regardless to the cleanup mode
+			if !volumeDetachRequested(patched) {
+				markDetachRequest(patched, &patchRequired)
 			}
-			if cleanUp == cleanUpAttachment || patched.Spec.RequestedRole == azdiskv1beta2.ReplicaRole {
-				patched.Status.Annotations = azureutils.AddToMap(patched.Status.Annotations, consts.VolumeDetachRequestAnnotation, string(caller))
+			// append cleanup annotation to prevent replica recreations except for when the clean up was triggered by azdrivernode
```
landreasyan, line 1166: np: triggered by node controller due to node failure.

pkg/controller/shared_state.go
```diff
@@ -1135,20 +1135,37 @@ func (c *SharedState) cleanUpAzVolumeAttachmentByNode(ctx context.Context, azDri
...
+			} else if deleteRequested, _ := objectDeletionRequested(&attachment); !deleteRequested {
+				// if primary azvolumeattachments are being cleaned up for driver uninstall, issue a DELETE call and continue
+				if err = c.cachedClient.Delete(ctx, patched); err != nil {
+					return err
+				}
+			}
+		} else {
+			// replica mount should always be detached in cleanup regardless to the cleanup mode
+			if !volumeDetachRequested(patched) {
+				markDetachRequest(patched, &patchRequired)
```
alice-zheyan-yu, line 1164: If we markDetachRequest for replica AzVolumeAttachments, we couldn't recover them from VMs because them are detached already when we re-install driver. Then this PR will make no sense: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1666

Sould we do the same for both Replica and Primary roles here(only delete them by calling: c.cachedClient.Delete())? Don't detach any disks while uninstalling driver.

sunpa93: Replica `AzVolumeAttachments` are resources only used in v2 driver. That said, if we don't detach replica `AzVolumeAttachment` on v2 driver uninstallation, the user will end up with disk attachments unknown to k8s.

The main objective of https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1666 is so that we can properly recover `AzVolumeAttachment.Status.State` on driver reinstall. And being able to recreate replica `AzVolumeAttachment` that had failed to detach for any reason during driver uninstall is an additional benefit this PR will bring.

alice-zheyan-yu: I see! Thanks for the information.

pkg/controller/shared_state.go
sunpa93: > np: I believe we should create a standalone addAnnotation function instead because in these methods we are not doing the update to complete the marking.
I agree that a standalone `addAnnotation` function would be useful. To your second point, I am not sure if I am onboard as our code already has the naming convention, as `updateState`, `updateStatusDetail`, `removeFinalizer`, etc., that modifies the object without making the Update API call.

pkg/controller/shared_state.go
sunpa93: The annotation would only be added if certain conditions are met and if annotations hasn't yet been added in prior operations. `patchRequired` is to reduce unnecessary API call. It is either we use `patchRequired` or to run a `DeepEqual` to determine if status patch is required or not.

pkg/controller/shared_state.go
sunpa93: On second thought, I feel like `addAnnotation` would be a bit redundant as it doesn't add much extra to `AddToMap`'s functionality.

pkg/controller/shared_state.go
```diff
@@ -1136,23 +1137,30 @@ func (c *SharedState) cleanUpAzVolumeAttachments(ctx context.Context, attachment
...
+				}
+			}
+		} else {
+			// replica mount should always be detached in cleanup regardless to the cleanup mode
+			if !volumeDetachRequested(patched) {
+				markDetachRequest(patched, caller)
 			}
-			if cleanUp == cleanUpAttachment || patched.Spec.RequestedRole == azdiskv1beta2.ReplicaRole {
-				patched.Status.Annotations = azureutils.AddToMap(patched.Status.Annotations, consts.VolumeDetachRequestAnnotation, string(caller))
+			// append cleanup annotation to prevent replica recreations except for when the clean up was triggered by triggered by node controller due to node failure.
```
alice-zheyan-yu, line 1157: Duplicatated "triggered by" here. I guess you mean: "... was triggered by node controller ..."

sunpa93: Oops

pkg/controller/shared_state.go
```diff
@@ -1132,27 +1133,34 @@ func (c *SharedState) cleanUpAzVolumeAttachmentByNode(ctx context.Context, azDri
...
-			patchRequired = true
-			if caller != azdrivernode {
-				patched.Status.Annotations = azureutils.AddToMap(patched.Status.Annotations, consts.CleanUpAnnotation, string(caller))
+		if attachment.Spec.RequestedRole == azdiskv1beta2.PrimaryRole {
+			if cleanUpMode == cleanUpAttachment && !volumeDetachRequested(patched) {
+				markDetachRequest(patched, caller)
+			} else if deleteRequested, _ := objectDeletionRequested(&attachment); !deleteRequested {
+				// if primary azvolumeattachments are being cleaned up for driver uninstall, issue a DELETE call and continue
+				// note that this DELETE request will remove AzVolumeAttachment CRI without detaching the volume from node
+				if err = c.cachedClient.Delete(ctx, patched); err != nil {
```
edreed, line 1148: The deletion timestamp is already set, so this `Delete` is a no-op. According to your comment, it seems like we should remove the finalizer instead.

PR #1679, opened 2023-01-04
[V2]fix: update predicate function of node controller

hccheng72:
**What type of PR is this?**
/kind bug

When we scale out a new node, the instance type label isn't there as it's just created so the node won't be added to `availableAttachmentsMap`. This PR update the logic in predicate Update function to trigger reconciliation when the label is added.
 
Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/node.go
+++ b/pkg/controller/node.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -82,13 +83,13 @@ func (r *ReconcileNode) Reconcile(ctx context.Context, request reconcile.Request
 	}
 
 	// If the node has no DeletionTimestamp, it means it's either create event or update event
-	if n.ObjectMeta.DeletionTimestamp.IsZero() && !n.Spec.Unschedulable {
+	if n.ObjectMeta.DeletionTimestamp.IsZero() {
 		// Add the new schedulable node in availableAttachmentsMap
-		r.addNodeToAvailableAttachmentsMap(ctx, n.Name, n.GetLabels())
-
-		// Node is schedulable, proceed to attempt creation of replica attachment
-		logger.Info("Node is now available. Will requeue failed replica creation requests.")
-		r.tryCreateFailedReplicas(ctx, nodeavailability)
+		if ok := r.addNodeToAvailableAttachmentsMap(ctx, n.Name, n.GetLabels()); ok && !n.Spec.Unschedulable {
+			// Node is schedulable, proceed to attempt creation of replica attachment
+			logger.Info("Node is now available. Will requeue failed replica creation requests.")
+			r.tryCreateFailedReplicas(ctx, nodeavailability)
+		}
 	}
 
 	return reconcile.Result{}, nil
@@ -108,7 +109,8 @@ func (r *ReconcileNode) Recover(ctx context.Context, recoveryID string) error {
 	errCount := 0
 	for _, azNode := range azNodes.Items {
 		// if the corresponding node has been deleted, delete the azdrivernode object
-		_, err = r.kubeClient.CoreV1().Nodes().Get(ctx, azNode.Spec.NodeName, metav1.GetOptions{})
+		var node *v1.Node
+		node, err = r.kubeClient.CoreV1().Nodes().Get(ctx, azNode.Spec.NodeName, metav1.GetOptions{})
 		if err != nil {
 			if errors.IsNotFound(err) {
 				n := r.azClient.DiskV1beta2().AzDriverNodes(r.config.ObjectNamespace)
@@ -131,7 +133,7 @@ func (r *ReconcileNode) Recover(ctx context.Context, recoveryID string) error {
 				errCount++
 				continue
 			}
-			r.addNodeToAvailableAttachmentsMap(ctx, azNode.Name, azNode.GetLabels())
+			r.addNodeToAvailableAttachmentsMap(ctx, node.Name, node.GetLabels())
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/node.go
```diff
@@ -84,11 +86,11 @@ func (r *ReconcileNode) Reconcile(ctx context.Context, request reconcile.Request
 	// If the node has no DeletionTimestamp, it means it's either create event or update event
 	if n.ObjectMeta.DeletionTimestamp.IsZero() && !n.Spec.Unschedulable {
 		// Add the new schedulable node in availableAttachmentsMap
-		r.addNodeToAvailableAttachmentsMap(ctx, n.Name, n.GetLabels())
-
-		// Node is schedulable, proceed to attempt creation of replica attachment
-		logger.Info("Node is now available. Will requeue failed replica creation requests.")
-		r.tryCreateFailedReplicas(ctx, nodeavailability)
+		if ok := r.addNodeToAvailableAttachmentsMap(ctx, n.Name, n.GetLabels()); ok {
```
landreasyan, line 89: I think that even if the node is unscheduleable we should add it to the map to avoid skipping it in the future,

landreasyan: Since we are reconciling both on updates to the schedulable property and updates to label changes, this will work but I think we are creating some indirect dependency here and this might cause an issue if we later change the update predicate.

landreasyan, line 89: With the added check, if the node exists in the map, we will return false and will fail to create replicas if the node was uncordonded

hccheng72: Do you mean `!n.Spec.Unschedulable` in line 87? Based on existing code, if the node becomes schedulable from unschedulable, it will be an Update event and add the node to map.

hccheng72: Exactly, I'll change it to return true if the node is in the map anyway.

alice-zheyan-yu: if addNodeToAvailableAttachmentsMap() returns false here, should we do something like "return reconcile.Result{Requeue: true}, err" based on the err. So we can re-try addNodeToAvailableAttachmentsMap() if previously it failed because the label was not there yet.

hccheng72: I'm worrying if the error is unresolvable, will this make an unlimited retry?

pkg/controller/node.go
```diff
@@ -166,25 +168,32 @@ func NewNodeController(mgr manager.Manager, controllerSharedState *SharedState)
...
 				}
-			}
 
-			if oldOk && newOk && wasUnschedulable && nowSchedulable {
-				return true
+				// update event when LabelInstanceTypeStable is added to node
+				_, oldLableOk := old.GetLabels()[v1.LabelInstanceTypeStable]
+				_, newLabelOk := new.GetLabels()[v1.LabelInstanceTypeStable]
+
+				if (wasUnschedulable && nowSchedulable) || (!oldLableOk && newLabelOk) {
```
sunpa93, line 194: This can be truncated to `return (wasUnschedulable && nowSchedulable) || (!oldLableOk && newLabelOk)`

pkg/controller/node.go
alice-zheyan-yu: Yes, I agree with you. No need to retry.

hccheng72: I'll remove the `!n.Spec.Unschedulable` check before adding the node to the map.

pkg/controller/node.go
```diff
@@ -82,13 +83,13 @@ func (r *ReconcileNode) Reconcile(ctx context.Context, request reconcile.Request
 	}
 
 	// If the node has no DeletionTimestamp, it means it's either create event or update event
-	if n.ObjectMeta.DeletionTimestamp.IsZero() && !n.Spec.Unschedulable {
+	if n.ObjectMeta.DeletionTimestamp.IsZero() {
```
alice-zheyan-yu, line 86: How about adding more checks here, like below:
if n.ObjectMeta.DeletionTimestamp.IsZero() && !n.Spec.Unschedulable && LabelInstanceTypeStable_is_added _to_node == true {
Because only when schedulable and label_added, addNodeToAvailableAttachmentsMap and tryCreateFailedReplicas can pass.

hccheng72: We detect if the label is added from Update predicate, I don't think we need to check `LabelInstanceTypeStable_is_added _to_node == true` here.

landreasyan: I am not sure we should tie node "schedulablility" and capacity tracking. I think we should add the node to the map if the deletion timestamp is not set and if the label is added (latter is managed by the predicate). We could, however, check if it is schedulable before calling `tryCreateFailedReplicas`.

pkg/controller/node.go
```diff
@@ -82,13 +83,13 @@ func (r *ReconcileNode) Reconcile(ctx context.Context, request reconcile.Request
...
 		// Add the new schedulable node in availableAttachmentsMap
-		r.addNodeToAvailableAttachmentsMap(ctx, n.Name, n.GetLabels())
-
-		// Node is schedulable, proceed to attempt creation of replica attachment
-		logger.Info("Node is now available. Will requeue failed replica creation requests.")
-		r.tryCreateFailedReplicas(ctx, nodeavailability)
+		if added := r.addNodeToAvailableAttachmentsMap(ctx, n.Name, n.GetLabels()); added {
```
landreasyan, line 88: np: "added" -> "ok" because we will return true if the element is already in the map

pkg/controller/node.go
alice-zheyan-yu: If we don't add the check `LabelInstanceTypeStable_is_added _to_node == true`, when `schedulable==true` and `LabelInstanceTypeStable_is_added _to_node == false`, `addNodeToAvailableAttachmentsMap()` will be called which is unnecessary(because it will fail for sure).

alice-zheyan-yu: Hi @landreasyan if LabelInstanceTypeStable_is_added _to_node == false, addNodeToAvailableAttachmentsMap will fail at this line:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/7d15002e62d9ab7ea8f0fc43b4396b4528e36749/pkg/azureutils/azure_disk_utils.go#L222

landreasyan: I am wondering if we should even return true in the reconciler if the label is not set yet?

hccheng72: I think it makes more sense to check the label inside `addNodeToAvailableAttachmentsMap()`. In `addNodeToAvailableAttachmentsMap()`, we check the node capacity from the cache first then label, so we shouldn't stop it just by if label exists.

pkg/controller/node.go
```diff
@@ -166,25 +168,30 @@ func NewNodeController(mgr manager.Manager, controllerSharedState *SharedState)
...
 				}
-			}
 
-			if oldOk && newOk && wasUnschedulable && nowSchedulable {
-				return true
+				// update event when LabelInstanceTypeStable is added to node
+				_, oldLableOk := old.GetLabels()[v1.LabelInstanceTypeStable]
+				_, newLabelOk := new.GetLabels()[v1.LabelInstanceTypeStable]
+
+				return (wasUnschedulable && nowSchedulable) || (!oldLableOk && newLabelOk)
```
landreasyan, line 194: I am wondering if we could update the check to: 
return newLabelOk && ((wasUnschedulable && nowSchedulable) || !oldLableOk )

alice-zheyan-yu: For this case: `newLabelOk==true`, `wasUnschedulable==false`,  `nowSchedulable==true`, `oldLableOk==true`,
`return newLabelOk_true && ((wasUnschedulable_false && nowSchedulable_true) || !oldLableOk_!true ) `.  It will return false.
But it should return true, because nowSchedulable==true and newLabelOk==true which means we can retry attachments.

In addition, the readability is low. But I like the idea to keep most conditional checks in `predicate`.

hccheng72: @alice-zheyan-yu for your case, I think we want it to return false since either schedulability or label doesn't change.

alice-zheyan-yu: @hccheng72 oh I see. Yes, you are right!

pkg/controller/node.go
hccheng72: > I am wondering if we should even return true in the predicate if the label is not set yet?

but we want to catch the label update to trigger conciliation so that the node can be added to the map, right?

landreasyan: Are there any concerns with the predicate change I suggested below? Should we just go that route?

hccheng72: After discussion, we agree to keep predicate check as current iteration and add `!n.Spec.Unschedulable` before calling `tryCreateFailedReplicas`. We could optimize it in the future.

PR #1680, opened 2023-01-04
[V2] fix: azvolume recovery bug fix - status update, parameter parsing

sunpa93:
**What type of PR is this?**
/kind bug

* Make sure `AzVolume.Status` is updated when necessary for AzVolume recovery.
** Failure to do so causes an unnecessary call to `cloudprovisioner.CreateVolume`
* Make sure unsupported fields are filtered from `PV.Spec.CSI.VolumeAttributes` for populating `AzVolume.Spec.Parameters` .

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azureutils/azure_disk_utils.go
+++ b/pkg/azureutils/azure_disk_utils.go
@@ -98,6 +98,13 @@ const (
 	AllRoles
 )
 
+type FilterMode int
+
+const (
+	StrictValidation FilterMode = iota
+	IgnoreUnknown
+)
+
 type LabelPair struct {
 	Key      string
 	Operator selection.Operator
@@ -376,8 +383,10 @@ func NormalizeNetworkAccessPolicy(networkAccessPolicy string) (compute.NetworkAc
 	return "", fmt.Errorf("azureDisk - %s is not supported NetworkAccessPolicy. Supported values are %s", networkAccessPolicy, compute.PossibleNetworkAccessPolicyValues())
 }
 
-func ParseDiskParameters(parameters map[string]string) (ManagedDiskParameters, error) {
+func ParseDiskParameters(parameters map[string]string, filterMode FilterMode) (ManagedDiskParameters, error) {
 	var err error
+	var invalidParameters []string
+
 	if parameters == nil {
 		parameters = make(map[string]string)
 	}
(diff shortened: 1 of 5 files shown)


pkg/azureutils/azure_disk_utils.go
```diff
@@ -476,11 +478,19 @@ func ParseDiskParameters(parameters map[string]string) (ManagedDiskParameters, e
 			if deviceSettings, err := optimization.GetDeviceSettingFromAttribute(k); err == nil {
 				diskParams.DeviceSettings[filepath.Join(consts.DummyBlockDevicePathLinux, deviceSettings)] = v
 			} else {
-				return diskParams, fmt.Errorf("invalid parameter %s in storage class", k)
+				invalidParameters = append(invalidParameters, k)
```
edreed, line 481: We should explicitly ignore `azureconstants.RequestedSizeGib` since this parameter is [added by `CreateVolume`](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/13b204b5dc205f08bdc7f67449e01b25b401c209/pkg/provisioner/cloudprovisioner.go#L280).

pkg/azureutils/azure_disk_utils.go
```diff
@@ -476,11 +478,19 @@ func ParseDiskParameters(parameters map[string]string) (ManagedDiskParameters, e
...
 			}
 		}
 	}
-	return diskParams, nil
+
+	if len(invalidParameters) > 0 {
+		err = fmt.Errorf("invalid parameters %s in storage class", strings.Join(invalidParameters, ", "))
```
edreed, line 487: Instead on unconditionally filtering out unknown parameters **and** returning an error, this function should take an `enum` parameter that would specify either return an error on unknown parameters (`strictValidation`) or filter out unknown parameters (`ignoreUnknown`). This way, the intent and functionality is clear at the call point, and other errors can be responded to.

pkg/controller/shared_state.go
```diff
@@ -1531,14 +1530,16 @@ func (c *SharedState) createAzVolume(ctx context.Context, desiredAzVolume *azdis
...
+		statusUpdateRequired = true
 
-	} else {
-		return nil
 	}
 
-	if _, err = c.azClient.DiskV1beta2().AzVolumes(c.config.ObjectNamespace).UpdateStatus(ctx, updated, metav1.UpdateOptions{}); err != nil {
-		return err
+	if statusUpdateRequired {
+		if _, err = c.azClient.DiskV1beta2().AzVolumes(c.config.ObjectNamespace).UpdateStatus(ctx, updated, metav1.UpdateOptions{}); err != nil {
```
edreed, line 1538: Shouldn't we use `UpdateCRIWithRetry` here? If we did, we wouldn't need `statusUpdateRequired` since it would no-op if there was no change.

sunpa93: Good point.

pkg/azureutils/azure_disk_utils.go
edreed: Isn't two always going to be a possibility since `Status` cannot be updated on create if we need to create a new `AzVolume` during `Recover` e.g. in upgrade scenarios?

sunpa93: We short circuit reconciliation until recovery is complete so as long as we populate and update the status of `AzVolume` during recovery, 2) shouldn't happen.

PR #1692, opened 2023-01-13
[V2] fix: update uninstallation path in replica.go

hccheng72:
**What type of PR is this?**
/kind bug

When the driver is uninstalling and the detachment failed, we should delete the CRI and return. No need to update the CRI state.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


sunpa93 (approved):
/lgtm

PR #1709, opened 2023-01-30
Add metrics port to azuredisk-node manifests

forselli-stratio:
**What type of PR is this?**

/kind feature

Metrics are enabled for azuredisk-node but the port is not exposed in kubernetes daemonset

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node-windows.yaml
+++ b/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node-windows.yaml
@@ -130,6 +130,9 @@ spec:
             - containerPort: {{ .Values.node.livenessProbe.healthPort }}
               name: healthz
               protocol: TCP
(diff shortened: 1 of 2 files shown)


andyzhangx:
hi @forselli-stratio thanks for contribution, for node daemonset, actually we don't collect any metrics now, we only collect metrics in azure disk controller, so I am thinking we should even remove `- "--metrics-address=0.0.0.0:29605"` in azuredisk config.

forselli-stratio:
> hi @forselli-stratio thanks for contribution, for node daemonset, actually we don't collect any metrics now, we only collect metrics in azure disk controller, so I am thinking we should even remove `- "--metrics-address=0.0.0.0:29605"` in azuredisk config.

Thanks @andyzhangx , i close the PR then.

PR #1724, opened 2023-02-09
[V2] feat: add optimistic lun implementation

connorbradshaw10:
**What type of PR is this?**
/kind feature

- adds optimistic lun change from cloud provider


edreed:
kubernetes/test-infra#28865 Caused the V1 tests to be run on the V2 branch but without the needed setup. This is the cause of the additional test failures.

PR #1738, opened 2023-02-23
[V2] fix: matching scale test manifest with aurora

connorbradshaw10:
**What type of PR is this?**
/kind test

matching scale test manifest with aurora


edreed:
This is a manifest change only.

Issue #1729, opened 2023-02-15
[V2] Fix race condition: cachedClient can't Get newly created AzVolume.


alice-zheyan-yu:
**What happened**:

azvolume controller's `Recover()` will recreate `AzVolume` on driver installation. But it may fail if the `cachedClient` can't `Get` the newly created `AzVolume` with error message `pvc-xxxxxxx not found` in this line: https://github.com/alice-zheyan-yu/azuredisk-csi-driver/blob/main_v2/pkg/controller/shared_state.go#L1583. (But by far it is still ok because `Reconcile()` will take care of it. let's see what will happen next)

Since previously `func (c *SharedState) createAzVolume()` fails, it skiped `c.createOperationQueue(desiredAzVolume.Name)`. Later when pod controller's `Recover()` tries to create replica attachments (we need it when upgrade driver from V1 to V2), it fails when it Load `volumeOperationQueues` because `c.createOperationQueue(desiredAzVolume.Name)` is not called yet. (But it is still ok because pod `Reconcile()` will call `createReplicas()` even if `Recover()` fails)

**What you expected to happen**:
Although by the end, `Reconcile()`s  are smart enough to fix everything.
In order to retry less:
`azureutils.UpdateCRIWithRetry()` should be able to handle the case when `cachedClient` can't `Get` object. Should we add a logic: if `cachedClient` Get object fails, try to call other clients instead of return err directly?

**How to reproduce it**:

1, Create some AzVolumeAttachments
2, Uninstall driver V2
3, Re-Install driver V2

**Anything else we need to know?**:

**Environment**:

- CSI Driver version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sunpa93:
> In order to retry less:
azureutils.UpdateCRIWithRetry() should be able to handle the case when cachedClient can't Get object. Should we add a logic: if cachedClient Get object fails, try to call other clients instead of return err directly?

Just had a discussion with @edreed on this one, and we think there another area of optimization we can add to `UpdateCRIWithRetry`. Currently, we *always* issue a GET call before we issue an update, but we can avoid the first iteration of `GET` and use the CRI object passed into the `UpdateCRIWithRetry` call instead. If this `Update` fails due to conflict, then we can issue a `GET` starting the second iteration of the update loop.

[...]

PR #1739, opened 2023-02-23
[V2] fix: enhance UpdateCRIWithRetry to fix the cachedClient Get not found…

alice-zheyan-yu:
… issue

**What type of PR is this?**

Add one of the following kinds:
/kind bug

https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/1729

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azureutils/azure_disk_utils.go
+++ b/pkg/azureutils/azure_disk_utils.go
@@ -1118,68 +1118,35 @@ func GetAzVolumeAttachmentState(volumeAttachmentStatus storagev1.VolumeAttachmen
 
 type UpdateCRIFunc func(client.Object) error
 
-func UpdateCRIWithRetry(ctx context.Context, informerFactory azdiskinformers.SharedInformerFactory, cachedClient client.Client, azDiskClient azdisk.Interface, obj client.Object, updateFunc UpdateCRIFunc, maxNetRetry int, updateMode CRIUpdateMode) (client.Object, error) {
+func UpdateCRIWithRetry(ctx context.Context, informerFactory azdiskinformers.SharedInformerFactory, cachedClient client.Client, azDiskClient azdisk.Interface, originalObj client.Object, updateFunc UpdateCRIFunc, maxNetRetry int, updateMode CRIUpdateMode) (client.Object, error) {
 	var err error
 
 	curNetRetry := 0
 	curRetry := 0
 
-	objName := obj.GetName()
 	ctx, w := workflow.New(ctx, workflow.WithCaller(1))
 	defer func() {
 		w.AddDetailToLogger(consts.RetryKey, curRetry, consts.NetRetryKey, curNetRetry)
 		w.Finish(err)
 	}()
 
+	if err = validateParamsForUpdateCRIWithRetry(cachedClient, azDiskClient, originalObj); err != nil {
+		w.Logger().Error(err, "The parameters passed to the function UpdateCRIWithRetry are invalid.")
+		return nil, err
+	}
+
 	var updatedObj client.Object
+	objName := originalObj.GetName()
 
 	conditionFunc := func() error {
 		var err error
-		var originalObj client.Object
-		switch target := obj.(type) {
-		case *azdiskv1beta2.AzVolume:
-			if informerFactory != nil {
-				originalObj, err = informerFactory.Disk().V1beta2().AzVolumes().Lister().AzVolumes(target.Namespace).Get(objName)
-			} else if cachedClient != nil {
-				originalObj = &azdiskv1beta2.AzVolume{}
-				err = cachedClient.Get(ctx, types.NamespacedName{Namespace: target.Namespace, Name: objName}, originalObj)
-			} else {
-				originalObj, err = azDiskClient.DiskV1beta2().AzVolumes(target.Namespace).Get(ctx, objName, metav1.GetOptions{})
-			}
-		case *azdiskv1beta2.AzVolumeAttachment:
-			if informerFactory != nil {
-				originalObj, err = informerFactory.Disk().V1beta2().AzVolumeAttachments().Lister().AzVolumeAttachments(target.Namespace).Get(objName)
-			} else if cachedClient != nil {
-				originalObj = &azdiskv1beta2.AzVolumeAttachment{}
-				err = cachedClient.Get(ctx, types.NamespacedName{Namespace: target.Namespace, Name: objName}, originalObj)
-			} else {
-				originalObj, err = azDiskClient.DiskV1beta2().AzVolumeAttachments(target.Namespace).Get(ctx, objName, metav1.GetOptions{})
-			}
-		case *azdiskv1beta2.AzDriverNode:
-			if informerFactory != nil {
-				originalObj, err = informerFactory.Disk().V1beta2().AzDriverNodes().Lister().AzDriverNodes(target.Namespace).Get(objName)
-			} else if cachedClient != nil {
-				originalObj = &azdiskv1beta2.AzDriverNode{}
-				err = cachedClient.Get(ctx, types.NamespacedName{Namespace: target.Namespace, Name: objName}, originalObj)
-			} else {
-				originalObj, err = azDiskClient.DiskV1beta2().AzDriverNodes(target.Namespace).Get(ctx, objName, metav1.GetOptions{})
-			}
-		case *storagev1.VolumeAttachment:
-			if cachedClient != nil {
-				originalObj = &storagev1.VolumeAttachment{}
-				err = cachedClient.Get(ctx, types.NamespacedName{Namespace: target.Namespace, Name: objName}, originalObj)
-			} else {
-				return status.Errorf(codes.Internal, "cannot update VolumeAttachment object if controller runtime client is not provided.")
-			}
-		default:
-			return status.Errorf(codes.Internal, "object (%v) not supported.", reflect.TypeOf(target))
-		}
 
-		if err != nil {
-			if !k8serrors.IsNotFound(err) {
-				err = status.Errorf(codes.Internal, "failed to get object: %v", err)
+		if curRetry > 0 {
+			originalObj, err = getOriginalObjFromClientForUpdate(ctx, w, informerFactory, cachedClient, azDiskClient, originalObj)
+			if err != nil {
+				w.Logger().Errorf(err, "failed to get original object (%s)", objName)
+				return err
 			}
-			return err
 		}
 
 		copyForUpdate := originalObj.DeepCopyObject().(client.Object)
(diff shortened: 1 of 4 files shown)


pkg/azureutils/azure_disk_utils.go
```diff
@@ -1090,68 +1090,29 @@ func GetAzVolumeAttachmentState(volumeAttachmentStatus storagev1.VolumeAttachmen
 
 type UpdateCRIFunc func(client.Object) error
 
-func UpdateCRIWithRetry(ctx context.Context, informerFactory azdiskinformers.SharedInformerFactory, cachedClient client.Client, azDiskClient azdisk.Interface, obj client.Object, updateFunc UpdateCRIFunc, maxNetRetry int, updateMode CRIUpdateMode) (client.Object, error) {
+func UpdateCRIWithRetry(ctx context.Context, informerFactory azdiskinformers.SharedInformerFactory, cachedClient client.Client, azDiskClient azdisk.Interface, obj client.Object, originalObj client.Object, updateFunc UpdateCRIFunc, maxNetRetry int, updateMode CRIUpdateMode) (client.Object, error) {
```
edreed, line 1093: Why add `originalObj` when we already pass the original object in `obj`?

pkg/azureutils/azure_disk_utils.go
```diff
@@ -1090,68 +1090,29 @@ func GetAzVolumeAttachmentState(volumeAttachmentStatus storagev1.VolumeAttachmen
...
-				return status.Errorf(codes.Internal, "cannot update VolumeAttachment object if controller runtime client is not provided.")
-			}
-		default:
-			return status.Errorf(codes.Internal, "object (%v) not supported.", reflect.TypeOf(target))
-		}
 
-		if err != nil {
-			if !k8serrors.IsNotFound(err) {
-				err = status.Errorf(codes.Internal, "failed to get object: %v", err)
+		if originalObj == nil || needGetOriginalObj {
```
edreed, line 1111: Instead of maintaining a separate Boolean value, you can infer the need to get the original object from `curRetry`, i.e. `curRetry > 0`. This is because `curRetry` is updated on conflict errors in the `isRetriable` function below (line 1190). This would avoid the need to check the error explicitly in `conditionFunc`.

pkg/azureutils/azure_disk_utils.go
alice-zheyan-yu: I agree that `obj` and `originalObj` contain duplicated data.

However, `obj` can be the original object or updated object. This can provide some flexibility to call `UpdateCRIWithRetry()`. So, when we loss the original object we can pass nil.

For example, inside `ReconcileAttachDetach.updateVolumeAttachmentWithResult()`, we don't have original object.
```
va := storagev1.VolumeAttachment{ObjectMeta: metav1.ObjectMeta{Name: vaName}}
_, err = azureutils.UpdateCRIWithRetry(ctx, nil, r.cachedClient, r.azClient, &va, nil, vaUpdateFunc, consts.ForcedUpdateMaxNetRetry, azureutils.UpdateCRIStatus)

 ```

In addition, in `CrdProvisioner.PublishVolume()`, it seems also loss the original `attachmentObj`, because this line: https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/main_v2/pkg/provisioner/crdprovisioner.go#L591
is updating `attachmentObj.Annotations`

Also, inside `CrdProvisioner.UnpulishVolume()`, it will first `demoteVolume(ctx, azVolumeAttachmentInstance)`, then `detachVolume(ctx, azVolumeAttachmentInstance)`. `azVolumeAttachmentInstance` might be updated, and we loss the original object. For this case, it passes nil to `originalObj`, which can force `client.Get()` being called to get the original object.

I was thinking that we can pass `objInfo{Namespace, Name, Type}` instead of obj to reduce the duplicated data passing. But not sure if it worths or not.

**Comment by alice-zheyan-yu** [comment]
Will add more unit tests.

File: pkg/azureutils/azure_disk_utils.go (hunk 1 above)
Comment (edreed) [reply pkg/azureutils/azure_disk_utils.go:1093]: BTW, it's fine to have `originalObj` defined as a local variable and used to track the original object. This is needed to prevent unnecessary PUT calls to API server if the state of the object doesn't change. However, it should start out as a copy of the input `obj` and refreshed by GET on conflict.

File: pkg/azureutils/azure_disk_utils.go [hunk 3]
```diff
@@ -1165,52 +1125,49 @@ func UpdateCRIWithRetry(ctx context.Context, informerFactory azdiskinformers.Sha
 			w.Logger().V(5).Info("Skip update. No update needed.")
 			return nil
 		}
-
+		var updateStatusError error
```
edreed, line 1128: There's no need for `updateStatusError` and `updateError` now and these changes can be reverted.

alice-zheyan-yu: ah yes. I just wished to save some lines. But it should return err immediately if UpdateStatus() has err.

pkg/azureutils/azure_disk_utils.go
```diff
@@ -1279,6 +1240,68 @@ func UpdateCRIWithRetry(ctx context.Context, informerFactory azdiskinformers.Sha
...
+		if informerFactory != nil {
+			originalObj, err = informerFactory.Disk().V1beta2().AzVolumes().Lister().AzVolumes(target.Namespace).Get(objName)
+		} else if cachedClient != nil {
+			originalObj = &azdiskv1beta2.AzVolume{}
+			err = cachedClient.Get(ctx, types.NamespacedName{Namespace: target.Namespace, Name: objName}, originalObj)
+		}
+
+		if azDiskClient != nil && err != nil {
+			originalObj, err = azDiskClient.DiskV1beta2().AzVolumes(target.Namespace).Get(ctx, objName, metav1.GetOptions{})
+		}
```
hccheng72, line 1261: In the case if informerFactory, cachedClient, and azDiskClient are all `nil`, this function will return (nil, nil) and get mishandled in the calling function. Should we also return an error when azDiskClient is `nil`?

alice-zheyan-yu: Yes, good catch. And it's better to check if azDiskClient is nil first. If it's nil, we even don't need to Get the original object since we need azDiskClient to Update().

pkg/controller/shared_state.go
```diff
@@ -1536,18 +1537,24 @@ func (c *SharedState) createAzVolume(ctx context.Context, desiredAzVolume *azdis
 
 	var err error
 	var azVolume *azdiskv1beta2.AzVolume
+	originalAzVolume := azVolume.DeepCopy()
```
edreed, line 1540: `azVolume` is `nil` so this does nothing but declare a variable. You can simple do so directly.

edreed: We should need to copy the `AzVolume` object though. What's the reason for this change?

pkg/provisioner/crdprovisioner.go
```diff
@@ -517,6 +517,7 @@ func (c *CrdProvisioner) PublishVolume(
 			return nil, err
 		}
 		// if Replica AzVolumeAttachment CRI does not exist for the volume-node pair for failover, create a new Primary
+		// aliceyu??? why replica? it could be primary does not exist
```
edreed, line 520: `CrdProvisioner.PublishVolume` is called to for the primary disk attachment. If an `AzVolumeAttachment` already exists, we assume that it is replica attachment and just need to promote it.

pkg/controller/shared_state.go
alice-zheyan-yu: No special reason. I was dumb lol. I changed it back to `var originalAzVolume *azdiskv1beta2.AzVolume`. `DeepCopy()` should be done later.

pkg/provisioner/crdprovisioner.go
alice-zheyan-yu: ah thanks! I forgot to delete my own notes!

edreed:
kubernetes/test-infra#28865 Caused the V1 tests to be run on the V2 branch but without the needed setup. This is the cause of the additional test failures.

pkg/azureutils/azure_disk_utils.go
```diff
@@ -1279,6 +1246,77 @@ func UpdateCRIWithRetry(ctx context.Context, informerFactory azdiskinformers.Sha
...
+	var err error
+	var originalObj client.Object
+	objName := obj.GetName()
+
+	w.Logger().V(5).Infof("getting the original object (%s) from clients", objName)
+
+	switch target := obj.(type) {
+	case *azdiskv1beta2.AzVolume:
+		if azDiskClient == nil {
+			return nil, status.Errorf(codes.Internal, "cannot update AzVolume object if azDiskClient is not provided.")
```
edreed, line 1259: This check and error should be performed by `UpdateCRIWithRetry` since that is what would perform the update. It is not required for this method to function if `informerFactory` or `cachedClient` is provided.

alice-zheyan-yu: Yes, it would make more sense to check clients inside `UpdateCRIWithRetry`. Let me move it to line 1133, same place I checked for originalObj.

pkg/controller/shared_state.go
```diff
@@ -1536,18 +1537,24 @@ func (c *SharedState) createAzVolume(ctx context.Context, desiredAzVolume *azdis
 
 	var err error
 	var azVolume *azdiskv1beta2.AzVolume
+	var originalAzVolume *azdiskv1beta2.AzVolume
```
edreed, line 1540: `originalAzVolume` is just a copy of  `azVolume`. We can remove its usage.

alice-zheyan-yu: `azVolume` is updated in line 1587: `azVolume.Status.Annotations = azureutils.AddToMap(azVolume.Status.Annotations, statusAnnotation...)` without updating our clients. I feel like we need `originalAzVolume` to track the unchanged object, otherwise we may get `conflict` error when we update CRI. Please let me know if my understanding is wrong.

pkg/controller/shared_state.go
```diff
@@ -1571,16 +1578,18 @@ func (c *SharedState) createAzVolume(ctx context.Context, desiredAzVolume *azdis
...
 
 		updated = azVolume.DeepCopy()
-		if azVolume, err = c.azClient.DiskV1beta2().AzVolumes(c.config.ObjectNamespace).Update(ctx, updated, metav1.UpdateOptions{}); err != nil {
+		azVolume, err = c.azClient.DiskV1beta2().AzVolumes(c.config.ObjectNamespace).Update(ctx, updated, metav1.UpdateOptions{})
+		if err != nil {
 			return err
 		}
+		originalAzVolume = azVolume.DeepCopy()
 
 		azVolume.Status.Annotations = azureutils.AddToMap(azVolume.Status.Annotations, statusAnnotation...)
```
edreed, line 1587: Any updates to the target object *must* be done within the update function passed to `UpdateCRIWithRetry`. Otherwise, the update will be lost of the initial update attempt fails due to a conflict or other error.

pkg/controller/shared_state.go
edreed: This is a pre-existing bug. See my comment below.

pkg/controller/shared_state.go
edreed: Also since you now pass `originalAzVolume` instead of `updated`, this change will not be applied.

edreed: NM. I see that `updated` is used to carry the changes *into* the update function.

edreed: If you make the deep copy of `azVolume` to `updated` and then update the status annotations of `updated`, you wouldn't need the `originalAzVolume` copy also.

PR #1753, opened 2023-03-06
[V2] style: delete duplicated import of package k8s.io/api/core/v1

alice-zheyan-yu:
**What type of PR is this?**

/kind cleanup

Package "k8s.io/api/core/v1" has been imported twice. Let's delete it.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/controller/node.go
```diff
@@ -21,7 +21,6 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
-	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
```
edreed, line 24: We should use `corev1` rather than `v1`. The former is more specific and there are many `v1` packages so including `core` is important to disambiguate among them.

PR #1754, opened 2023-03-07
[V2] chore: limit max shares for all disks

NedAnd1:
**What type of PR is this?**

/kind design

Limits the max value of max shares to a common value as per PRD discussions.

**Requirements**:
  
- [x] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/docs/design-v2.md
+++ b/docs/design-v2.md
@@ -36,7 +36,7 @@ In addition to the existing StorageClass parameters described in [driver-paramet
 
 | Name | Meaning  | Available Value | Mandatory | Default value |
 |------|----------|-----------------|-----------|---------------|
-| `maxShares` | The total number of shared disk mounts allowed for the disk. Setting the value to 2 or more enables attachment replicas. | Supported values depend on the disk size. See [Share an Azure managed disk](https://docs.microsoft.com/en-us/azure/virtual-machines/disks-shared) for supported values. | No | 1 |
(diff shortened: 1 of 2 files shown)


docs/design-v2.md
```diff
@@ -36,7 +36,7 @@ In addition to the existing StorageClass parameters described in [driver-paramet
 
 | Name | Meaning  | Available Value | Mandatory | Default value |
 |------|----------|-----------------|-----------|---------------|
-| `maxShares` | The total number of shared disk mounts allowed for the disk. Setting the value to 2 or more enables attachment replicas. | Supported values depend on the disk size. See [Share an Azure managed disk](https://docs.microsoft.com/en-us/azure/virtual-machines/disks-shared) for supported values. | No | 1 |
+| `maxShares` | The total number of shared disk mounts allowed for the disk. Setting the value to 2 or more enables attachment replicas. | This value must currently be in the range `[0..3]` and must be less than the [maxShares value of the disk itself](https://docs.microsoft.com/en-us/azure/virtual-machines/disks-shared).  | No | 1 |
```
alice-zheyan-yu, line 39: I'm wondering why we limit the max value to 3. Could I learn more details? From the wiki it seems like the value can be bigger.

edreed: @alice-zheyan-yu You are correct in that the *actual* [number of allowable shares depends on the disk size](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-shared#disk-sizes), the minimum value is 3. In the V2 driver, `maxShares` is used to define the number of replica mounts and 3 should be a reasonable limit for customer workloads. We are limiting the value based on feedback from the AKS team.

PR #1756, opened 2023-03-08
[V2] fix: replica AzVolumeAttachment retries return inconsistent error type

hccheng72:
**What type of PR is this?**
/kind bug

When running the scale test with replicas, we saw that cloudprovisioner could return retriable error at the first attempt to create a replica AzVolumeAttachment, and then return not retriable error due to throttling. Under this scenario, the driver will delete the CRI without detach, so we lose the track of that disk attachment that would cause exhausting MaxShares or not being able to delete disks that are still attached to VMs. This PR is to handle inconsistent error types returned during retries.

This PR also fixes that we only add `ReplicaVolumeAttachRetryAnnotation` to replica AzVolumeAttachment.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/attach_detach.go
+++ b/pkg/controller/attach_detach.go
@@ -274,11 +274,12 @@ func (r *ReconcileAttachDetach) triggerAttach(ctx context.Context, azVolumeAttac
 
 			updateFunc := func(obj client.Object) error {
 				azva := obj.(*azdiskv1beta2.AzVolumeAttachment)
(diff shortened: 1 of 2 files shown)


sunpa93 (approved):
/lgtm

edreed:
The pull-azuredisk-csi-driver-e2e-capz-windows-2019-mainv2 test failure is unrelated to this PR.

PR #1762, opened 2023-03-09
[V2] chore: add parse max shares unit test

NedAnd1:
**What type of PR is this?**

/kind test

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azureutils/azure_disk_utils_test.go
+++ b/pkg/azureutils/azure_disk_utils_test.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"reflect"
 	"runtime"
+	"strconv"
 	"strings"
 	"syscall"
 	"testing"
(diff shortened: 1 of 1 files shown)


landreasyan:
Looking at the `GetMaxShares` and `ParseMaxShares` implementations highlighted by this PR, I noticed that there is a bit of code duplication. Can we also have `GetMaxShares` call `ParseMaxShares` instead of duplicating the logic? In that case we can have different unit tests for each of those methods, `ParseMaxShares` will be tested like we do now, but for `GetMaxShares` we should pass invalid attributes array and check that function handles that properly instead of checking the string to int parsing.

edreed:
> Looking at the `GetMaxShares` and `ParseMaxShares` implementations highlighted by this PR, I noticed that there is a bit of code duplication. Can we also have `GetMaxShares` call `ParseMaxShares` instead of duplicating the logic? In that case we can have different unit tests for each of those methods, `ParseMaxShares` will be tested like we do now, but for `GetMaxShares` we should pass invalid attributes array and check that function handles that properly instead of checking the string to int parsing.

@landreasyan The change you propose was already made in #1754.

landreasyan:
> > Looking at the `GetMaxShares` and `ParseMaxShares` implementations highlighted by this PR, I noticed that there is a bit of code duplication. Can we also have `GetMaxShares` call `ParseMaxShares` instead of duplicating the logic? In that case we can have different unit tests for each of those methods, `ParseMaxShares` will be tested like we do now, but for `GetMaxShares` we should pass invalid attributes array and check that function handles that properly instead of checking the string to int parsing.
> 
> @landreasyan The change you propose was already made in #1754.

Oh, I see, that is great! Although the change in the unit tests I was proposing is not done. We still have unit tests for `GetMaxShares` and `ParseMaxShares` that check the same thing.

NedAnd1:
> > > Looking at the `GetMaxShares` and `ParseMaxShares` implementations highlighted by this PR, I noticed that there is a bit of code duplication. Can we also have `GetMaxShares` call `ParseMaxShares` instead of duplicating the logic? In that case we can have different unit tests for each of those methods, `ParseMaxShares` will be tested like we do now, but for `GetMaxShares` we should pass invalid attributes array and check that function handles that properly instead of checking the string to int parsing.
> > 
> > 
> > @landreasyan The change you propose was already made in #1754.
> 
> Oh, I see, that is great! Although the change in the unit tests I was proposing is not done. We still have unit tests for `GetMaxShares` and `ParseMaxShares` that check the same thing.

@landreasyan Thanks for the feedback; I made your suggested changes, and hopefully it looks good.

edreed:

PR #1781, opened 2023-03-21
[V2] feat: persist replica attachment failures

NedAnd1:
**What type of PR is this?**

/kind feature

Replica attachment failures may last longer than the life-span of an event,
so this PR introduces a lightweight event refresher
to ensure that these failures remain visible for as long as they still apply.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/apis/azuredisk/v1beta2/types.go
+++ b/pkg/apis/azuredisk/v1beta2/types.go
@@ -542,8 +542,6 @@ type AzDiskDriverConfiguration struct {
 	DriverName string `json:"driverName,omitempty"`
 	// The address to expose profiling data from the pprof package. If empty, no profiler server is started.
 	ProfilerAddress string `json:"profilerAddress,omitempty"`
-	// The duration for which events are retained by kube-apiserver
-	EventTTLInSec int `json:"eventTTLInSec,omitempty"`
 }
 
 type ControllerConfiguration struct {
@@ -579,6 +577,8 @@ type ControllerConfiguration struct {
 	WaitForLunEnabled bool `json:"waitForLunEnabled,omitempty"`
 	// The maximum number of retries for creating a replica attachment.
 	ReplicaVolumeAttachRetryLimit int `json:"replicaVolumeAttachRetryLimit,omitempty"`
+	// The duration for which events are retained by kube-apiserver
+	EventTTLInSec int `json:"eventTTLInSec,omitempty"`
 }
 
 type NodeConfiguration struct {
--- a/pkg/apis/azuredisk/v1beta2/types.go
+++ b/pkg/apis/azuredisk/v1beta2/types.go
@@ -542,6 +542,8 @@ type AzDiskDriverConfiguration struct {
 	DriverName string `json:"driverName,omitempty"`
 	// The address to expose profiling data from the pprof package. If empty, no profiler server is started.
 	ProfilerAddress string `json:"profilerAddress,omitempty"`
+	// The duration for which events are retained by kube-apiserver
+	EventTTLInSec int `json:"eventTTLInSec,omitempty"`
 }
 
 type ControllerConfiguration struct {
--- a/pkg/controller/common.go
+++ b/pkg/controller/common.go
@@ -1188,3 +1188,53 @@ func WatchObject(mgr manager.Manager, objKind source.Kind) error {
 	})
 	return err
 }
+
+type eventRefresherList struct {
+	curr *eventRefresherNode
+}
+
+type eventRefresherNode struct {
+	prev    *eventRefresherNode
+	next    *eventRefresherNode
+	message string
+	delay   time.Duration
+	objects []runtime.Object
+}
+
+func (events eventRefresherList) isEmpty() bool {
+	return events.curr == nil
+}
+
+func (events *eventRefresherList) clear() {
+	events.curr = nil
+}
+
+func (events *eventRefresherList) add(newEvent *eventRefresherNode) {
+	if events.curr == nil {
+		events.curr = newEvent
+	} else {
+		events.curr.prev.next = newEvent
+		events.curr.delay -= newEvent.delay
+	}
+	newEvent.next = events.curr
+	events.curr.prev = newEvent
+}
+
+// Moves to the next event in the list and retrieves it.
+func (events *eventRefresherList) next() *eventRefresherNode {
+	nextEvent := events.curr.next
+	events.curr = nextEvent
+	return nextEvent
+}
+
+func (oldEvent *eventRefresherNode) remove() {
+	oldEvent.prev.next = oldEvent.next
+	oldEvent.next.prev = oldEvent.prev
+	oldEvent.next.delay += oldEvent.delay
+}
+
+func (oldEvent *eventRefresherNode) tryRemove() {
+	if oldEvent == nil {
+		oldEvent.remove()
+	}
+}
--- a/pkg/controller/common.go
+++ b/pkg/controller/common.go
@@ -1189,51 +1189,48 @@ func WatchObject(mgr manager.Manager, objKind source.Kind) error {
 	return err
 }
 
-// circular linked list with simple accessibility to the current event, etc.
-type eventRefresherList struct {
-	curr *eventRefresherNode
+type circularLinkedList[T any] struct {
+	*circularLinkedListNode[T]
 }
 
-type eventRefresherNode struct {
-	prev      *eventRefresherNode
-	next      *eventRefresherNode
-	message   string
-	timestamp time.Time
-	objects   []runtime.Object
+func (list circularLinkedList[T]) isEmpty() bool {
+	return list.circularLinkedListNode == nil
 }
 
-func (events eventRefresherList) isEmpty() bool {
-	return events.curr == nil
+func (list *circularLinkedList[T]) clear() {
+	list.circularLinkedListNode = nil
 }
 
-func (events *eventRefresherList) clear() {
-	events.curr = nil
-}
-
-func (events *eventRefresherList) add(newEvent *eventRefresherNode) {
-	if events.curr == nil {
-		events.curr = newEvent
+func (list *circularLinkedList[T]) add(newNode *circularLinkedListNode[T]) {
+	if list.circularLinkedListNode == nil {
+		list.circularLinkedListNode = newNode
 	} else {
-		events.curr.prev.next = newEvent
+		list.prev.next = newNode
 	}
-	newEvent.next = events.curr
-	events.curr.prev = newEvent
+	newNode.next = list.circularLinkedListNode
+	list.prev = newNode
 }
 
-// Moves to the next event in the list and retrieves it.
-func (events *eventRefresherList) next() *eventRefresherNode {
-	nextEvent := events.curr.next
-	events.curr = nextEvent
-	return nextEvent
+// Moves to the next node in the list and retrieves it.
+func (list *circularLinkedList[T]) next() *circularLinkedListNode[T] {
+	nextNode := list.circularLinkedListNode.next
+	list.circularLinkedListNode = nextNode
+	return nextNode
 }
 
-func (oldEvent *eventRefresherNode) remove() {
-	oldEvent.prev.next = oldEvent.next
-	oldEvent.next.prev = oldEvent.prev
+type circularLinkedListNode[T any] struct {
+	curr T
+	prev *circularLinkedListNode[T]
+	next *circularLinkedListNode[T]
 }
 
-func (oldEvent *eventRefresherNode) tryRemove() {
-	if oldEvent != nil {
-		oldEvent.remove()
+func (oldNode *circularLinkedListNode[T]) remove() { //nolint:golint // receiver name should be consistent with previous receiver name for invalid-type
+	oldNode.prev.next = oldNode.next
+	oldNode.next.prev = oldNode.prev
+}
+
+func (oldNode *circularLinkedListNode[T]) tryRemove() { //nolint:golint // receiver name should be consistent with previous receiver name for invalid-type
+	if oldNode != nil {
+		oldNode.remove()
 	}
 }
--- a/pkg/controller/common.go
+++ b/pkg/controller/common.go
@@ -1189,16 +1189,17 @@ func WatchObject(mgr manager.Manager, objKind source.Kind) error {
 	return err
 }
 
+// circular linked list with simple accessibility to the current event, etc.
 type eventRefresherList struct {
 	curr *eventRefresherNode
 }
 
 type eventRefresherNode struct {
-	prev    *eventRefresherNode
-	next    *eventRefresherNode
-	message string
-	delay   time.Duration
-	objects []runtime.Object
+	prev      *eventRefresherNode
+	next      *eventRefresherNode
+	message   string
+	timestamp time.Time
+	objects   []runtime.Object
 }
 
 func (events eventRefresherList) isEmpty() bool {
@@ -1214,7 +1215,6 @@ func (events *eventRefresherList) add(newEvent *eventRefresherNode) {
 		events.curr = newEvent
 	} else {
 		events.curr.prev.next = newEvent
-		events.curr.delay -= newEvent.delay
 	}
 	newEvent.next = events.curr
 	events.curr.prev = newEvent
@@ -1230,11 +1230,10 @@ func (events *eventRefresherList) next() *eventRefresherNode {
 func (oldEvent *eventRefresherNode) remove() {
 	oldEvent.prev.next = oldEvent.next
 	oldEvent.next.prev = oldEvent.prev
-	oldEvent.next.delay += oldEvent.delay
 }
 
 func (oldEvent *eventRefresherNode) tryRemove() {
-	if oldEvent == nil {
+	if oldEvent != nil {
 		oldEvent.remove()
 	}
 }
--- a/pkg/controller/common_test.go
+++ b/pkg/controller/common_test.go
@@ -55,7 +55,6 @@ const (
 	verifyCRIInterval          = time.Duration(1) * time.Second
 	testAttachableVolumesValue = "8"
 	testEventTTLInSec          = 10
-	defaultTestVerbosity       = 5
 )
 
 var (
--- a/pkg/controller/common_test.go
+++ b/pkg/controller/common_test.go
@@ -18,10 +18,8 @@ package controller
 
 import (
 	"context"
-	"flag"
 	"fmt"
 	"reflect"
-	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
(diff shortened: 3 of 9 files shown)


pkg/controller/shared_state.go
```diff
@@ -1345,13 +1371,171 @@ func (c *SharedState) createReplicas(ctx context.Context, remainingReplicas int,
...
+			oldEvent.prev.next = oldEvent.next
+			oldEvent.next.prev = oldEvent.prev
+			oldEvent.next.delay += oldEvent.delay
+		}
+		eventMap[newFailureInfo.volumeName] = newEvent
+
+	NoNewFailureLoop:
+		for {
+			select {
+			case newFailureInfo = <-c.eventsToPersistQueue:
```
sunpa93, line 1463: Wouldn't this risk (although with a very low likelihood) not being able to refresh current event (HEAD of the linked list) at the right time if event gets queued when timer is about to go off? Just wondering because I see that you have added a lot of detail to ensure events get refreshed without drift.

sunpa93:
I see that you have in large used linked list and a variance to serialize events and trigger a timer with computed delay to iterate the events in order to refresh/emit the event logs at the right time.

A few things to consider:
1. How will this implementation work out in a scale-out scenario, where we are likely to get a large number of events at the same time?
2. Consider adding abstraction to make the logic in main loop in `_eventRefresherRoutine` more visible and to make the subordinate functions more maintainable (e.g. abstraction for linked-list and for variance computation).
3. Is there an alternate approach you had in mind and what's the trade-off?

pkg/controller/shared_state.go
```diff
@@ -1345,13 +1371,171 @@ func (c *SharedState) createReplicas(ctx context.Context, remainingReplicas int,
...
+}
+
+func (c *SharedState) _eventRefresherInitFunc() {
+	c.eventsToPersistQueue = make(chan ReplicaAttachmentFailureInfo, maxPersistEventConcurrency)
+	c.eventsToUnpersistQueue = make(chan string, maxUnpersistEventConcurrency)
+	go c._eventRefresherRoutine()
+}
+
+func (c *SharedState) _eventRefresherRoutine() {
+	const eventLifeSpan = time.Hour // ToDo: query k8s config?
```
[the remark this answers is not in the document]
NedAnd1: b/c a customer can configure the event life span in kube-apiserver, I do agree that maybe we should move this to azure constants w/ a disclaimer for more visibility, or maybe this should at least be converted to a driver parameter until we figure out how to retrieve that value from the server if it's feasible, what do you think?

pkg/controller/shared_state.go
NedAnd1: Yeah, if a new event is seen by this loop, it would add that new failure and go back into the no new failure loop; if multiple channels happen to be ready at the same time there should be no loss of events & in the scenario you describe, the alarm would/should stay in its triggered state & a delay due to consuming events would be seen by the alarm handler the same as a delay due to sleep inaccuracy.

pkg/controller/shared_state.go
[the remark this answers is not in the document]
sunpa93: We probably should define a `defaultEventLifespan` in `azure_constants.go` and have it be configurable via parameter.

NedAnd1:
> I see that you have in large used linked list and a variance to serialize events and trigger a timer with computed delay to iterate the events in order to refresh/emit the event logs at the right time.
> 
> A few things to consider:
> 
> 1. How will this implementation work out in a scale-out scenario, where we are likely to get a large number of events at the same time?
> 2. Consider adding abstraction to make the logic in main loop in `_eventRefresherRoutine` more visible and to make the subordinate functions more maintainable (e.g. abstraction for linked-list and for variance computation).
> 3. Is there an alternate approach you had in mind and what's the trade-off?

@sunpa93

1. That's a good point; made changes to batch the refresh of events which are close together.
2. Done.

3:
Like in the scenario of your 1st question, having a goroutine for each individual event seems wasteful,
although the go scheduler is very good at minimizing the number of context switches,
minimizing the burden we put on it is not a bad thing to do.
Moreover, a multi-goroutine approach would still need some mechanism for successful attachments
to potentially un-persist their respective events.
Lock contention would occur when successful attachments try to acquire the requisite map around the same time,
and the burden for the requisite map lookups, deletions & insertions etc
would be placed on the goroutine performing the attachments,
instead of a single event-refresher routine allowing the work to be parallelized in a more efficient way.

The trade-off is the complexity of the routine for refreshing events in a timely manner,
but I do believe it's worth it, on top of its potential to be used for more than just attachment failures.
	
Another approach with the single event-refresher routine is starting it whenever the controller starts,
but in its low-cost, no-op nil check state.
Attachment failures would be slightly less costly,
and there is probably not a meaningful trade-off to doing this,
the same might be able to be said about starting the single event-refresher routine on demand,
but I probably took that approach before creating the no-op nil check state,
and when the potential impact of the event refresher routine was more significant to me.

pkg/controller/shared_state.go
```diff
@@ -1345,13 +1371,160 @@ func (c *SharedState) createReplicas(ctx context.Context, remainingReplicas int,
...
+	c.eventsToPersistQueue <- ReplicaAttachmentFailureInfo{volumeName, message, podCopies, timestamp}
+}
+
+// Only requests an event un-persist if the event refresher is active
+func (c *SharedState) unpersistAttachmentFailure(volumeName string) {
+	trySend(c.eventsToUnpersistQueue, volumeName) // also mitigates unlikely race condition of sending to closed channel
+}
+
+func (c *SharedState) _eventRefresherInitFunc() {
+	c.eventsToPersistQueue = make(chan ReplicaAttachmentFailureInfo, maxPersistEventConcurrency)
```
edreed, line 1400: Since there is only a single goroutine processing event persist/unpersist notifications, the concurrency constants control concurrency of production rather than consumption. This means that if persist/unpersist events are produced at a faster pace than the goroutine can process them the producers will block. This could provide to be an unnecessary serialization point on the producer side. I believe we can choose an arbitrarily high buffer size. IIRC, this is just a max count of queued items before blocking the producer and doesn't result in any real allocation.

pkg/controller/shared_state.go
```diff
@@ -1345,13 +1371,160 @@ func (c *SharedState) createReplicas(ctx context.Context, remainingReplicas int,
...
+	c.eventsToUnpersistQueue = make(chan string, maxUnpersistEventConcurrency)
+	go c._eventRefresherRoutine()
+}
+
+func (c *SharedState) _eventRefresherRoutine() {
+	var eventTTL = time.Duration(c.config.EventTTLInSec) * time.Second
+	eventMap := map[string]*eventRefresherNode{}
+	newFailureInfo := <-c.eventsToPersistQueue
+	klog.V(2).Infof("Starting the event refresher") // should this be logged via workflow instead?
```
edreed, line 1409: Logging the start of the event refresher only after the first persist request comes doesn't really have much value. It would result in a clear flow if we simply entered the select wait loop and processed the persist request there. It's a little confusing as-is because the receiving a new persist request appears to do nothing by breaking out of the inner for loop. It's only if you follow the loop until its end that you realize it loops back to the outer for loop and process the request there.

pkg/controller/shared_state.go
```diff
@@ -1345,13 +1371,160 @@ func (c *SharedState) createReplicas(ctx context.Context, remainingReplicas int,
...
+		}
+		newEventTimeVariance := currTime.Sub(newFailureInfo.timestamp) + maxEventOverlapDuration
+
+		if events.isEmpty() {
+			newEvent.delay = eventTTL
+			timeVariance = newEventTimeVariance
+			alarm = time.NewTimer(eventTTL - timeVariance)
+			lastTime = currTime
+		} else {
+			newEvent.delay = newFailureInfo.timestamp.Sub(lastTime) + (timeVariance - newEventTimeVariance)
```
edreed, line 1429: A priority queue ordered by time of refresh would be easier to reason about and validate correctness. You can build a priority queue in Go using a [`Heap`](https://pkg.go.dev/container/heap@go1.20.5). To accommodate unpersist requests, you could simply maintain a separate set (implemented as map[string]struct{}) of volume names and garbage collect when its refresh timer expires.

pkg/controller/common.go
```diff
@@ -1188,3 +1188,53 @@ func WatchObject(mgr manager.Manager, objKind source.Kind) error {
 	})
 	return err
 }
+
+type eventRefresherList struct {
+	curr *eventRefresherNode
+}
+
+type eventRefresherNode struct {
+	prev    *eventRefresherNode
```
edreed, line 1197: Consider using ['List'](https://pkg.go.dev/container/list@go1.20.5) rather than rolling your own.

pkg/controller/shared_state.go
NedAnd1: Logging the start of the event refresher is just an additive log message to let us know when the event refresher is first started, w/ messages for when it's deactivated, and reactivated, to help us know how its operating, like if we ever need to debug a potential decrease in overall performance b/c of it.

There are currently 3 entry-points to a new failure for the event refresher (when it's started for the first time, when receives a new event while active, and when it receives a new event while inactive), so we can't handle new failures by simply moving the new failure logic into that waiting for a new event case, an alternative could be a NewFailure section w/ goto's... thoughts? or other recommendations?

pkg/apis/azuredisk/v1beta2/types.go
```diff
@@ -542,6 +542,8 @@ type AzDiskDriverConfiguration struct {
 	DriverName string `json:"driverName,omitempty"`
 	// The address to expose profiling data from the pprof package. If empty, no profiler server is started.
 	ProfilerAddress string `json:"profilerAddress,omitempty"`
+	// The duration for which events are retained by kube-apiserver
+	EventTTLInSec int `json:"eventTTLInSec,omitempty"`
```
edreed, line 546: This should be part of the `ControllerConfiguration`.

pkg/controller/common_test.go
```diff
@@ -475,6 +479,21 @@ func initState(client client.Client, azClient azdisk.Interface, kubeClient kuber
...
+	if vflag == nil {
+		overwriteVerbosity := true
+		klog.InitFlags(nil)
+		flag.Visit(func(f *flag.Flag) {
+			if f.Name == "v" {
+				overwriteVerbosity = false // use the explicitly provided verbosity
+			}
+		})
+		if overwriteVerbosity {
+			flag.Set("v", strconv.Itoa(defaultTestVerbosity)) // use `defaultTestVerbosity` instead of the default verbosity of 0
```
edreed, line 495: We shouldn't overwrite the verbosity if not specified.

pkg/controller/shared_state.go
NedAnd1: I did like the ability to just use `events.curr` in the event retriever to get the current event w/ all its properties right there instead of using something more complicated to get the current event w/ two levels of indirection & a lack of type specificity (unless we were to create a generic circular linked list),
as well as the simplicity of removing an event from the list w/ by calling `remove()` or `tryRemove()` on the event itself (from the event map), w/ all its properties / fields readily accessible.

pkg/controller/shared_state.go
edreed: You could probably make the queue length equal to the number of controller [`WorkerThreads`](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/ce8a4cf7054558f2a1ac68a9f47a6ab0f1432a24/pkg/apis/azuredisk/v1beta2/types.go#L575C6-L575C6).

pkg/controller/shared_state.go
edreed: Given that this is a single goroutine processing the list, it isn't very likely to directly impact performance. The most likely impact would be caused by the additional PUT calls to API server leading to client-side throttling. Since event creation is actually asynchronous (it spawns off a goroutine to do the PUT), this impact may not be directly correlated to these debug traces but we can rely on the API client throttling traces for detection. However, the debug traces *do* impact logging because the driver is typically run with verbosity of 3 in production. While the debug traces are useful to debug operation, they shouldn't be enabled in prod. I would suggest using level 6 for them.

Do we really need to have an "inactive" state for the refresher? It's seems to be this is to support unconditional "un-persist" of the event but we do log a success event when the replica is attached. This could be used to trigger the un-persist and enable the refresher to be "active" always. This would also allow you to simplify the loop.

pkg/controller/shared_state.go
edreed: I do like the idea of creating a generic circular linked list implementation. This would more clearly highlight its use here and possibly be applicable to other problems in the future.

pkg/controller/shared_state.go
```diff
@@ -1345,13 +1373,152 @@ func (c *SharedState) createReplicas(ctx context.Context, remainingReplicas int,
...
+		newEvent := &eventRefresherNode{
+			message:   newFailureInfo.message,
+			objects:   newFailureInfo.pods,
+			timestamp: newFailureInfo.timestamp.Add(eventTTL - eventOverlapDuration),
+		}
+
+		if events.isEmpty() {
+			lastTime = time.Now()
+			delay = newEvent.timestamp.Sub(lastTime) - expLatency
+			alarm = time.NewTimer(delay)
```
edreed, line 1428: On first activation `delay` is `duration.Zero`, so the alarm will fire immediately and add a duplicate event?

pkg/controller/common_test.go
NedAnd1: for our unit tests, you don't think the default verbosity of 0 is too low? the default verbosity as it is now seems geared towards production; I noticed other k8s projects using non-static klogging w/ 5 as their default verbosity for testing, but we use a mix of traditional klog calls (reporting mostly non-zero level events) & workflow logging (reporting mostly zero-level events) & useful messages while debugging failed unit tests might be missed

pkg/apis/azuredisk/v1beta2/types.go
NedAnd1: this is really more of kube-apiserver configuration & we allow the customer to tell us what they gave kube-apiserver b/c retrieving that ourselves wouldn't be easy, but I'll move it

pkg/controller/shared_state.go
NedAnd1: `{event}.timestamp` is when the event should next be refreshed, initialized to the original timestamp plus the ttl, etc

pkg/controller/shared_state.go
```diff
@@ -1345,13 +1373,152 @@ func (c *SharedState) createReplicas(ctx context.Context, remainingReplicas int,
...
+}
+
+func trySend[T any](channel chan T, value T) {
+	if channel != nil {
+		trySendNonNil(channel, value)
+	}
+}
+
+func trySendNonNil[T any](nonNilChannel chan T, value T) {
+	defer func() { _ = recover() }()
```
edreed, line 1518: We shouldn't rely on ignoring panics to handle the race between `unmake` and `trySend`.

pkg/controller/shared_state.go
NedAnd1: sending to the event persistence queue / channel still includes a slight window of locking & channel operations etc on both the producer & consumer side which we could avoid altogether during attaches when we know there are no events currently being persisted.

we always attempt un-persisting an attachment failure, b/c errors like the volume no longer existing might make that attachment failure no longer relevant & if we retry attaching a volume and it fails again IMHO we should stop refreshing the old failure events for that volume & start refreshing the new ones

pkg/controller/shared_state.go
NedAnd1: please elaborate on the potential pitfalls of doing this for this scenario, as sending to a non-nil channel only panics if the channel is closed; the fact that go panics when attempting to send to a closed channel is worked around in this way in [docker & k8s](https://github.com/search?q=%28org%3Akubernetes+OR+org%3Adocker+OR+org%3Amoby++%29+%2F%5Csrecover%5C%28%5B%5Cs%5CS%5Cr%5Cn%5D*%5C%7D%5C%28%5C%29%5Cn*%5Cs*%5Cw%2B+%3C-+%5Cw%2B%2F+-path%3A**test**&type=code)

pkg/controller/common_test.go
edreed: The unit tests already produce a lot of output (when using the `go test -v` option) which makes it difficult to find the failing test case. Providing an option to increase verbosity to help debug a specific failing case when one manually would be preferable to emitting additional logs by default.

pkg/controller/shared_state.go
edreed: Panics should be reserved for genuinely unexpected scenarios and code should be structured to avoid them. If we expect them by design, that can mask cases were the channel was closed unexpectedly leading to undefined behavior.

NedAnd1: Sending to a closed channel is not something that is _expected_ to happen, but there _are_ race conditions in which the channel might be unexpectedly closed (really, really small chance).
The moby / docker project mitigates this in the same way: https://github.com/moby/moby/issues/37735
`trySend` is also only used for un-persisting attachment failures, and if such a channel was unintentionally closed due to a programming mistake, the consequences would be very insignificant...
Perhaps we could move `trySend` and `trySendNonNil` inside the un-persist function to restrict its use?
And/or log a warning when an error is encountered?

pkg/controller/common_test.go
NedAnd1: We could only log the output of tests which succeed by removing the go test `-v` option: https://pkg.go.dev/cmd/go#hdr-Testing_flags; even with the go test `-v` option, the default klog verbosity is 0, but there are non-zero messages which would be useful for our workflows & makes...
Nevertheless, I'll remove this logic, if we still want to keep everything as is, but I'd definitely agree that our unit tests produce a _lot_ of output esp. due to our workflow logging.

pkg/controller/shared_state.go
edreed: Why close the channel at all? It doesn't take a significant amount of resources and the process itself is long-lived (i.e. it will only stop if it panics or the pod is explicitly terminated e.g. for node shutdown).

As I mentioned in [another comment](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1781#discussion_r1258925137), I believe we can simplify this loop by eliminating the "inactive" state. This would essentially require having the channel open always. Go channels are designed as the primary concurrency mechanism support multiple producer/single (or multiple) consumer scenarios. Though not "lock-free", using channels is designed to be fast and scalable, so optimizing to avoid sending on a channel isn't really necessary.

pkg/controller/common_test.go
edreed: As we discussed, I am fine with adding an option to enable additional logging when the unit tests are run manually, but it should not emit them by default.

pkg/controller/shared_state.go
```diff
@@ -86,8 +88,30 @@ type SharedState struct {
...
+	message    string
+	pods       []runtime.Object
+	timestamp  time.Time
+}
+
+const (
+	// persistent event and their refreshes should overlap by this amount of time
+	eventOverlapDuration = time.Millisecond
+	// allow persistent events to overlap by up to this additional amount of time for batched reporting etc.
+	eventOverlapVariance = time.Millisecond
```
edreed, line 107: I believe 1ms would be too small for actually end up batching. Since batching would result in more calls to API server in a short window of time thereby increasing the likelihood of throttling, should we even bother?

pkg/controller/shared_state.go
```diff
@@ -86,8 +88,30 @@ type SharedState struct {
...
+type ReplicaAttachmentFailureInfo struct {
+	volumeName string
+	message    string
+	pods       []runtime.Object
+	timestamp  time.Time
+}
+
+const (
+	// persistent event and their refreshes should overlap by this amount of time
+	eventOverlapDuration = time.Millisecond
```
edreed, line 105: We should probably use a larger overlap time. During periods of heavy server activity, throttling could result in delays over 1s. 30s would probably be a reasonable value to ensure the event is always visible, but minimize the period in which a duplicate may appear.

edreed:
I'm not sure how this was closed. I don't recall doing that even though I'm the one recorded as doing so.

edreed:
/approve

@NedAnd1 Need to fix the linter issues, but otherwise LGTM!

PR #1786, opened 2023-03-22
[V2] refactor: rename functions and use atomic pkg feature

hccheng72:
**What type of PR is this?**
/kind cleanup

An infinite loop was used to secure atomic operation for decrement of node capacity in `decrementNodeCapacity` function of shared_state.go, but we can just leverage built-in feature from atomic package. Also, renaming functions to avoid confusion.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/controller/shared_state.go
```diff
@@ -1788,35 +1788,28 @@ func (c *SharedState) deleteNodeFromAvailableAttachmentsMap(ctx context.Context,
...
-					klog.Errorf("Failed to decrement attachment count for node(%s), because no available attachment", node)
-					return false
-				}
-				return true
-			}
+		if currentCapacity := remainingCapacity.(*atomic.Int32).Load(); currentCapacity == int32(0) {
+			klog.Errorf("Failed to decrement disk capacity for node(%s) because no remaining capacity", node)
+			return false
 		}
+		remainingCapacity.(*atomic.Int32).Add(-1)
```
alice-zheyan-yu, line 1798: The value of `remainingCapacity` is loaded in line 1792 and gets updated in this line 1798. However, if another thread modifies the value of `remainingCapacity` in the meantime, the number will be inaccurate.

pkg/controller/shared_state.go
```diff
@@ -1788,35 +1788,28 @@ func (c *SharedState) deleteNodeFromAvailableAttachmentsMap(ctx context.Context,
...
-		for {
-			currentCapacity = remainingCapacity.(*atomic.Int32).Load()
-			if currentCapacity == int32(0) || remainingCapacity.(*atomic.Int32).CompareAndSwap(currentCapacity, currentCapacity-1) {
-				if currentCapacity == int32(0) {
-					klog.Errorf("Failed to decrement attachment count for node(%s), because no available attachment", node)
-					return false
-				}
-				return true
-			}
+		if currentCapacity := remainingCapacity.(*atomic.Int32).Load(); currentCapacity == int32(0) {
```
edreed, line 1794: The previous pattern using `CompareAndSwap` was intentional. Without doing so, multiple go routines can overdecrement the remaining capacity resulting in a negative value. Consider the following execution sequence:

|goroutine|action|capacity|
|-|-|-|
| A | `currentCapacity := remainingCapacity.(*atomic.Int32).Load()` | 1 |
| A | `currentCapacity == int32(0)` is `false` | 1 |
| B | `currentCapacity := remainingCapacity.(*atomic.Int32).Load()` | 1 |
| B | `currentCapacity == int32(0)` is `false` | 1 |
| B | `remainingCapacity.(*atomic.Int32).Add(-1)` | 0 |
| A | `remainingCapacity.(*atomic.Int32).Add(-1)` | -1 |

hccheng72: You're right, so the key part is to make the check and update in an atomic operation. Can we make it simpler like this?
```
func (c *SharedState) decrementNodeCapacity(ctx context.Context, node string) bool {
	remainingCapacity, nodeExists := c.availableAttachmentsMap.Load(node)
	if nodeExists && remainingCapacity != nil {
		for {
			if remainingCapacity.(*atomic.Int32).Load() == int32(0) {
				klog.Errorf("Failed to decrement disk capacity for node(%s) because no remaining capacity", node)
				return false
			}
			remainingCapacity.(*atomic.Int32).Add(-1)
			return true
		}	
	}

	klog.Errorf("Failed to decrement disk capacity because node(%s) not found", node)
	return false
}
```

hccheng72: Yeah, You're right. Thanks for the detailed explanation. I change it back and only have a minor update.

PR #1787, opened 2023-03-23
[V2] fix: recover nodes with correct remaining capacity

### LINKED ISSUE #1780: [V2] incorrect remaining capacity of a node after recovery
**What happened**:
When runnning the scale test with replicas, some replica azVolumeAttachments weren't created due to no capacity from any selected nodes. 
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/944b31cccffa459b5a6707783a14f26ddc40850a/pkg/controller/shared_state.go#L1398 
However, when I checked the number of disks attached to those nodes, there're remaining capacities.

I'm guessing it's a bug happening in node recovery. 
We call [`addNodeToAvailableAttachmentsMap`](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/944b31cccffa459b5a6707783a14f26ddc40850a/pkg/controller/shared_state.go#L1766) when a node is recovered. In this function, we call [`GetAzVolumeAttachmentsForNode`](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/f743329bd8249a3bcd1b1685a09820b1b8d5cf0a/pkg/azureutils/azure_disk_utils.go#L1074) to list all azVolumeAttachments CRIs (no matter what their the states are) to calculate how many disks have been attached to a node, and save the reaming capacity to `availableAttachmentsMap`. However, we only call [`decrementAttachmentCount`](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/944b31cccffa459b5a6707783a14f26ddc40850a/pkg/controller/shared_state.go#L1791) when the disk attachment is succeeded. This will cause the driver to save an incorrect remaining capacity of a node after recovery and stop the driver to create replica azVolumeAttachments.

**How to reproduce it**:

**Environment**:

- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hccheng72:
**What type of PR is this?**
/kind bug

In addition to fix the issue mentioned below, this PR also add a new unit test and updates an error from the existing one:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/7c05b95d1f7c5d7927d4e2a4e12b4264152ba216/pkg/controller/shared_state_test.go#L1273 the test case is "when node is in cache", but it returns "NotFound" error for `GET` method.

Fixes #1780

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azureutils/azure_disk_utils.go
+++ b/pkg/azureutils/azure_disk_utils.go
@@ -240,15 +240,36 @@ func GetNodeMaxDiskCount(ctx context.Context, cachedReader client.Reader, nodeNa
 	return GetNodeMaxDiskCountWithLabels(nodeObj.Labels)
 }
 
-func GetNodeRemainingDiskCount(ctx context.Context, cachedReader client.Reader, nodeName string) (int, error) {
-	nodeObj := &v1.Node{}
-	if err := cachedReader.Get(ctx, types.NamespacedName{Name: nodeName}, nodeObj); err != nil {
+// GetNodeRemainingDiskCountApprox returns an approximate capacity to the node
+func GetNodeRemainingDiskCountApprox(ctx context.Context, cachedReader client.Reader, nodeName string) (int, error) {
+	attachments, err := GetAzVolumeAttachmentsForNode(ctx, cachedReader, nodeName, AllRoles)
+	if err != nil {
 		return -1, err
 	}
 
-	// get all replica azvolumeattachments on the node
+	return calculateNodeCapacity(ctx, cachedReader, nodeName, len(attachments))
+}
+
+// GetNodeRemainingDiskCountActual returns the actual capacity to the node
+func GetNodeRemainingDiskCountActual(ctx context.Context, cachedReader client.Reader, nodeName string) (int, error) {
 	attachments, err := GetAzVolumeAttachmentsForNode(ctx, cachedReader, nodeName, AllRoles)
 	if err != nil {
+		return -1, err
+	}
+
+	numberOfAttachedAtts := 0
+	for _, attachment := range attachments {
+		if attachment.Status.State == azdiskv1beta2.Attached {
+			numberOfAttachedAtts++
+		}
+	}
+
+	return calculateNodeCapacity(ctx, cachedReader, nodeName, numberOfAttachedAtts)
+}
+
+func calculateNodeCapacity(ctx context.Context, cachedReader client.Reader, nodeName string, numOfAttachments int) (int, error) {
+	nodeObj := &v1.Node{}
+	if err := cachedReader.Get(ctx, types.NamespacedName{Name: nodeName}, nodeObj); err != nil {
 		return -1, err
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/azureutils/azure_disk_utils.go
```diff
@@ -240,18 +240,39 @@ func GetNodeMaxDiskCount(ctx context.Context, cachedReader client.Reader, nodeNa
...
+// GetNodeRemainingDiskCountActual returns the actual capacity to the node
+func GetNodeRemainingDiskCountActual(ctx context.Context, cachedReader client.Reader, nodeName string) (int, error) {
 	attachments, err := GetAzVolumeAttachmentsForNode(ctx, cachedReader, nodeName, AllRoles)
 	if err != nil {
 		return -1, err
 	}
 
+	numberOfAttachedAtts := 0
+	for _, attachment := range attachments {
+		if attachment.Status.State == azdiskv1beta2.Attached {
```
edreed, line 262: We should include disks in the `Attaching` state in the disk count.

hccheng72: Currently, we only decrement the capacity when attach succeeded, so the `availableAttachmentsMap` stores actual attachments.
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/f743329bd8249a3bcd1b1685a09820b1b8d5cf0a/pkg/controller/attach_detach.go#L298
 I think including disks in `Attaching` state doesn't align to this logic.

hccheng72:
Should we increment the capacity once detach succeeded rather than in deletion event? Some CRIs may get deleted without detach. https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/f743329bd8249a3bcd1b1685a09820b1b8d5cf0a/pkg/controller/attach_detach.go#L123

pkg/azureutils/azure_disk_utils.go
landreasyan: Success on attach will decrement the count, so we don't need to decrement for pending states here otherwise we double count. https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/944b31cccffa459b5a6707783a14f26ddc40850a/pkg/controller/attach_detach.go#L298

alice-zheyan-yu:
> Should we increment the capacity once detach succeeded rather than in deletion event? Some CRIs may get deleted without detach. @alice-zheyan-yu @sunpa93 could you help check on this? Thank you.
> 
> https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/f743329bd8249a3bcd1b1685a09820b1b8d5cf0a/pkg/controller/attach_detach.go#L123

In my PR, the `Recover` process of `AzVolumeAttachment` will delete the `AzVolumeAttachment` CRI if the check of `isAttached` returns false. Based on it, I'd agree that it seems more accurate if we could increment the capacity once the detachment is succeeded rather than in delete event.
The PR link: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/1666

sunpa93:
@alice-zheyan-yu, @hccheng72
If `nodeAvailabilityMap` were to track the count of attached `AzVolumeAttachment` for a node, I believe we should decrement and incremented after a successful attach and detach event respectively. And if it were to track the total count of `AzVolumeAttachment` instances for a node, I believe we should decrement and increment based on the CR creation and deletion event respectively.

sunpa93 (approved):
/lgtm

edreed:
The test failures are unrelated to this PR.

PR #1790, opened 2023-03-25
fix: PremiumV2_LRS caching mode issue

### LINKED ISSUE #1770: First PremiumV2_LRS disk attached to AKS node has read-only host caching enabled if not specified in storageclass
**What happened**:
Unless the custom storageclass used by a persistent volume specifies the parameter `cachingMode: None` AKS will attach the first Premium SSD v2 (PremiumV2_LRS) disk to an AKS node with read-only host caching, despite host caching not being supported for Premium SSD v2 disks. Attempts to attach any additional persistent volumes using the `PremiumV2_LRS` to the AKS node will fail, even if the `cachingMode` is specified as `None ` on the storageclass.

Since PremiumV2_LRS disks do not support host-caching, AKS should not allow the first persistent volume disk to attach to an AKS node with host-caching enabled. Further, AKS documentation should make it obvious that it is necessary to include the `cachingMode` parameter in the storageclass spec.

**How to reproduce it**:
1) Create storageclass A using the `PremiumV2_LRS`. Do not include the `cachingMode` in the spec.
2) Create a statefulset with PVC using storageclass A and observe the successful creation. Note in the Azure GUI that you can see the host-caching set to read-only under the specific instance's disk settings.
3) Create storageclass B using the `PremiumV2_LRS` and set `cachingMode: None` in the spec.
4) Create a statefulset with PVC using storageclass B, you will receive the following error:
`message: "AttachVolume.Attach failed for volume \"pvc-redacted\" : rpc error: code = Internal desc = Attach volume /subscriptions/redacted/resourceGroups/mc_redacted/providers/Microsoft.Compute/disks/pvc-redacted to instance aks-redacted-vmss000001 failed with Retriable: false, RetryAfter: 0s, HTTPStatusCode: 400, RawError: {\r\n  \"error\": {\r\n    \"code\": \"InvalidParameter\",\r\n    \"message\": \"Storage Account type PremiumV2_LRS is not supported with caching type ReadOnly. Please use caching type None.\",\r\n    \"target\": \"dataDisk.caching\"\r\n  }\r\n}"`

This is referenced in this issue: https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/1715

That ticket was closed without really addressing the actual issue, AKS should not be attaching disks in an unsupported configuration (see steps 1 and 2 in the steps to reproduce the issue). 

**Environment**:

- CSI Driver version: 1.26.2
- Kubernetes version (use `kubectl version`): 1.25.5
- OS (e.g. from /etc/os-release): AKSUbuntu-2204gen2containerd-2023.01.26

andyzhangx:
**What type of PR is this?**
/kind bug

fix: PremiumV2_LRS caching mode issue

https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types#differences-between-premium-ssd-and-premium-ssd-v2

Unlike Premium SSDs, Premium SSD v2 doesn't have dedicated sizes. You can set a Premium SSD v2 to any supported size you prefer, and make granular adjustments to the performance without downtime. Premium SSD v2 doesn't support host caching but, benefits significantly from lower latency which addresses some of the same core problems host caching addresses. The ability to adjust IOPS, throughput, and size at any time also means you can avoid the maintenance overhead of having to stripe disks to meet your needs.

Fixes #1770

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: PremiumV2_LRS caching mode issue
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -29,6 +29,7 @@ import (
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -36,11 +37,11 @@ import (
 	volerr "k8s.io/cloud-provider/volume/errors"
 	"k8s.io/klog/v2"
 
-	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 	consts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/optimization"
 	volumehelper "sigs.k8s.io/azuredisk-csi-driver/pkg/util"
+	azureconstants "sigs.k8s.io/cloud-provider-azure/pkg/consts"
 	"sigs.k8s.io/cloud-provider-azure/pkg/metrics"
 	azure "sigs.k8s.io/cloud-provider-azure/pkg/provider"
 )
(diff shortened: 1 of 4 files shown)


pkg/azuredisk/controllerserver.go
```diff
@@ -138,6 +139,10 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	if _, err := azureutils.NormalizeCachingMode(diskParams.CachingMode); err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	if skuName == azureconstants.PremiumV2LRS {
```
edreed, line 142: This validation should be done in `azureutils.ParseDiskParameters`.

ZeroMagic (approved):
/lgtm

PR #1791, opened 2023-03-27
feat: add copy snapshot in cross region directly

### LINKED ISSUE #1694: Store snapshot at different Azure regions

This is new feature request for backup and restore snapshots at different Azure region.
For now it is possible to create snapshots at different Azure resource groups but snapshot location is always the same as AKS location (disk location).

For Disaster Recovery in paired or different Azure region, Azure Snapshot can be copied between Azure regions via [procedure]( https://learn.microsoft.com/en-us/azure/virtual-machines/disks-copy-incremental-snapshot-across-regions) and it is not possible to use copied snapshots to restore persistent volumes in AKS using CSI driver.

1. CR VolumeSnapshotLocation in case of different Azure resource group to respect Azure resource group default location.
2. During restore, CR VolumeSnapshotLocation should point to Azure resource group with copied snapshots and create pv based on snapshot and do not refer to initial disk location. 

Currently it is error that disk cannot be mounted as pv refers to disk which is already mouted:

```
Warning  FailedAttachVolume  106s (x12 over 10m)  attachdetach-controller  AttachVolume.Attach failed for volume "pvc-f4579593-9b42-4188-b5a3-bb8e24ce027b" : rpc error: code = Internal desc = Attach volume /subscriptions/subID/resourceGroups/backupedClusterResourceGroup/providers/Microsoft.Compute/disks/pvc-f4579593-9b42-4188-b5a3-bb8e24ce027b to instance aks-default-vmss000001 failed with disk(/subscriptions/subID/resourceGroupsbackupedClusterResourceGroup/providers/Microsoft.Compute/disks/pvc-f4579593-9b42-4188-b5a3-bb8e24ce027b) already attached to node(/subscriptions/subID/resourceGroups/backupedClusterResourceGroup/providers/Microsoft.Compute/virtualMachineScaleSets/aks-default-backupId-vmss/virtualMachines/aks-default-18145209-vmss_51), could not be attached to node(aks-default-newId-vmsNodeAtNewCluster)
```

I try to use Velero with and without EnableCSI plugin and no success. 
It may be possible to use storage service with geo-redundancy and do not use Azure snapshots for Disaster Recovery on different region.

umagnus:
**What type of PR is this?**

/kind feature

feat: add copy snapshot in cross region directly

Fixes #1694

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

test log
```
I0327 11:35:05.765036       1 controllerserver.go:893] begin to create snapshot(local_snapshot-4d074b39-8a9b-493a-a101-dc110548227b, incremental: true) under rg(mc_aks-yxytest_group_aks-yxytest_eastus) region(eastus)
I0327 11:35:08.224270       1 controllerserver.go:902] create snapshot(local_snapshot-4d074b39-8a9b-493a-a101-dc110548227b) under rg(mc_aks-yxytest_group_aks-yxytest_eastus) region(eastus) successfully
I0327 11:35:08.273543       1 util.go:124] Send.sendRequest got response with ContentLength -1, StatusCode 200 and responseBody length 1295
I0327 11:35:08.273723       1 controllerserver.go:909] create snapshot location(westus) is different from disk location(eastus)
I0327 11:35:08.273745       1 controllerserver.go:911] begin to copy an incremental snapshot from region(eastus) to new region(westus)
I0327 11:35:08.273755       1 controllerserver.go:917] begin to create snapshot(snapshot-4d074b39-8a9b-493a-a101-dc110548227b, incremental: true) under rg(mc_aks-yxytest_group_aks-yxytest_eastus) region(westus)
I0327 11:35:12.444469       1 controllerserver.go:926] create snapshot(snapshot-4d074b39-8a9b-493a-a101-dc110548227b) under rg(mc_aks-yxytest_group_aks-yxytest_eastus) region(westus) successfully
I0327 11:35:12.444498       1 controllerserver.go:928] begin to delete snapshot(local_snapshot-4d074b39-8a9b-493a-a101-dc110548227b) under rg(mc_aks-yxytest_group_aks-yxytest_eastus) region(westus)
I0327 11:35:17.695038       1 controllerserver.go:933] delete snapshot(local_snapshot-4d074b39-8a9b-493a-a101-dc110548227b) under rg(mc_aks-yxytest_group_aks-yxytest_eastus) region(westus) successfully
I0327 11:35:17.791222       1 util.go:124] Send.sendRequest got response with ContentLength -1, StatusCode 200 and responseBody length 1334
I0327 11:35:17.791401       1 azure_metrics.go:115] "Observed Request Latency" latency_seconds=12.026353477 request="azuredisk_csi_driver_controller_create_snapshot" resource_group="mc_aks-yxytest_group_aks-yxytest_eastus" subscription_id="8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8" source="disk.csi.azure.com" source_resource_id="/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/mc_aks-yxytest_group_aks-yxytest_eastus/providers/Microsoft.Compute/disks/pvc-8deda4a0-f6c8-4e82-9f9d-10e7f40bcc2c" snapshot_name="snapshot-4d074b39-8a9b-493a-a101-dc110548227b" result_code="succeeded"
I0327 11:35:17.791421       1 utils.go:84] GRPC response: {"snapshot":{"creation_time":{"nanos":47815400,"seconds":1679916910},"ready_to_use":true,"size_bytes":10737418240,"snapshot_id":"/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/mc_aks-yxytest_group_aks-yxytest_eastus/providers/Microsoft.Compute/snapshots/snapshot-4d074b39-8a9b-493a-a101-dc110548227b","source_volume_id":"/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/mc_aks-yxytest_group_aks-yxytest_eastus/providers/Microsoft.Compute/disks/pvc-8deda4a0-f6c8-4e82-9f9d-10e7f40bcc2c"}}
```
image

**Release note**:
```
none
```

### DIFF
--- a/deploy/csi-azuredisk-controller.yaml
+++ b/deploy/csi-azuredisk-controller.yaml
@@ -86,6 +86,7 @@ spec:
             - "-leader-election"
             - "--leader-election-namespace=kube-system"
             - "--v=2"
+            - "--timeout=600s"
           env:
             - name: ADDRESS
               value: /csi/csi.sock
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2022-03-01/compute"
 	"github.com/container-storage-interface/spec/lib/go/csi"
@@ -43,6 +44,11 @@ import (
 	volumehelper "sigs.k8s.io/azuredisk-csi-driver/pkg/util"
 	"sigs.k8s.io/cloud-provider-azure/pkg/metrics"
 	azure "sigs.k8s.io/cloud-provider-azure/pkg/provider"
+)
+
+const (
+	waitForSnapshotCopyInterval = 5 * time.Second
+	waitForSnapshotCopyTimeout  = 10 * time.Minute
 )
 
 // listVolumeStatus explains the return status of `listVolumesByResourceGroup`
@@ -868,7 +874,7 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 			},
 			Incremental: &incremental,
 		},
-		Location: &location,
+		Location: &d.cloud.Location,
 		Tags:     tags,
 	}
 
@@ -879,13 +885,28 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 		snapshot.SnapshotProperties.DataAccessAuthMode = compute.DataAccessAuthMode(dataAccessAuthMode)
 	}
 
+	if acquired := d.volumeLocks.TryAcquire(snapshotName); !acquired {
+		return nil, status.Errorf(codes.Aborted, volumeOperationAlreadyExistsFmt, snapshotName)
+	}
+	defer d.volumeLocks.Release(snapshotName)
+
+	var crossRegionSnapshotName string
+	if location != "" && location != d.cloud.Location {
+		if incremental {
+			crossRegionSnapshotName = snapshotName
+			snapshotName = azureutils.CreateValidDiskName("local_" + snapshotName)
+		} else {
+			return nil, status.Errorf(codes.InvalidArgument, "could not create snapshot cross region with incremental is false")
+		}
+	}
+
 	mc := metrics.NewMetricContext(consts.AzureDiskCSIDriverName, "controller_create_snapshot", d.cloud.ResourceGroup, d.cloud.SubscriptionID, d.Name)
 	isOperationSucceeded := false
 	defer func() {
 		mc.ObserveOperationWithResult(isOperationSucceeded, consts.SourceResourceID, sourceVolumeID, consts.SnapshotName, snapshotName)
 	}()
 
-	klog.V(2).Infof("begin to create snapshot(%s, incremental: %v) under rg(%s)", snapshotName, incremental, resourceGroup)
+	klog.V(2).Infof("begin to create snapshot(%s, incremental: %v) under rg(%s) region(%s)", snapshotName, incremental, resourceGroup, d.cloud.Location)
 	if rerr := d.cloud.SnapshotsClient.CreateOrUpdate(ctx, subsID, resourceGroup, snapshotName, snapshot); rerr != nil {
 		if strings.Contains(rerr.Error().Error(), "existing disk") {
 			return nil, status.Error(codes.AlreadyExists, fmt.Sprintf("request snapshot(%s) under rg(%s) already exists, but the SourceVolumeId is different, error details: %v", snapshotName, resourceGroup, rerr.Error()))
@@ -894,11 +915,52 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 		azureutils.SleepIfThrottled(rerr.Error(), azureconstants.SnapshotOpThrottlingSleepSec)
 		return nil, status.Error(codes.Internal, fmt.Sprintf("create snapshot error: %v", rerr.Error()))
 	}
-	klog.V(2).Infof("create snapshot(%s) under rg(%s) successfully", snapshotName, resourceGroup)
+	klog.V(2).Infof("create snapshot(%s) under rg(%s) region(%s) successfully", snapshotName, resourceGroup, d.cloud.Location)
 
 	csiSnapshot, err := d.getSnapshotByID(ctx, subsID, resourceGroup, snapshotName, sourceVolumeID)
 	if err != nil {
 		return nil, err
+	}
+
+	if crossRegionSnapshotName != "" {
+		copySnapshot := snapshot
+		if copySnapshot.SnapshotProperties == nil {
+			copySnapshot.SnapshotProperties = &compute.SnapshotProperties{}
+		}
+		if copySnapshot.SnapshotProperties.CreationData == nil {
+			copySnapshot.SnapshotProperties.CreationData = &compute.CreationData{}
+		}
+		copySnapshot.SnapshotProperties.CreationData.SourceResourceID = &csiSnapshot.SnapshotId
+		copySnapshot.SnapshotProperties.CreationData.CreateOption = compute.CopyStart
+		copySnapshot.Location = &location
+
+		klog.V(2).Infof("begin to create snapshot(%s, incremental: %v) under rg(%s) region(%s)", crossRegionSnapshotName, incremental, resourceGroup, location)
+		if rerr := d.cloud.SnapshotsClient.CreateOrUpdate(ctx, subsID, resourceGroup, crossRegionSnapshotName, copySnapshot); rerr != nil {
+			if strings.Contains(rerr.Error().Error(), "existing disk") {
+				return nil, status.Error(codes.AlreadyExists, fmt.Sprintf("request snapshot(%s) under rg(%s) already exists, but the SourceVolumeId is different, error details: %v", crossRegionSnapshotName, resourceGroup, rerr.Error()))
+			}
+
+			azureutils.SleepIfThrottled(rerr.Error(), consts.SnapshotOpThrottlingSleepSec)
+			return nil, status.Error(codes.Internal, fmt.Sprintf("create snapshot error: %v", rerr.Error()))
+		}
+		klog.V(2).Infof("create snapshot(%s) under rg(%s) region(%s) successfully", crossRegionSnapshotName, resourceGroup, location)
+
+		if err := d.waitForSnapshotCopy(ctx, subsID, resourceGroup, crossRegionSnapshotName, waitForSnapshotCopyInterval, waitForSnapshotCopyTimeout); err != nil {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("failed to wait for snapshot copy cross region: %v", err.Error()))
+		}
+
+		klog.V(2).Infof("begin to delete snapshot(%s) under rg(%s) region(%s)", snapshotName, resourceGroup, d.cloud.Location)
+		if rerr := d.cloud.SnapshotsClient.Delete(ctx, subsID, resourceGroup, snapshotName); rerr != nil {
+			azureutils.SleepIfThrottled(rerr.Error(), consts.SnapshotOpThrottlingSleepSec)
+			klog.Errorf("delete snapshot error: %v", rerr.Error())
+		} else {
+			klog.V(2).Infof("delete snapshot(%s) under rg(%s) region(%s) successfully", snapshotName, resourceGroup, d.cloud.Location)
+		}
+
+		csiSnapshot, err = d.getSnapshotByID(ctx, subsID, resourceGroup, crossRegionSnapshotName, sourceVolumeID)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
 	}
 
 	createResp := &csi.CreateSnapshotResponse{
@@ -1031,6 +1093,31 @@ func (d *Driver) getSnapshotInfo(snapshotID string) (snapshotName, resourceGroup
 		return "", "", "", fmt.Errorf("cannot get SubscriptionID from %s", snapshotID)
 	}
 	return snapshotName, resourceGroup, subsID, err
+}
+
+// waitForSnapshotCopy wait for copy incremental snapshot to a new region until completionPercent is 100.0
+func (d *Driver) waitForSnapshotCopy(ctx context.Context, subsID, resourceGroup, copySnapshotName string, intervel, timeout time.Duration) error {
+	timeAfter := time.After(timeout)
+	timeTick := time.Tick(intervel)
+
+	for {
+		select {
+		case <-timeTick:
+			copySnapshot, rerr := d.cloud.SnapshotsClient.Get(ctx, subsID, resourceGroup, copySnapshotName)
+			if rerr != nil {
+				return rerr.Error()
+			}
+
+			completionPercent := *copySnapshot.SnapshotProperties.CompletionPercent
+			klog.V(2).Infof("copy snapshot(%s) under rg(%s) region(%s) completionPercent: %f", copySnapshotName, resourceGroup, *copySnapshot.Location, completionPercent)
+			if completionPercent >= float64(100.0) {
+				klog.V(2).Infof("copy snapshot(%s) under rg(%s) region(%s) complete", copySnapshotName, resourceGroup, *copySnapshot.Location)
+				return nil
+			}
+		case <-timeAfter:
+			return fmt.Errorf("timeout waiting for copy snapshot(%s) under rg(%s)", copySnapshotName, resourceGroup)
+		}
+	}
 }
 
 func isAsyncAttachEnabled(defaultValue bool, volumeContext map[string]string) bool {
--- a/test/e2e/dynamic_provisioning_test.go
+++ b/test/e2e/dynamic_provisioning_test.go
@@ -799,6 +799,98 @@ func (t *dynamicProvisioningTestSuite) defineTests(isMultiZone bool) {
 		test.Run(cs, snapshotrcs, ns)
 	})
 
+	ginkgo.It("should create a pod with small storage size, take a volume snapshot cross region, and restore disk in another region [disk.csi.azure.com]", func() {
+		skipIfUsingInTreeVolumePlugin()
+		skipIfTestingInWindowsCluster()
+
+		pod := testsuites.PodDetails{
+			IsWindows:    isWindowsCluster,
+			WinServerVer: winServerVer,
+			Cmd:          convertToPowershellorCmdCommandIfNecessary("echo 'hello world' > /mnt/test-1/data"),
+			Volumes: t.normalizeVolumes([]testsuites.VolumeDetails{
+				{
+					FSType:    getFSType(isWindowsCluster),
+					ClaimSize: "10Gi",
+					VolumeMount: testsuites.VolumeMountDetails{
+						NameGenerate:      "test-volume-",
+						MountPathGenerate: "/mnt/test-",
+					},
+					VolumeAccessMode: v1.ReadWriteOnce,
+				},
+			}, isMultiZone),
+		}
+		podWithSnapshot := testsuites.PodDetails{
+			IsWindows:    isWindowsCluster,
+			WinServerVer: winServerVer,
+			Cmd:          convertToPowershellorCmdCommandIfNecessary("grep 'hello world' /mnt/test-1/data"),
+		}
+		test := testsuites.DynamicallyProvisionedVolumeSnapshotCrossRegionTest{
+			CSIDriver:              testDriver,
+			Pod:                    pod,
+			PodWithSnapshot:        podWithSnapshot,
+			StorageClassParameters: map[string]string{"skuName": "StandardSSD_LRS"},
+			SnapshotStorageClassParameters: map[string]string{
+				"incremental": "true", "dataAccessAuthMode": "AzureActiveDirectory", "location": "westus2",
+			},
+		}
+		if location == "westus2" {
+			test.SnapshotStorageClassParameters["location"] = "westeurope"
+		}
+		if isAzureStackCloud {
+			test.StorageClassParameters = map[string]string{"skuName": "Standard_LRS"}
+		}
+		if !isUsingInTreeVolumePlugin && supportsZRS {
+			test.StorageClassParameters = map[string]string{"skuName": "StandardSSD_ZRS"}
+		}
+		test.Run(cs, snapshotrcs, ns)
+	})
+
+	ginkgo.It("should create a pod with large storage size, take a volume snapshot cross region, and restore disk in another region [disk.csi.azure.com]", func() {
+		skipIfUsingInTreeVolumePlugin()
+		skipIfTestingInWindowsCluster()
+
+		pod := testsuites.PodDetails{
+			IsWindows:    isWindowsCluster,
+			WinServerVer: winServerVer,
+			Cmd:          convertToPowershellorCmdCommandIfNecessary("echo 'hello world' > /mnt/test-1/data"),
+			Volumes: t.normalizeVolumes([]testsuites.VolumeDetails{
+				{
+					FSType:    getFSType(isWindowsCluster),
+					ClaimSize: "100Gi",
+					VolumeMount: testsuites.VolumeMountDetails{
+						NameGenerate:      "test-volume-",
+						MountPathGenerate: "/mnt/test-",
+					},
+					VolumeAccessMode: v1.ReadWriteOnce,
+				},
+			}, isMultiZone),
+		}
+		podWithSnapshot := testsuites.PodDetails{
+			IsWindows:    isWindowsCluster,
+			WinServerVer: winServerVer,
+			Cmd:          convertToPowershellorCmdCommandIfNecessary("grep 'hello world' /mnt/test-1/data"),
+		}
+		test := testsuites.DynamicallyProvisionedVolumeSnapshotCrossRegionTest{
+			CSIDriver:              testDriver,
+			Pod:                    pod,
+			PodWithSnapshot:        podWithSnapshot,
+			StorageClassParameters: map[string]string{"skuName": "StandardSSD_LRS"},
+			SnapshotStorageClassParameters: map[string]string{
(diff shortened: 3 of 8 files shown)


pkg/azuredisk/controllerserver.go
```diff
@@ -885,16 +885,66 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 		mc.ObserveOperationWithResult(isOperationSucceeded, consts.SourceResourceID, sourceVolumeID, consts.SnapshotName, snapshotName)
 	}()
 
-	klog.V(2).Infof("begin to create snapshot(%s, incremental: %v) under rg(%s)", snapshotName, incremental, resourceGroup)
-	if rerr := d.cloud.SnapshotsClient.CreateOrUpdate(ctx, subsID, resourceGroup, snapshotName, snapshot); rerr != nil {
-		if strings.Contains(rerr.Error().Error(), "existing disk") {
-			return nil, status.Error(codes.AlreadyExists, fmt.Sprintf("request snapshot(%s) under rg(%s) already exists, but the SourceVolumeId is different, error details: %v", snapshotName, resourceGroup, rerr.Error()))
+	if location != d.cloud.Location {
+		snapshotNameLocal := azureutils.CreateValidDiskName("local_" + snapshotName)
```
edreed, line 889: Instead of duplicating the flow of taking the original snapshot, consider using the same flow to create the local snapshot and doing the copy only if the target region is different.

pkg/azuredisk/controllerserver.go
```diff
@@ -885,16 +885,66 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
...
+			klog.V(2).Infof("begin to create snapshot(%s, incremental: %v) under rg(%s) region(%s)", snapshotName, incremental, resourceGroup, location)
+			if rerr := d.cloud.SnapshotsClient.CreateOrUpdate(ctx, subsID, resourceGroup, snapshotName, copySnapshot); rerr != nil {
+				if strings.Contains(rerr.Error().Error(), "existing disk") {
+					return nil, status.Error(codes.AlreadyExists, fmt.Sprintf("request snapshot(%s) under rg(%s) already exists, but the SourceVolumeId is different, error details: %v", snapshotName, resourceGroup, rerr.Error()))
+				}
+
+				azureutils.SleepIfThrottled(rerr.Error(), azureconstants.SnapshotOpThrottlingSleepSec)
+				return nil, status.Error(codes.Internal, fmt.Sprintf("create snapshot error: %v", rerr.Error()))
+			}
+			klog.V(2).Infof("create snapshot(%s) under rg(%s) region(%s) successfully", snapshotName, resourceGroup, location)
```
edreed, line 927: It takes a non-trivial amount of time for the snapshot to be copied to the destination region, and this copy could fail. We should monitor the copy progress as described in [Copy an incremental snapshot to a new region](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-copy-incremental-snapshot-across-regions?tabs=azure-cli) and report status.

pkg/azuredisk/controllerserver.go
umagnus: Due to the local and copy snapshotName is the same if we use the same flow to create the local snapshot, then I use snapshotNameLocal instead of snapshotName when target region is different. I think the same name will receive exist error. I consider we can set different snapshotName in the beginning.

andyzhangx:
> The concern I have with this solution is that the snapshot is not useable in the current region and there would be no way to tell from the `VolumeSnapshotContent` object created for it.
> 
> Another solution would be to preserve the original snapshot in the local region and copy it to the target region with a known (or configurable) suffix. This way, the snapshot can be used in the local region to restore a disk and full CSI snapshot semantics preserved. If the user doesn't want the local snapshot copy, they can always delete it manually.

@edreed if we create two snapshots(locan & cross region) in one `CreateSnapshot` func, and we could only bound one azure snapshot resource to one VolumeSnapshotContent object, the other would be a ghost resource. Instead, user could create a cross region snapshot, and restore disk from that snapshot in the same region. And if users really want to create a local snapshot in the same region, they could specify the same region, I think that's more straightforward, the main concern is that we don't want to create a ghost resource in one `CreateSnapshot` func.

edreed:
> > The concern I have with this solution is that the snapshot is not useable in the current region and there would be no way to tell from the `VolumeSnapshotContent` object created for it.
> > Another solution would be to preserve the original snapshot in the local region and copy it to the target region with a known (or configurable) suffix. This way, the snapshot can be used in the local region to restore a disk and full CSI snapshot semantics preserved. If the user doesn't want the local snapshot copy, they can always delete it manually.
> 
> @edreed if we create two snapshots(local & cross region) in one `CreateSnapshot` func, and we could only bound one azure snapshot resource to one VolumeSnapshotContent object, the other would be a ghost resource. Instead, user could create a cross region snapshot, and restore disk from that snapshot in the same region. And if users really want to create a local snapshot in the same region, they could specify the same region, I think that's more straightforward, the main concern is that we don't want to create a ghost resource in one `CreateSnapshot` func.

It's either have a ghost resource or have an object that is not usable in the context in which it is represented. If the purpose of the cross-region snapshot is for recovery in the alternate region, it doesn't really make sense to have a snapshot object representing it in the local region. Accidentally deleting the local VS will delete the remote snapshot and data could be lost. This is likely to happen if someone tries to restore the snapshot to a local disk, discovers it doesn't work and deletes it as useless.

IMO, the CSI driver shouldn't be the one copying the snapshot to the remote region because of these consistency issues. This should be orchestrated by the application that is safeguarding the data because it has the whole view of the scenario where the CSI driver is limited to object represented in the local cluster context.

pkg/azuredisk/controllerserver.go
```diff
@@ -894,13 +900,47 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
...
+	klog.V(2).Infof("create snapshot(%s) under rg(%s) region(%s) successfully", snapshotName, resourceGroup, d.cloud.Location)
 
 	csiSnapshot, err := d.getSnapshotByID(ctx, subsID, resourceGroup, snapshotName, sourceVolumeID)
 	if err != nil {
 		return nil, err
 	}
 
+	if location != d.cloud.Location {
+		klog.V(2).Infof("create snapshot location(%s) is different from disk location(%s)", location, d.cloud.Location)
+		if incremental {
```
andyzhangx, line 912: what if `incremental` is false?

umagnus: Due to we can only copy incremental snapshot to new region, if `incremental` is false, we will return log for location is different and create local snapshot.

andyzhangx: if so, if `incremental` is false, and location is different from current k8s cluster, then it should return error.

umagnus: ok, I return error before the local snapshot is created.

pkg/azuredisk/controllerserver.go
```diff
@@ -879,13 +885,23 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 		snapshot.SnapshotProperties.DataAccessAuthMode = compute.DataAccessAuthMode(dataAccessAuthMode)
 	}
 
+	var copySnapshotName string
+	if location != d.cloud.Location {
```
andyzhangx, line 889: if location != "" && location != d.cloud.Location {

}

pkg/azuredisk/controllerserver.go
```diff
@@ -879,13 +885,23 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 		snapshot.SnapshotProperties.DataAccessAuthMode = compute.DataAccessAuthMode(dataAccessAuthMode)
 	}
 
+	var copySnapshotName string
```
andyzhangx, line 888: crossRegionSnapshotName

pkg/azuredisk/controllerserver.go
```diff
@@ -894,13 +910,51 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
...
 	}
-	klog.V(2).Infof("create snapshot(%s) under rg(%s) successfully", snapshotName, resourceGroup)
+	klog.V(2).Infof("create snapshot(%s) under rg(%s) region(%s) successfully", snapshotName, resourceGroup, d.cloud.Location)
 
 	csiSnapshot, err := d.getSnapshotByID(ctx, subsID, resourceGroup, snapshotName, sourceVolumeID)
 	if err != nil {
 		return nil, err
 	}
 
+	if location != d.cloud.Location {
```
andyzhangx, line 920: just use `if crossRegionSnapshotName != "" { ... }`

pkg/azuredisk/controllerserver.go
umagnus: done

pkg/azuredisk/controllerserver.go
umagnus: done

pkg/azuredisk/controllerserver.go
umagnus: done

test/e2e/dynamic_provisioning_test.go
```diff
@@ -799,6 +801,94 @@ func (t *dynamicProvisioningTestSuite) defineTests(isMultiZone bool) {
...
+		if isAzureStackCloud {
+			test.StorageClassParameters = map[string]string{"skuName": "Standard_LRS"}
+		}
+		if !isUsingInTreeVolumePlugin && supportsZRS {
+			test.StorageClassParameters = map[string]string{"skuName": "StandardSSD_ZRS"}
+		}
+		test.Run(cs, snapshotrcs, ns)
+	})
+
+	ginkgo.It("should create a pod with large storage size, take a volume snapshot cross region, and restore disk in another region [disk.csi.azure.com]", func() {
```
andyzhangx, line 848: this is the same case as in L804?

test/e2e/testsuites/dynamically_provisioned_volume_snapshot_tester.go
```diff
@@ -44,6 +44,7 @@ type DynamicallyProvisionedVolumeSnapshotTest struct {
 	CSIDriver                      driver.PVTestDriver
 	Pod                            PodDetails
 	ShouldOverwrite                bool
+	IsCrossRegion                  bool
```
andyzhangx, line 47: there is `location                  string` in `suite_test.go`, so this var is not necessary.

test/e2e/dynamic_provisioning_test.go
andyzhangx: for such case, what about create a 100GB disk in `eastus` and then do cross region disk snapshot in `eastus2`, then the whole time cost is controllable, and also write a standalone test struct, e.g. `DynamicallyProvisionedVolumeSnapshotCrossRegionTest`

deploy/csi-azuredisk-controller.yaml
```diff
@@ -86,6 +86,7 @@ spec:
             - "-leader-election"
             - "--leader-election-namespace=kube-system"
             - "--v=2"
+            - "--timeout=660s"
```
andyzhangx, line 89: let's make it as 600s

pkg/azuredisk/controllerserver.go
```diff
@@ -1033,6 +1093,31 @@ func (d *Driver) getSnapshotInfo(snapshotID string) (snapshotName, resourceGroup
...
+		select {
+		case <-timeTick:
+			copySnapshot, rerr := d.cloud.SnapshotsClient.Get(ctx, subsID, resourceGroup, copySnapshotName)
+			if rerr != nil {
+				return status.Error(codes.Internal, fmt.Sprintf("get snapshot error: %v", rerr.Error()))
+			}
+
+			completionPercent := *copySnapshot.SnapshotProperties.CompletionPercent
+			klog.V(2).Infof("copy snapshot(%s) under rg(%s) region(%s) completionPercent: %f", copySnapshotName, resourceGroup, *copySnapshot.Location, completionPercent)
+			if completionPercent == float64(100.0) {
```
andyzhangx, line 1111: `>=`

pkg/azuredisk/controllerserver.go
```diff
@@ -894,13 +915,48 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
...
 
 	csiSnapshot, err := d.getSnapshotByID(ctx, subsID, resourceGroup, snapshotName, sourceVolumeID)
 	if err != nil {
 		return nil, err
 	}
 
+	if crossRegionSnapshotName != "" {
+		klog.V(2).Infof("begin to copy an incremental snapshot from region(%s) to new region(%s)", d.cloud.Location, location)
+		copySnapshot := snapshot
+		copySnapshot.SnapshotProperties.CreationData.SourceResourceID = &csiSnapshot.SnapshotId
```
andyzhangx, line 928: add nil pointer check for `SnapshotProperties` and `CreationData`

pkg/azuredisk/controllerserver.go
```diff
@@ -894,13 +915,48 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
...
+		copySnapshot.SnapshotProperties.CreationData.CreateOption = compute.CopyStart
+		copySnapshot.Location = &location
+
+		klog.V(2).Infof("begin to create snapshot(%s, incremental: %v) under rg(%s) region(%s)", crossRegionSnapshotName, incremental, resourceGroup, location)
+		if rerr := d.cloud.SnapshotsClient.CreateOrUpdate(ctx, subsID, resourceGroup, crossRegionSnapshotName, copySnapshot); rerr != nil {
+			if strings.Contains(rerr.Error().Error(), "existing disk") {
```
andyzhangx, line 934: if it's existing disk, then return success?

pkg/azuredisk/controllerserver.go
```diff
@@ -894,13 +915,48 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
...
+		klog.V(2).Infof("create snapshot(%s) under rg(%s) region(%s) successfully", crossRegionSnapshotName, resourceGroup, location)
+
+		if err := d.waitForSnapshotCopy(ctx, subsID, resourceGroup, crossRegionSnapshotName, waitForSnapshotCopyInterval, waitForSnapshotCopyTimeout); err != nil {
+			return nil, fmt.Errorf("failed to wait for snapshot copy cross region: %w", err)
+		}
+
+		klog.V(2).Infof("begin to delete snapshot(%s) under rg(%s) region(%s)", snapshotName, resourceGroup, d.cloud.Location)
+		if rerr := d.cloud.SnapshotsClient.Delete(ctx, subsID, resourceGroup, snapshotName); rerr != nil {
+			azureutils.SleepIfThrottled(rerr.Error(), consts.SnapshotOpThrottlingSleepSec)
+			return nil, status.Error(codes.Internal, fmt.Sprintf("delete snapshot error: %v", rerr.Error()))
```
andyzhangx, line 950: don't return error here, just provide a warning log here since snapshot deletion failure is not blocking issue

pkg/azuredisk/controllerserver.go
```diff
@@ -879,13 +885,28 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
...
+	if location != "" && location != d.cloud.Location {
+		if incremental {
+			crossRegionSnapshotName = snapshotName
+			snapshotName = azureutils.CreateValidDiskName("local_" + snapshotName)
+		} else {
+			return nil, status.Errorf(codes.InvalidArgument, "could not create snapshot cross region with incremental is false")
+		}
+	}
+
+	if acquired := d.volumeLocks.TryAcquire(snapshotName); !acquired {
```
andyzhangx, line 898: move to L887, let's only lock on target snapshotName

pkg/azuredisk/controllerserver.go
```diff
@@ -927,6 +983,10 @@ func (d *Driver) DeleteSnapshot(ctx context.Context, req *csi.DeleteSnapshotRequ
 		}
 	}
 
+	if hasLock := d.volumeLocks.HasLock(snapshotName); hasLock {
+		return nil, status.Errorf(codes.Aborted, "snapshot %s is being copied to a new region", snapshotName)
```
andyzhangx, line 987: this is incorrect, just use following way to lock:
```
	if acquired := d.volumeLocks.TryAcquire(snapshotName); !acquired {
		return nil, status.Errorf(codes.Aborted, volumeOperationAlreadyExistsFmt, snapshotName)
	}
	defer d.volumeLocks.Release(snapshotName)
```

test/e2e/dynamic_provisioning_test.go
umagnus: We test two cases about succeed and timeout, the test result of timeout is failed because of timeout error. So should we only use one case?

test/e2e/testsuites/dynamically_provisioned_volume_snapshot_tester.go
umagnus: use `DynamicallyProvisionedVolumeSnapshotCrossRegionTest` and use different location in test case

deploy/csi-azuredisk-controller.yaml
umagnus: done

pkg/azuredisk/controllerserver.go
umagnus: done

pkg/azuredisk/controllerserver.go
umagnus: But they are different SourceVolumeId, I see in local snapshot creating it also return error.

pkg/azuredisk/controllerserver.go
umagnus: done.

pkg/azuredisk/controllerserver.go
umagnus: done. No need to add lock in delete snapshot.

test/e2e/dynamic_provisioning_test.go
umagnus: add `DynamicallyProvisionedVolumeSnapshotCrossRegionTest` standalone test struct.

pkg/azuredisk/controllerserver.go
umagnus: done

pkg/azuredisk/controllerserver.go
```diff
@@ -45,6 +46,11 @@ import (
 	azure "sigs.k8s.io/cloud-provider-azure/pkg/provider"
 )
 
+const (
+	waitForSnapshotCopyInterval = 20 * time.Second
```
andyzhangx, line 50: 20s interval is too big, what about 5s?

pkg/azuredisk/controllerserver.go
```diff
@@ -894,13 +915,55 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
...
+		if rerr := d.cloud.SnapshotsClient.Delete(ctx, subsID, resourceGroup, snapshotName); rerr != nil {
+			azureutils.SleepIfThrottled(rerr.Error(), consts.SnapshotOpThrottlingSleepSec)
+			klog.Warningf("delete snapshot error: %v", rerr.Error())
+		} else {
+			klog.V(2).Infof("delete snapshot(%s) under rg(%s) region(%s) successfully", snapshotName, resourceGroup, d.cloud.Location)
+		}
+
+		csiSnapshot, err = d.getSnapshotByID(ctx, subsID, resourceGroup, crossRegionSnapshotName, sourceVolumeID)
+		if err != nil {
+			return nil, err
```
andyzhangx, line 963: `status.Error(codes.Internal, ...`

pkg/azuredisk/controllerserver.go
```diff
@@ -1033,6 +1096,31 @@ func (d *Driver) getSnapshotInfo(snapshotID string) (snapshotName, resourceGroup
...
+func (d *Driver) waitForSnapshotCopy(ctx context.Context, subsID, resourceGroup, copySnapshotName string, intervel, timeout time.Duration) error {
+	timeAfter := time.After(timeout)
+	timeTick := time.Tick(intervel)
+
+	for {
+		select {
+		case <-timeTick:
+			copySnapshot, rerr := d.cloud.SnapshotsClient.Get(ctx, subsID, resourceGroup, copySnapshotName)
+			if rerr != nil {
+				return status.Error(codes.Internal, fmt.Sprintf("get snapshot error: %v", rerr.Error()))
```
andyzhangx, line 1109: just `return rerr.Error()`

andyzhangx: I think you could set as 1s as intervel, but print percentage log once every 10s, is that possible? otherwise there would be too many logs.

pkg/azuredisk/controllerserver.go
```diff
@@ -894,13 +915,55 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
...
+				return nil, status.Error(codes.AlreadyExists, fmt.Sprintf("request snapshot(%s) under rg(%s) already exists, but the SourceVolumeId is different, error details: %v", crossRegionSnapshotName, resourceGroup, rerr.Error()))
+			}
+
+			azureutils.SleepIfThrottled(rerr.Error(), consts.SnapshotOpThrottlingSleepSec)
+			return nil, status.Error(codes.Internal, fmt.Sprintf("create snapshot error: %v", rerr.Error()))
+		}
+		klog.V(2).Infof("create snapshot(%s) under rg(%s) region(%s) successfully", crossRegionSnapshotName, resourceGroup, location)
+
+		if err := d.waitForSnapshotCopy(ctx, subsID, resourceGroup, crossRegionSnapshotName, waitForSnapshotCopyInterval, waitForSnapshotCopyTimeout); err != nil {
+			return nil, fmt.Errorf("failed to wait for snapshot copy cross region: %w", err)
```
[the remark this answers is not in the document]
umagnus: done

pkg/azuredisk/controllerserver.go
```diff
@@ -894,13 +915,55 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
...
-	klog.V(2).Infof("create snapshot(%s) under rg(%s) successfully", snapshotName, resourceGroup)
+	klog.V(2).Infof("create snapshot(%s) under rg(%s) region(%s) successfully", snapshotName, resourceGroup, d.cloud.Location)
 
 	csiSnapshot, err := d.getSnapshotByID(ctx, subsID, resourceGroup, snapshotName, sourceVolumeID)
 	if err != nil {
 		return nil, err
 	}
 
+	if crossRegionSnapshotName != "" {
+		klog.V(2).Infof("begin to copy an incremental snapshot from region(%s) to new region(%s)", d.cloud.Location, location)
```
andyzhangx, line 926: remove this logging since it's duplicated as L938

pkg/azuredisk/controllerserver.go
```diff
@@ -894,13 +915,55 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
...
+		klog.V(2).Infof("create snapshot(%s) under rg(%s) region(%s) successfully", crossRegionSnapshotName, resourceGroup, location)
+
+		if err := d.waitForSnapshotCopy(ctx, subsID, resourceGroup, crossRegionSnapshotName, waitForSnapshotCopyInterval, waitForSnapshotCopyTimeout); err != nil {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("failed to wait for snapshot copy cross region: %v", err.Error()))
+		}
+
+		klog.V(2).Infof("begin to delete snapshot(%s) under rg(%s) region(%s)", snapshotName, resourceGroup, d.cloud.Location)
+		if rerr := d.cloud.SnapshotsClient.Delete(ctx, subsID, resourceGroup, snapshotName); rerr != nil {
+			azureutils.SleepIfThrottled(rerr.Error(), consts.SnapshotOpThrottlingSleepSec)
+			klog.Warningf("delete snapshot error: %v", rerr.Error())
```
andyzhangx, line 956: klog.Errorf

umagnus: done

andyzhangx (approved):
/lgtm

PR #1826, opened 2023-05-02
[V2] fix: update predicate of replica controller to prevent over creating replica AzVolumeAttachment

hccheng72:
**What type of PR is this?**
/kind bug

Once detachment of an azvolumeattachment failed, we observed that more than one `handleReplicaDelete` and `manageReplicas` were callled in separated goroutines. We don't expect to overly create the CRI with the reasons that could make attach workload heavier or take a risk that more than expected number of CRI get attached.

This happened because all update events will trigger reconciler of replica controller. For a detaching operation, adding detach request annotation, update state to Detaching, DetachmentFailed, and ForceDetachPending all trigger the reconciler to call `handleReplicaDelete` in a goroutine. We want to ensure only one call during a detaching cycle.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/controller/replica.go
+++ b/pkg/controller/replica.go
@@ -75,85 +75,77 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
 			// If not, cancel scheduled garbage collection if there is one enqueued
 			r.removeGarbageCollection(azVolumeAttachment.Spec.VolumeName)
 
-			// If promotion event, create a replacement replica
-			if isAttached(azVolumeAttachment) && azVolumeAttachment.Status.Detail.PreviousRole == azdiskv1beta2.ReplicaRole {
-				r.triggerManageReplica(azVolumeAttachment.Spec.VolumeName)
+			// If promotion event, create a replacement replica after promotion update is done
+			if isAttached(azVolumeAttachment) && isPromotionRequested(azVolumeAttachment) {
+				go r.handlePromoteUpdate(context.Background(), azVolumeAttachment)
 			}
 		}
 	} else {
 		// queue garbage collection if a primary attachment is being demoted
 		if isDemotionRequested(azVolumeAttachment) {
-			//nolint:contextcheck // Garbage collection is asynchronous; context is not inherited by design
-			r.triggerGarbageCollection(ctx, azVolumeAttachment.Spec.VolumeName)
-			return reconcile.Result{}, nil
-		}
-
-		if deleteRequested, _ := objectDeletionRequested(azVolumeAttachment); !deleteRequested {
-			// create a replica attachment was requested for a detach, create a replacement replica if the attachment is not being cleaned up
-			if volumeDetachRequested(azVolumeAttachment) {
-				if azVolumeAttachment.Status.State == azdiskv1beta2.DetachmentFailed {
-					// if detachment failed and the driver is uninstalling, delete azvolumeattachment CRI to let uninstallation proceed
-					if r.driverLifecycle.IsDriverUninstall() {
-						r.logger.Info("Deleting AzVolumeAttachment in DetachmentFailed state without detaching disk", workflow.GetObjectDetails(azVolumeAttachment)...)
-						if err := r.cachedClient.Delete(ctx, azVolumeAttachment); err != nil {
-							return reconcile.Result{Requeue: true}, err
-						}
-					} else {
-						if _, err := azureutils.UpdateCRIWithRetry(ctx, nil, r.cachedClient, r.azClient, azVolumeAttachment, func(obj client.Object) error {
-							azVolumeAttachment := obj.(*azdiskv1beta2.AzVolumeAttachment)
-							_, err = updateState(azVolumeAttachment, azdiskv1beta2.ForceDetachPending, normalUpdate)
-							return err
-						}, consts.NormalUpdateMaxNetRetry, azureutils.UpdateCRIStatus); err != nil {
-							return reconcile.Result{Requeue: true}, err
-						}
-					}
-				}
-
-				if !isCleanupRequested(azVolumeAttachment) {
-					go r.handleReplicaDelete(context.Background(), azVolumeAttachment)
-				}
-			} else if azVolumeAttachment.Status.State == azdiskv1beta2.AttachmentFailed {
-				// if the failed attachment isn't retriable, delete the CRI so that replacing replica AzVolumeAttachment can be created
-				// once the attachment has ever been retriable, it should be retried until it succeeds or gets detached
-				if !azureutils.MapContains(azVolumeAttachment.Status.Annotations, consts.ReplicaVolumeAttachRetryAnnotation) && !azureutils.MapContains(azVolumeAttachment.Status.Annotations, consts.ReplicaVolumeAttachRetryCount) {
+			go r.handleDemoteUpdate(context.Background(), azVolumeAttachment)
+		} else if volumeDetachRequested(azVolumeAttachment) {
+			if azVolumeAttachment.Status.State == azdiskv1beta2.DetachmentFailed {
+				// if detachment failed and the driver is uninstalling, delete azvolumeattachment CRI to let uninstallation proceed
+				if r.driverLifecycle.IsDriverUninstall() {
+					r.logger.Info("Deleting AzVolumeAttachment in DetachmentFailed state without detaching disk", workflow.GetObjectDetails(azVolumeAttachment)...)
 					if err := r.cachedClient.Delete(ctx, azVolumeAttachment); err != nil {
 						return reconcile.Result{Requeue: true}, err
 					}
-					go r.handleReplicaDelete(context.Background(), azVolumeAttachment)
-					// else, reset the state to AttachmentPending to retry the attachment
 				} else {
-					retryCount := 0
-					if value, exists := azureutils.GetFromMap(azVolumeAttachment.Status.Annotations, consts.ReplicaVolumeAttachRetryCount); exists {
-						if n, err := strconv.Atoi(value); err == nil {
-							retryCount = n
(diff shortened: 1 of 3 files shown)


hccheng72:
@sunpa93

pkg/controller/replica.go
```diff
@@ -86,74 +86,68 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
...
-							azva := obj.(*azdiskv1beta2.AzVolumeAttachment)
-							azureutils.RemoveFromMap(azva.Status.Annotations, consts.ReplicaVolumeAttachRetryAnnotation)
-							azureutils.AddToMap(azva.Status.Annotations, consts.ReplicaVolumeAttachRetryCount, strconv.Itoa(retryCount+1))
-							_, derr := updateState(azva, azdiskv1beta2.AttachmentPending, forceUpdate)
-							azva.Status.Detail = nil
-							azva.Status.Error = nil
-							return derr
-						}
+				}
+			} else if !isCleanupRequested(azVolumeAttachment) {
```
sunpa93, line 107: Now that there is no check for `objectDeletionRequested` at https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/0c0cbfc785a90526cc7beb51b04e7bb6e82d493a/pkg/controller/replica.go#LL89C55-L89C55, we should move this check to https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/0c0cbfc785a90526cc7beb51b04e7bb6e82d493a/pkg/controller/replica.go#LL111C4-L111C4

pkg/controller/replica.go
```diff
@@ -300,13 +291,29 @@ func NewReplicaController(mgr manager.Manager, controllerSharedState *SharedStat
 			return false
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			return true
+			objOld, okOld := e.ObjectOld.(*azdiskv1beta2.AzVolumeAttachment)
+			objNew, okNew := e.ObjectNew.(*azdiskv1beta2.AzVolumeAttachment)
+
```
sunpa93, line 296: Consider moving `if okOld && okNew` check to here.

pkg/controller/replica.go
```diff
@@ -300,13 +291,29 @@ func NewReplicaController(mgr manager.Manager, controllerSharedState *SharedStat
...
+			attachFailedUpdate := objOld.Status.State != azdiskv1beta2.AttachmentFailed && objNew.Status.State == azdiskv1beta2.AttachmentFailed
+			detachFailedUpdate := objOld.Status.State != azdiskv1beta2.DetachmentFailed && objNew.Status.State == azdiskv1beta2.DetachmentFailed
+
+			deleteRequestedOld, _ := objectDeletionRequested(objOld)
+			deleteRequestedNew, _ := objectDeletionRequested(objNew)
+			deleteTimestampUpdate := !deleteRequestedOld && deleteRequestedNew
+
+			updateEventsForReplica := demoteUpdate || attachFailedUpdate || deleteTimestampUpdate || detachFailedUpdate
+
+			if okOld && okNew && (objNew.Spec.RequestedRole != azdiskv1beta2.ReplicaRole || updateEventsForReplica) {
```
sunpa93, line 307: Consider defining `updateEventsForPrimaries := (objNew.Spec.RequestedRole == azdiskv1beta2.PrimaryRole) && (attachEvent || promotionEvent || deleteEvent)`. I

pkg/controller/replica.go
```diff
@@ -300,13 +291,29 @@ func NewReplicaController(mgr manager.Manager, controllerSharedState *SharedStat
...
+
+			demoteUpdate := !isDemotionRequested(objOld) && isDemotionRequested(objNew)
+			attachFailedUpdate := objOld.Status.State != azdiskv1beta2.AttachmentFailed && objNew.Status.State == azdiskv1beta2.AttachmentFailed
+			detachFailedUpdate := objOld.Status.State != azdiskv1beta2.DetachmentFailed && objNew.Status.State == azdiskv1beta2.DetachmentFailed
+
+			deleteRequestedOld, _ := objectDeletionRequested(objOld)
+			deleteRequestedNew, _ := objectDeletionRequested(objNew)
+			deleteTimestampUpdate := !deleteRequestedOld && deleteRequestedNew
+
+			updateEventsForReplica := demoteUpdate || attachFailedUpdate || deleteTimestampUpdate || detachFailedUpdate
```
sunpa93, line 305: Consider adding `objNew.Spec.RequestedRole == azdiskv1beta2.ReplicaRole && (demoteUpdate || attachFailedUpdate || deleteTimestampUpdate || detachFailedUpdate))` to make distinction between update events to be watched for _replicas_ vs. _primaries_.

sunpa93: We would need to check for a detach requested event. Otherwise, we wouldn't never be able to trigger a `go replicaHandleDelete` for detach operation.

hccheng72: We would trigger `go replicaHandleDelete` with `deleteTimestampUpdate` after the completion of a detach operation.

sunpa93: Right 👍

pkg/controller/replica.go
```diff
@@ -300,13 +313,34 @@ func NewReplicaController(mgr manager.Manager, controllerSharedState *SharedStat
 			return false
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			return true
+			objOld, okOld := e.ObjectOld.(*azdiskv1beta2.AzVolumeAttachment)
+			objNew, okNew := e.ObjectNew.(*azdiskv1beta2.AzVolumeAttachment)
+			if okOld && okNew {
+				demoteRequestUpdate := isDemotionRequested(objOld) && isDemotionRequested(objNew)
```
sunpa93, line 319: Should this be `!isDemotionRequested(objOld) && isDemotionRequested(objNew)`?

pkg/controller/replica.go
```diff
@@ -300,13 +313,34 @@ func NewReplicaController(mgr manager.Manager, controllerSharedState *SharedStat
...
+				promoteRequestUpdate := isPromotionRequested(objOld) && isPromotionRequested(objNew)
```
sunpa93, line 320: Should this be `!isPromotionRequested(objOld) && isPromotionRequested(objNew)`?

pkg/controller/replica.go
hccheng72: yeah, you're right : )

pkg/controller/replica.go
```diff
@@ -75,104 +75,117 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
...
+	defer waiter.Close()
+
+	if _, err := waiter.Wait(ctx); err == nil {
+		//nolint:contextcheck // Garbage collection is asynchronous; context is not inherited by design
+		r.triggerGarbageCollection(ctx, azVolumeAttachment.Spec.VolumeName)
+	}
+}
+
+func (r *ReconcileReplica) handleReplicaDelete(ctx context.Context, azVolumeAttachment *azdiskv1beta2.AzVolumeAttachment, caller operationRequester) {
+	w, _ := workflow.GetWorkflowFromContext(ctx)
```
sunpa93, line 178: `GetWorkflowFromContext` will return an empty workflow as the context being fed to `handleReplicaDelete` is `context.Background()`. For such, consider using `workflow.GetWorkflow(ctx, azVolumeAttachment)` as it constructs workflow from an object if workflow cannot be found within the context.

pkg/controller/replica.go
```diff
@@ -75,104 +75,117 @@ func (r *ReconcileReplica) Reconcile(ctx context.Context, request reconcile.Requ
...
+	w.Logger().Infof("handleReplicaDelete requested by %s for AzVolumeAttachment (%s)", caller, azVolumeAttachment.Name)
```
sunpa93, line 179: Consider increasing the verbosity of this log.

pkg/controller/replica.go
```diff
@@ -300,13 +313,34 @@ func NewReplicaController(mgr manager.Manager, controllerSharedState *SharedStat
...
+				deleteRequestedOld, _ := objectDeletionRequested(objOld)
+				deleteRequestedNew, _ := objectDeletionRequested(objNew)
+				deleteTimestampUpdate := !deleteRequestedOld && deleteRequestedNew
+
+				updateEventsForPrimary := (objNew.Spec.RequestedRole == azdiskv1beta2.PrimaryRole) && (attachRequestUpdate || promoteRequestUpdate || detachRequestUpdate)
+				updateEventsForReplica := (objNew.Spec.RequestedRole == azdiskv1beta2.ReplicaRole) && (demoteRequestUpdate || attachFailedUpdate || deleteTimestampUpdate || detachFailedUpdate)
+
+				if updateEventsForPrimary || updateEventsForReplica {
+					return true
+				}
```
sunpa93, line 335: This can be summed to `return updateEventsForPrimary || updateEventsForReplica` :)

pkg/controller/replica.go
hccheng72: Or can we feed `handleReplicaDelete` with `ctx`, the parameter of reconcile function, instead of `context.Background()`?

sunpa93: Using the reconciler context for a go routine would be unsafe as this context would be managed by the parent function.

pkg/controller/replica.go
hccheng72: Nice suggestion!

pkg/controller/replica.go
hccheng72: Understood, thanks for confirming.

PR #1868, opened 2023-06-02
[V2] chore: do not expose cloud-provider-azure from provisioner package

edreed:
**What type of PR is this?**

/kind cleanup

This PR changes the [`controller.CloudProvisioner`](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/df99e3c3b38acf9c3f3f4e159f7d75f4e1069c8c/pkg/controller/common.go#LL125C6-L125C6) interface to prevent exposing `cloud-provider-azure` to its dependent packages. This improves layering and reduces interdependency among packages in the V2 implementation.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azureutils/azure_disk_utils.go
```diff
@@ -43,6 +43,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 	clientset "k8s.io/client-go/kubernetes"
```
mukhoakash, line 47: I believe we can remove `clientset`?

pkg/azurediskplugin/main.go
```diff
@@ -271,11 +217,13 @@ func getDriverConfig() *azdiskv1beta2.AzDiskDriverConfiguration {
 			}
 		})
 	}
+
 	driverConfig.NodeConfig.NodeID = *nodeID
-	if driverConfig == (azdiskv1beta2.AzDiskDriverConfiguration{}) {
+	if *driverConfig == (azdiskv1beta2.AzDiskDriverConfiguration{}) {
```
landreasyan, line 222: Since `driverConfig` is a pointer and is initialized conditionally (although currently it will always be initialized), we should probably check for nil before dereferencing and checking for the empty value.

PR #1909, opened 2023-07-18
fix: cloned volume could not be recognized on Windows node when source volume is mounted on the same node

andyzhangx:
**What type of PR is this?**
/kind bug

fix: ensure that readonly is disabled when disks are brought online
This PR could fix the cloned volume not recognized on the same Windows node (when source volume is also mounted on the same node)

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: ensure that readonly is disabled when disks are brought online
```

### DIFF
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -58,6 +58,7 @@ var (
 	isWindowsCluster          = os.Getenv(testWindowsEnvVar) != ""
 	winServerVer              = os.Getenv(testWinServerVerEnvVar)
 	isAzureStackCloud         = strings.EqualFold(os.Getenv(cloudNameEnvVar), "AZURESTACKCLOUD")
+	isWindowsHPCDeployment    = strings.EqualFold(os.Getenv("WINDOWS_USE_HOST_PROCESS_CONTAINERS"), "true")
 	location                  string
 	supportsZRS               bool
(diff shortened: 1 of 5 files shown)


test/e2e/suite_test.go
```diff
@@ -346,6 +347,9 @@ func convertToPowershellorCmdCommandIfNecessary(command string) string {
 		return "echo 'overwrite' | Out-File -FilePath C:\\mnt\\test-1\\data.txt; Start-Sleep 3600"
 	case "grep 'hello world' /mnt/test-1/data":
 		return "Get-Content C:\\mnt\\test-1\\data.txt | findstr 'hello world'"
+	case "df -h | grep /mnt/test- | awk '{print $2}' | grep 20.0G":
+		// todo: find a way to get the size of the volume on Windows node
```
lippertmarkus, line 351: `fsutil volume diskfree C:\\mnt\\test-1\\ | Select-String "Total bytes" | Select-String "15.0 GB"`

lippertmarkus (changes requested):

test/e2e/suite_test.go
andyzhangx: thanks, fixed.

PR #1935, opened 2023-08-14
feat(helm) Allow setting an alternate name of the configmap for the azure credential path

mtougeron:
**What type of PR is this?**

/kind feature

This allows setting an name for the azure credential configmap. This allows this chart and the `azurefile-csi-driver` chart to not conflict with each other over the name of the configmap (PR incoming for that chart as well). This is important for a GitOps workflow where we don't want an order of operations dependency between these two charts.

We install these charts via a wrapper chart that installs the additional configmap manifest with the appropriate config information. However, both charts cannot contain that extra manifest or Argo CD would create warnings that more than 1 chart owns the configmap. This means that the azuredisk and azurefile wrapper charts would need to install the configmap with different names.

Otherwise, if only one of the two wrapper charts installed the additional manifest, there would be an order of operations dependency on the two charts. This creates additional complexity that Argo CD does not have a mechanism to handle.

Fixes # n/a

**Requirements**:
- [x] includes documentation
- [n/a] adds unit tests

**Release note**:
```
Allow setting an alternate name of the configmap for the azure credential path in the helm chart
```


adriananeci (commented):
/lgtm

mtougeron:
appears unrelated to my changes

/retest-required

andyzhangx:
@mtougeron thanks for the contribution, does that mean `azure-cred-file` config map differs with Azure Disk and File CSI drivers, usually you could share the same `azure-cred-file` config maps.

mtougeron:
The problem is less about the contents of the configmap and more about the management of the configmap from an Argo CD perspective. Argo CD doesn't let two different applications manage the same configmap. The way it is now, we would have to either have a 3rd helm application that manages just this configmap or have an order of operations issue that Argo CD can't handle. Technically the content of the configmap would be the same but this change would allow each application to manage its own configmap without conflict with the other. It would allow us to install just one without the other or any sort of dependency on each other.

andyzhangx (approved):
/lgtm

mtougeron:
Thanks for the approval/merge!

PR #1981, opened 2023-09-28
fix: ensure Kubernets conformant format for location

daniel-weisse:
**What type of PR is this?**

/kind bug

When creating the [cloud provider config file](https://cloud-provider-azure.sigs.k8s.io/install/configs/), one can specify the `location` in two (valid for Azure) formats:
* upper case with spaces, e.g. "East US"
* lower case with no spaces, e.g. "eastus"

The driver (as well as other applications making use of the config) will accept and work with both formats.
An issue only arises if nodes in the cluster are created in specific availability zones, e.g. "East US 2":

The topology key (`topology.disk.csi.azure.com/zone`) of the node will always be set to non zone.

The `CreateVolume` rpc of the controller server will be unable to schedule the disk in the correct zone, since the azuredisk topology key is set to an empty string.
If one sets up  the well-known topology key instead (`topology.kubernetes.io/zone`) (or updates the azuredisk key manually), creating the disk in the correct zone still fails, because of how the [`PickAvailabilityZone`](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/7e9f4106c06eef1eac81e13425f62f67b23cb715/pkg/azureutils/azure_disk_utils.go#L715) checks for a valid zone.

This PR aims to fix this issue by ensuring the driver is always set up using the lower case no-space format, even if the config declares the format with spaces.

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azureutils/azure_disk_utils.go
+++ b/pkg/azureutils/azure_disk_utils.go
@@ -224,6 +224,11 @@ func GetCloudProviderFromClient(ctx context.Context, kubeClient *clientset.Clien
 			return nil, fmt.Errorf("no cloud config provided, error: %v", err)
 		}
 	} else {
+		// Location may be either upper case with spaces (e.g. "East US") or lower case without spaces (e.g. "eastus")
+		// Kubernetes does not allow whitespaces in label values, e.g. for topology keys
+		// ensure Kubernetes compatible format for Location by enforcing lowercase-no-space format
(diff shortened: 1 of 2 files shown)


andyzhangx (commented):
/ok-to-test

andyzhangx (approved):
/lgtm
/retest

PR #2155, opened 2024-01-15
Refactor: migrate disk client to track2 sdk

MartinForReal:
**What type of PR is this?**
Refactor: migrate disk client to track2 sdk

/kind cleanup

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


pkg/azuredisk/azure_controller_common.go
```diff
@@ -643,13 +646,17 @@ func (c *controllerCommon) checkDiskExists(ctx context.Context, diskURI string)
 		return false, err
 	}
 
-	if _, rerr := c.cloud.DisksClient.Get(ctx, subsID, resourceGroup, diskName); rerr != nil {
-		if rerr.HTTPStatusCode == http.StatusNotFound {
+	diskClient, err := c.clientFactory.GetDiskClientForSub(subsID)
+	if err != nil {
+		return false, err
+	}
+	if _, err := diskClient.Get(ctx, resourceGroup, diskName); err != nil {
```
andyzhangx, line 653: why not add `subsID` in `DisksClient.Get` func? there would be easier to use in csi driver

MartinForReal: Yes. I agree. but the subid is part of client object. https://github.com/Azure/azure-sdk-for-go/blob/775c7070836e584f796bc72c6de5c77649a67bb2/sdk/resourcemanager/compute/armcompute/disks_client.go#L27

andyzhangx: could you write a wrapper func in this driver for this? I see this code is duplicated multiple times in this PR.
```
diskClient, err := c.clientFactory.GetDiskClientForSub(subsID)
if err != nil {
    return false, err
}
if _, err := diskClient.Get(ctx, resourceGroup, diskName); err != nil {
```

andyzhangx: this `diskClient.Get` func is called 13 times in this PR, why not wrap as a Driver func, that could reduce duplicated code a lot

MartinForReal: `diskClient.Delete` is also invoked , diskclient can't be shared in one function if we put it into wrapper.

andyzhangx (approved):
/lgtm

PR #2383, opened 2024-06-28
feat: add ReservedDataDiskSlotNum to helm-chart 

### LINKED ISSUE #2382: ReservedDataDiskSlotNum not available in helm chart
ReservedDataDiskSlotNum should be put into the helm-chart
ReservedDataDiskSlotNum is not given the the Driver struct from the DriverOptions struct

Being able to use it in the helmchart

ClementLachaussee:
**What type of PR is this?**

/kind feature

The option is not available in the HelmChart, and the option is never given from the DriverOptiont struct to Driver struct. 

Fixes #2382


charts/latest/azuredisk-csi-driver/values.yaml
```diff
@@ -168,6 +168,8 @@ driver:
   # maximum number of attachable volumes per node,
   # maximum number is defined according to node instance type by default(-1)
   volumeAttachLimit: -1
+  # reserved data disk slot number per node
+  reservedDataDiskSlotNum: 0
```
andyzhangx, line 172: could you move to `node.reservedDataDiskSlotNum`

ClementLachaussee:
Hey @andyzhangx !
Thanks for your quick review, I have modified as requested and taken the liberty to modify the comment.

pkg/azuredisk/azuredisk.go
```diff
@@ -145,6 +145,7 @@ func newDriverV1(options *DriverOptions) *Driver {
 	driver.Version = driverVersion
 	driver.NodeID = options.NodeID
 	driver.VolumeAttachLimit = options.VolumeAttachLimit
+	driver.ReservedDataDiskSlotNum = options.ReservedDataDiskSlotNum
```
andyzhangx, line 148: it seems this is a bug, `ReservedDataDiskSlotNum` never takes effect in the before?

andyzhangx: can you make this change as a separate PR? I will cherrypick to old releases, thx

ClementLachaussee: done in !2384

ClementLachaussee: No it doesn't after testing the CSIDriver object is never modified

andyzhangx (approved):
/lgtm

andyzhangx:
> @andyzhangx do you know when this can be added to a release ?

@ClementLachaussee I am adding this chart config into v1.29.7 release since it does not have that ReservedDataDiskSlotNum assign bug:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/2395

PR #2384, opened 2024-06-28
fix: add ReservedDataDiskSlotNum copy from DriverOptions to Driver struct

ClementLachaussee:
**What type of PR is this?**

/kind bug

Adds code that is required for the parameter ReservedDataDiskSlotNum to be used 

https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/pkg/azuredisk/nodeserver.go#L395 was not modified even with VolumeAttachLimit < 0 
It seems the option wasn't given to the Driver.ReservedDataDiskSlotNum which was = to 0

**Release note**:


andyzhangx:
can you squash all commits to one? `git rebase -i HEAD~7`

andyzhangx (approved):
/lgtm
/retest

Issue #2323, opened 2024-05-07
Allow tag values containing commas and equal sign


dominiquehunziker:
Currently, it is not possible to create tags with a value that contains `,` or `=` as this breaks the parsing logic for the `tags` parameter.
``` yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: managed-premium-retain
provisioner: disk.csi.azure.com
parameters:
  skuName: Premium_ZRS
  tags: 'tag-1="aGVsbG8=",tag-2="value-2, value-3"'
reclaimPolicy: Retain
volumeBindingMode: WaitForFirstConsumer
allowVolumeExpansion: true
```

It would be great, if support for values with `,` and `=` can be added. For the previous (non functioning example) the desired result would be
``` json
{
  "tag-1": "aGVsbG8=",
  "tag-2": "value-2, value-3"
}
```
where `tag-1` could be a base64 encoded value with `=` padding or `tag-2` could be a comma separated list.

I'm uncertain what would be possible solutions as changing the parsing logic might change existing behavior. Possible options I can imagine are:
- At least for `=` the logic could be relaxed to only split on the first `=` and treat everything afterwards as the value. Similar to how query strings are parsed in URIs.
- Allow to escape `,` and `=`, i.e., `tags: 'tag-1=aGVsbG8\=,tag-2=value-2\, value-3'`
- Quote values with `,` or `=`, i.e., `tags: 'tag-1="aGVsbG8=",tag-2="value-2, value-3"'`. Similar to how data with commas in CSV is handled.
- Add another parameter which can be used to control the format of the `tags` parameter, such that a JSON or YAML string could be passed. For example
  ``` yaml
  parameters:
    tags: |
      tag-1: aGVsbG8=
      tag-2: value-2, value-3
    tagsFormat: Yaml
  ```
  or
  ``` yaml
  parameters:
    tags: |
      {
        "tag-1": "aGVsbG8=",
        "tag-2": "value-2, value-3"
      }
    tagsFormat: Json
  ```

andyzhangx:
this should be fixed by https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/2246 in v1.30.1, v1.29.5, what is your csi driver version? @dominiquehunziker

andyzhangx:
nvm, that looks like k8s yaml config parsing issue

andyzhangx:
[...]

PR #2388, opened 2024-07-02
feat: add tagValueDelimiter parameter

umagnus:
**What type of PR is this?**

/kind feature

feat: add tag value delimieter parameter and allow tag values containing commas and equal sign
referring to issue https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/2323

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -918,7 +918,7 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 	var customTags string
 	// set incremental snapshot as true by default
 	incremental := true
-	var subsID, resourceGroup, dataAccessAuthMode string
+	var subsID, resourceGroup, dataAccessAuthMode, tagValueDelimiter string
 	var err error
 	localCloud := d.cloud
 	location := d.cloud.Location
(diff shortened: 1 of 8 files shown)


pkg/util/util.go
```diff
@@ -78,27 +77,27 @@ func RoundUpSize(volumeSizeBytes int64, allocationUnitBytes int64) int64 {
 	return roundedUp
 }
 
-// ConvertTagsToMap convert the tags from string to map
+// ConvertTagsToMap convert the tags from string to map, default tagDelimiter is ","
 // the valid tags format is "key1=value1,key2=value2", which could be converted to
 // {"key1": "value1", "key2": "value2"}
-func ConvertTagsToMap(tags string) (map[string]string, error) {
+func ConvertTagsToMap(tags string, tagsDelimiter string) (map[string]string, error) {
```
andyzhangx, line 83: `tagsDelimiter` could be empty, if it's empty, then it should be `,`

pkg/azuredisk/controllerserver.go
```diff
@@ -924,6 +924,7 @@ func (d *Driver) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequ
 	location := d.cloud.Location
 
 	parameters := req.GetParameters()
+	tagValueDelimiter := ","
```
andyzhangx, line 927: just define a var tagValueDelimiter, and then if it's empty, then func `ConvertTagsToMap` would treat as `,`

pkg/util/util.go
andyzhangx: pls also add unit test for `ConvertTagsToMap`, e.g. `tagsDelimiter` is empty, `,`, `;`

andyzhangx:
@umagnus pls also squash all commits, thx

pkg/util/util.go
umagnus: done

pkg/azuredisk/controllerserver.go
umagnus: done

andyzhangx (approved):

PR #2422, opened 2024-07-19
fix: increase azuredisk container memory limits as 600Mi

### LINKED ISSUE #2416: [bug] csi azuredisk node is crashlooping when ~20 pods with PV are scheduled at the same time
**What happened**:

Csi azuredisk pod is crashlooping preventing volumes from being mounted and pod stays in `Pending` state forever.

Details in https://github.com/Azure/AKS/issues/4421

It works

**How to reproduce it**:

Schedule 20 pods with PV as the same time to one node

N/A

**Environment**:

- CSI Driver version: `mcr.microsoft.com/oss/kubernetes-csi/azuredisk-csi:v1.28.8`
- Kubernetes version (use `kubectl version`): `1.27.x`

andyzhangx:
**What type of PR is this?**
/kind bug

Fixes #2416

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


grzesuav:
@andyzhangx this is not resolved , issue is with high memory usage, not limit being too small. I mentioned that in my case memory usage was about 1.2gb

andyzhangx:
> @andyzhangx this is not resolved , issue is with high memory usage, not limit being too small. I mentioned that in my case memory usage was about 1.2gb

@grzesuav then how did you solve the issue on your cluster? increasing the memory limit to 1200Mi?

PR #2426, opened 2024-07-21
fix: use least privileged LocalService account in hostprocess deployment on Windows

andyzhangx:
**What type of PR is this?**
/kind bug

fix: use least privileged LocalService account in hostprocess deployment on Windows

https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/#choosing-a-user-account
By default, HostProcess containers support the ability to run as one of three supported Windows service accounts:

[LocalSystem](https://docs.microsoft.com/windows/win32/services/localsystem-account)
[LocalService](https://docs.microsoft.com/windows/win32/services/localservice-account)
[NetworkService](https://docs.microsoft.com/windows/win32/services/networkservice-account)
You should select an appropriate Windows service account for each HostProcess container, aiming to limit the degree of privileges so as to avoid accidental (or even malicious) damage to the host. The LocalSystem service account has the highest level of privilege of the three and should be used only if absolutely necessary. Where possible, use the LocalService service account as it is the least privileged of the three options.

Local acc

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: use least privileged LocalService account in hostprocess deployment on Windows
```


andyzhangx:
Warning  FailedMount             12s (x4 over 2m4s)  kubelet                  MountVolume.MountDevice failed for volume "pvc-dd0cf69e-4031-4b23-82c4-8b649188c059" : rpc error: code = Internal desc = could not format 5(lun: 3), and mount it at \var\lib\kubelet\plugins\kubernetes.io\csi\disk.csi.azure.com\b317d38cd57651494dfdb8b499950c6004dd04b61dd87b49630709e520c0ae5d\globalmount, failed with error mount volume to path. cmd: Get-Volume -UniqueId "$Env:volumeID" | Get-Partition | Add-PartitionAccessPath -AccessPath $Env:path, output: Add-PartitionAccessPath : Access to a CIM resource was not available to the client.\r
At line:1 char:56\r
+ ... meID" | Get-Partition | Add-PartitionAccessPath -AccessPath $Env:path\r
+                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\r
    + CategoryInfo          : PermissionDenied: (MSFT_Partition ...soft/Window...):ROOT/Microsoft/.../MSFT_Partition)  \r
   [Add-PartitionAccessPath], CimException\r
    + FullyQualifiedErrorId : MI RESULT 2,Add-PartitionAccessPath\r
 \r
, error: exit status 1

PR #2432, opened 2024-07-26
fix: liveness probe failure when hostNetwork not enabled in node (linux)

### LINKED ISSUE #2431: azuredisk node liveness failure when hostnetwork is false
**What happened**:

Since last version of azuredisk csi driver, the liveness for azuredisk node component fails when hostnetwork (`linux.hostnetwork` in chart values) is set to false.
A [fix ](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/2315) has been made but seems to target only the controller deployment.

`linux.hostnetwork` false value to be supported.

**How to reproduce it**:

Deploy azuredisk csi driver using helm chart in version 1.30.1 or 1.30.2 (probably some 1.29.x also) with a `linux.hostnetwork` false value and see that azuredisk-node daemonset pods are in crashloopback due to liveness failure.

**Environment**:

- CSI Driver version: 1.30.1 or 1.30.2 (probably some 1.29.x also)
- Kubernetes version (use `kubectl version`): 1.25.16
- OS (e.g. from /etc/os-release): ubuntu 2204

arno-pons:
**What type of PR is this?**

/kind bug

Fix liveness for node daemonset pods in case hostnework is set to false

Fixes #2431

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx:
would be fixed by https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/2521

PR #2536, opened 2024-09-24
feat: add VerticalPodAutoscaler to csi controller pod

umagnus:
**What type of PR is this?**

/kind feature

feat: add VerticalPodAutoscaler to csi controller pod

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/deploy/example/vpa/uninstall-vpa.sh
+++ b/deploy/example/vpa/uninstall-vpa.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# Copyright 2025 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set +o xtrace
+
+SCRIPT_ROOT=$(dirname ${BASH_SOURCE[0]})
+COMPONENTS="vpa-v1-crd-gen vpa-rbac updater-deployment recommender-deployment admission-controller-deployment"
+
+echo "Uninstalling VPA ..."
+
+for i in "${COMPONENTS[@]}"; do
+  if [ $i == admission-controller-deployment ] ; then
+    ${SCRIPT_ROOT}/install-components/rmcerts.sh
+    ${SCRIPT_ROOT}/install-components/delete-webhook.sh
+  fi
+  kubectl delete -f ${SCRIPT_ROOT}/install-components/${i}.yaml --ignore-not-found
+done
+bash ${SCRIPT_ROOT}/install-components/warn-obsolete-vpa-objects.sh
+
+echo 'VPA uninstalled successfully.'
--- a/deploy/example/vpa/vertical-pod-autoscaler.yaml
+++ b/deploy/example/vpa/vertical-pod-autoscaler.yaml
@@ -1,0 +1,17 @@
+---
+apiVersion: autoscaling.k8s.io/v1
+kind: VerticalPodAutoscaler
+metadata:
+  name: csi-azuredisk-controller
+  namespace: kube-system
+spec:
+  targetRef:
+    apiVersion: "apps/v1"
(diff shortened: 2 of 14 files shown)


deploy/example/vpa/vertical-pod-autoscaler.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: autoscaling.k8s.io/v1
+kind: VerticalPodAutoscaler
```
andyzhangx, line 2: pls add this yaml config verification in hack/verify-yamllint.sh script

umagnus: done

umagnus:
AKS need to add a label for vpa admission controller webhook to impact csi controller pod in `kube-system` namespace, referenced to [Can admission controller webhooks impact kube-system and internal AKS namespaces](https://learn.microsoft.com/en-us/azure/aks/faq#can-admission-controller-webhooks-impact-kube-system-and-internal-aks-namespaces-), wait for pr:https://github.com/kubernetes/autoscaler/pull/7402 to completed

voelzmo:
Hey, I came here via the reference in the [VPA PR](https://github.com/kubernetes/autoscaler/pull/7402).
I see that you're about to use VPA for all the containers in the CSI controller Pod. My understanding is that the resource usage in those containers is usually pretty low – but I don't know the details of the Azure CSI controller.

I'm not sure if you tried this for a longer period of time already in some beta env or have long-term experience with other components with many containers in a single pod, but keep in mind that this change will lead to many additional interruptions for the Pod, as VPA will evict it when the recommendation for a container changes by more than 10% (which happens quickly for these small absolute values). This also happens if e.g. your initial requests are too high and the containers will be scaled down step by step. I'm not sure if those disruptions could become an issue – but if resource usage isn't expected to change much, it may just be better to not have VPA enabled for this and have fewer disruptions.

andyzhangx:
@voelzmo thanks for the tips. we would like to use VPA mainly for CSI driver controller sidecar containers, the memory usage of those containers would grow dramatically when PV num grows in the cluster, current memory limit 500Mi is far from enough when there are a few thousand PVs in the cluster, in that case, the CSI driver would be OOM, that's the reason we want to use VPA.

andyzhangx (commented):
/ok-to-test

deploy/example/vpa/uninstall-vpa.sh
```diff
@@ -0,0 +1,29 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set +o xtrace
+
+repo="https://github.com/kubernetes/autoscaler.git"
+echo "Uninstalling VPA ..."
+git clone $repo
+cd autoscaler/vertical-pod-autoscaler
```
andyzhangx, line 24: is there an easy way to install upstream autoscaler without cloning the whole repo? and it's better installing a stable version instead of master branch

deploy/example/vpa/vertical-pod-autoscaler.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  targetRef:
+    apiVersion: "apps/v1"
+    kind: Deployment
+    name: csi-azuredisk-controller
+  resourcePolicy:
+    containerPolicies:
+      - containerName: '*'
+        maxAllowed:
+          memory: 50Gi
```
andyzhangx, line 16: 50G is crazy, 10Gi should be enough now

deploy/example/vpa/uninstall-vpa.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+# Copyright 2020 The Kubernetes Authors.
```
andyzhangx, line 2: 2025

deploy/example/vpa/uninstall-vpa.sh
umagnus: No, the doc only provides the installation to download the source code of VPA https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/docs/installation.md#install-command

andyzhangx: Note: the script currently reads environment variables: $REGISTRY and $TAG. Make sure you leave them unset unless you want to use a non-default version of VPA.

deploy/example/vpa/uninstall-vpa.sh
umagnus: done

deploy/example/vpa/vertical-pod-autoscaler.yaml
umagnus: done

PR #2569, opened 2024-10-23
fix: support attach operations after premiumV2 disk migration

landreasyan:
**What type of PR is this?**

Customers can use VAC to migrate `PremiumLRS` disks to `PremiumV2LRS` disks while they are detached from the VM. While the driver already sets caching mode to `None` when it creates a `PremiumV2LRS` disk, it still defaults to either the caching mode set in the `VolumeContext` or to `ReadOnly` caching when it attaches the disk. Because caching in the `VolumeContext` defaults and is set to `ReadOnly` for `PremiumLRS` disks, to support attach operation after the migration, the driver must set the caching mode to `None` explicitly when it attaches a `PremiumV2LRS` disk.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (approved):
/retest
/lgtm

PR #2619, opened 2024-11-11
Add a file watcher for Azure client cert auth

bryan-cox:
**What type of PR is this?**
/kind bug

Add a file watcher, utilizing fsnotify, to monitor the Azure client certificate authentication. If the file changes, the pod will need to be restarted since Azure SDK doesn't support hotloading a new certificate yet.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (commented):
/ok-to-test

pkg/azureutils/azure_disk_utils.go
```diff
@@ -245,6 +246,10 @@ func GetCloudProviderFromClient(ctx context.Context, kubeClient clientset.Interf
 			config.AADFederatedTokenFile = federatedTokenFile
 			config.UseFederatedWorkloadIdentityExtension = true
 		}
+		if len(config.AADClientCertPath) > 0 {
+			// Watch the certificate for changes; if the certificate changes, the pod will be restarted
+			err = filewatcher.WatchFileForChanges(config.AADClientCertPath)
```
andyzhangx, line 251: there is golint error here, I think you could print the error if it's not nil

bryan-cox:
@andyzhangx are the tests failing legitimately or is something else wrong? The errors don't appear to be related to the PR but I could be missing something 🤔

pkg/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,67 @@
+package filewatcher
```
andyzhangx, line 1: +// WatchFileForChanges watches the file, fileToWatch, for changes. If the file contents have changed, the pod this
+// function is running on will be restarted.
+func WatchFileForChanges(fileToWatch string) error {
Boilerplate header is wrong for: /home/prow/go/src/sigs.k8s.io/azuredisk-csi-driver/hack/../pkg/filewatcher/filewatcher.go
make: *** [Makefile:91: verify] Error 1
+ EXIT_VALUE=2
+ set +o xtrace


you need to append  header to this file

PR #2787, opened 2025-01-03
test: refine UT for gen-disk-skus-map_test

Zhupku:
This pull request introduces a new test scenario for generating disk SKU maps and includes a sample JSON file for testing purposes. The changes primarily focus on adding conditional logic to handle the test scenario and updating the test setup accordingly.

Key changes include:

### Test Scenario Implementation:
* [`pkg/tool/gen-disk-skus-map.go`](diffhunk://#diff-1029231156796ebf6e7dcab9853facf923ad1796ab8aad9eca3567ebddd1fb50R98-R110): Added conditional logic to check for the `TEST_SCENARIO` environment variable and use the `skus-sample.json` file if set. This allows for a predefined set of SKUs to be used during testing.

### Test Setup:
* [`pkg/tool/gen-disk-skus-map_test.go`](diffhunk://#diff-9494507c5160c887e22c02223e9a45d7697145b3b9e48669713086fe4064235bR25-R28): Updated the test setup to set the `TEST_SCENARIO` environment variable and ensure it is unset after the test. This ensures that the test uses the sample SKU data.

### Sample Data:
* [`pkg/tool/skus-sample.json`](diffhunk://#diff-c765cadf992eada7197b3440525789678d78a8fc35b1eea75b89563765f345edR1-R366): Added a sample JSON file containing predefined SKU data to be used during the test scenario. This file includes various SKU attributes and capabilities.

**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/tool/gen-disk-skus-map.go
+++ b/pkg/tool/gen-disk-skus-map.go
@@ -43,6 +43,7 @@ var (
 	UncachedDiskBytesPerSecondCapability = "uncacheddiskbytespersecond"
 	MaxDataDiskCountCapability           = "maxdatadiskcount"
 	VCPUsCapability                      = "vcpus"
+	SkusFullfilePath                     = "skus-full.json"
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


pkg/tool/gen-disk-skus-map.go
```diff
@@ -95,8 +95,18 @@ import (
...
+	if os.Getenv("TEST_SCENARIO") == "true" {
+		if _, err := os.Stat("skus-sample.json"); err == nil {
+			skusFullfilePath = "skus-sample.json"
+		} else {
+			if err := getAllSkus(skusFullfilePath); err != nil {
+				klog.Errorf("Could not get skus. Error: %v", err)
+			}
+		}
+	} else {
+		if err := getAllSkus(skusFullfilePath); err != nil {
```
andyzhangx, line 107: I think you could put this logic in `getAllSkus` func

Zhupku: fixed

pkg/tool/gen-disk-skus-map.go
```diff
@@ -215,13 +214,19 @@ func formatInt(value int) string {
 
 	return strconv.Itoa(value)
 }
-func getAllSkus(filePath string) (err error) {
-	klog.V(2).Infof("Getting skus and writing to %s", filePath)
+func getAllSkus(filePath *string) (err error) {
```
andyzhangx, line 217: why use a string pointer here?

Zhupku: Since the value of filePath need to be changed in the getAllSkus function, do you think I should use the return value to override it?

andyzhangx: just define `skusFullfilePath` var out of main func, and then change the `skusFullfilePath` var in ut

andyzhangx (approved):
/lgtm

PR #2859, opened 2025-02-08
fix: issue #2858 incorrect controller.affinity property processing

### LINKED ISSUE #2858: It is impossible to customize node affinity for controllers due to a bug in template
**What happened**:

Affinity specified in `controller.affinity` property in the `values.yaml` is not included in the controllers' manifests.

Both azuredisk cotroller and snapshot controller are generated with the required affinity.

**How to reproduce it**:

Run the chart
```
helm template azuredisk-csi-driver \
  --repo https://raw.githubusercontent.com/kubernetes-sigs/azuredisk-csi-driver/master/charts azuredisk-csi-driver \
  --version 1.31.2 \
  -f values.yaml
```
with the following `values.yaml`:
```
controller:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: example.io/csi
            operator: Exists

snapshot:
  enabled: true
  VolumeSnapshotClass:
    enabled: true
```
Required affinity is not generated for any of the controllers.

It seems to be due to a mistake in usage of `contains` helm template function at 
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml#L42
and
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/charts/latest/azuredisk-csi-driver/templates/csi-snapshot-controller.yaml#L45

Arguments order must be reversed or pipeline should be used,
moreover indentation in the following `toYaml` line has to be fixed as follows:
```
    {{- if tpl "{{ .Values.controller.affinity }}" . | contains "nodeSelectorTerms" }}
      {{- with .Values.controller.affinity }}
      affinity:
{{ toYaml . | indent 8 }}
      {{- end }}
    ...
```
Moreover, the use of this condition does not seem necessary, or is even actually harmful.
What if the user just wants to specify a purely pod affinity or anti-affinity rather than node affinity.

**Environment**:

- CSI Driver version: 1.31.2 and all others
- Kubernetes version (use `kubectl version`): 1.31.3
- OS (e.g. from /etc/os-release): ubuntu
- Kernel (e.g. `uname -a`): 6.8.0-52-generic #53-Ubuntu SMP PREEMPT_DYNAMIC Sat Jan 11 00:06:25 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
- Install tools: helm
- Others:

olegch:
**What type of PR is this?**

/kind bug

fix an issue with incorrect `controller.affinity` property processing.

Fixes #2858

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: issue #2858 incorrect controller.affinity property processing
```


charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
```diff
@@ -39,10 +39,10 @@ spec:
 {{ toYaml . | indent 8 }}
 {{- end }}
     # runOnControlPlane=true or runOnMaster=true only takes effect if affinity is not set
-    {{- if contains (tpl "{{ .Values.controller.affinity }}" .) "nodeSelectorTerms" }}
+    {{- if tpl "{{ .Values.controller.affinity }}" . | contains "nodeSelectorTerms" }}
```
andyzhangx, line 42: I think the logic should be 
```
if or .Values.controller.runOnControlPlane .Values.controller.runOnMaster}}

else
  use Values.controller.affinity
```

olegch: It may be better to keep it as is for backward compatibility, and because it provides logical fallback from more specific to more generic configuration.
Users can provide both controller.affinity property and controller.runOnControlPlane property, in their values.yaml, and it will work for both older version of the chart, and newer version of the chart with the bug fixed.

andyzhangx (commented):
pls run \"helm package charts/latest/azuredisk-csi-driver -d charts/latest/\" to update tgz file

andyzhangx (approved):
/lgtm

PR #2992, opened 2025-04-04
fix: detach disk should have more priority than attach disk on the same node

### LINKED ISSUE #2851: Issue with batch attach in ControllerPublishVolume
**What happened**:
2132 out of 24371 pods experienced FailedAttachVolume errors, leading to significant delays in disk attachment during load testing across multiple clusters.
The incident involved a high frequency of volume attachment errors during a scale test, where a significant number of pods failed to attach volumes due to internal errors. The errors persisted for nearly 10 minutes before some attachments succeeded, indicating a critical issue with the volume attachment process.
The primary symptom observed was a high occurrence of FailedAttachVolume errors during the testing phase.

Attach volume calls should succeed

**How to reproduce it**:

There seem to be issues with the batch attach implementation. Multiple attach requests are coming at once and the batch attach fails with 409 errors, which are essentially MaximumDataDisksExceeded errors. This happens because there is no limit on the number of disks in the attach disk batch, due to which we might exceed the total number of disks allowed to be attached on a node in a batch attach request.
For fixing this, we could have a check that keeps a track of the number of disks already attached and while creating a batch of disks to be attached,  prevent batch size from going beyond the (number of disks allowed to be attached - number of disks already attached). 

**Environment**:

- CSI Driver version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

andyzhangx:
**What type of PR is this?**
/kind bug

fix: detach disk should have more priority than attach disk on the same node, this could reduce the MaximumDataDisksExceeded error occurrence. This PR also reverts https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/2883

Fixes #2851

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: detach disk should have more priority than attach disk on the same node
```

### DIFF
--- a/pkg/azuredisk/azure_controller_common.go
+++ b/pkg/azuredisk/azure_controller_common.go
@@ -98,8 +98,6 @@ type controllerCommon struct {
 	// AttachDetachInitialDelayInMs determines initial delay in milliseconds for batch disk attach/detach
 	AttachDetachInitialDelayInMs int
 	ForceDetachBackoff           bool
-	// a timed cache for disk attach hitting max data disk count, <nodeName, "">
-	hitMaxDataDiskCountCache azcache.Resource
 }
 
 // ExtendedLocation contains additional info about the location of resources.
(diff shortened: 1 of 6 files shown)


pkg/azuredisk/azure_controller_common.go
```diff
@@ -186,25 +184,35 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
...
 		return -1, err
 	}
-	if !isMaxDataDiskCountExceeded {
-		c.lockMap.LockEntry(node)
-		defer c.lockMap.UnlockEntry(node)
 
-		if c.AttachDetachInitialDelayInMs > 0 && requestNum == 1 {
-			klog.V(2).Infof("wait %dms for more requests on node %s, current disk attach: %s", c.AttachDetachInitialDelayInMs, node, diskURI)
-			time.Sleep(time.Duration(c.AttachDetachInitialDelayInMs) * time.Millisecond)
+	var waitForDetach bool
```
nearora-msft, line 192: Considering that with this approach we are serializing the attach calls instead of batching them in case there are detach requests pending, this could:
1. Slow down the attach requests even further as we are serializing the requests, even though there might be available slots to attach multiple disks at the same time.
2. Lead to even more throttling from the CRP side as the number of attach requests will increase.

Instead of this, could we just keep track of the number of the the number of the disks already attached to the node and adjust the batch size accordingly befor making a call to the CRP? That would still allow us to batch and wouldn't aggravate the throttling problem.

andyzhangx: this PR already reverts the original disk attach serializing fix, and make disk detach happen first, that could help avoiding hitting the `MaximumDataDisksExceeded` error, and even when this happens in rare cases, this PR could also make sure the next disk detach would happen before disk detach happen in retry.

And your suggestion would also make sense to mitigate this issue in another aspect. The tricky part of such fix is that after you get `MaximumDataDisksExceeded` error, current `AttachDisk` call is done with error, need to tell the next `AttachDisk` call to adjust the batch size, while next operation could be already recovered. I will check whether the above PR would fix the issue first, thanks.

nearora-msft: Do we need to wait for the MaximumDataDisksExceeded  error to be thrown to use this fix? Can we not add this check for every attach call where the number of attach requests > 1?

andyzhangx: this PR is to avoid MaximumDataDisksExceeded error since it makes sure detach disk would happen first when there are concurrent disk attach & detach requests. Even without this PR, disk attach could be still behind disk detach since there is a node level lock there, this PR just enforces the priority, it's like a priority lock.

PR #3039, opened 2025-04-19
fix: adjust the batch size as per the number of disks allowed

nearora-msft:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azure_controller_common.go
+++ b/pkg/azuredisk/azure_controller_common.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"regexp"
 	"strings"
 	"sync"
@@ -96,6 +97,7 @@ type controllerCommon struct {
 	// AttachDetachInitialDelayInMs determines initial delay in milliseconds for batch disk attach/detach
 	AttachDetachInitialDelayInMs int
 	ForceDetachBackoff           bool
+	CheckDiskCountForBatching    bool
 }
 
 // ExtendedLocation contains additional info about the location of resources.
@@ -196,13 +198,24 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 		time.Sleep(time.Duration(c.AttachDetachInitialDelayInMs) * time.Millisecond)
 	}
 
-	_, instanceType, err := getNodeInfoFromLabels(ctx, string(nodeName), c.cloud.KubeClient)
-	if err != nil {
-		return -1, err
+	numDisksAllowed := math.MaxInt
+	if c.CheckDiskCountForBatching {
+		_, instanceType, err := getNodeInfoFromLabels(ctx, string(nodeName), c.cloud.KubeClient)
+		if err != nil {
+			return -1, err
+		}
+		if instanceType != "" {
+			maxNumDisks, instanceExists := getMaxDataDiskCount(instanceType)
+			if instanceExists {
+				attachedDisks, _, err := c.GetNodeDataDisks(ctx, nodeName, azcache.CacheReadTypeDefault)
+				if err != nil {
+					return -1, err
+				}
+				numDisksAttached := len(attachedDisks)
+				numDisksAllowed = int(maxNumDisks) - numDisksAttached
+			}
+		}
 	}
-	maxNumDisks := getMaxDataDiskCount(instanceType)
-	numDisksAttached := len(occupiedLuns)
-	numDisksAllowed := int(maxNumDisks) - numDisksAttached
 
 	diskMap, err := c.cleanAttachDiskRequests(node)
 	if err != nil {
@@ -212,12 +225,13 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 	// Remove some disks from the batch if the number is more than the max number of disks allowed
 	removeDisks := len(diskMap) - numDisksAllowed
 	if removeDisks > 0 {
-		klog.V(2).Infof("azureDisk - too many disks to attach, remove %d disks from the request", removeDisks)
+		klog.V(2).Infof("too many disks to attach, remove %d disks from the request", removeDisks)
 		for diskURI, options := range diskMap {
 			if removeDisks == 0 {
 				break
 			}
 			if options != nil {
+				klog.V(2).Infof("remove disk(%s) from attach request", diskURI)
 				delete(diskMap, diskURI)
 			}
 			removeDisks--
--- a/pkg/azuredisk/azure_controller_common.go
+++ b/pkg/azuredisk/azure_controller_common.go
@@ -29,7 +29,6 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v6"
 
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/wait"
 	kwait "k8s.io/apimachinery/pkg/util/wait"
 	cloudprovider "k8s.io/cloud-provider"
 	volerr "k8s.io/cloud-provider/volume/errors"
@@ -72,7 +71,7 @@ const (
 )
 
 var defaultBackOff = kwait.Backoff{
-	Steps:    7,
+	Steps:    20,
 	Duration: 2 * time.Second,
 	Factor:   1.5,
 	Jitter:   0.0,
@@ -97,7 +96,6 @@ type controllerCommon struct {
 	// AttachDetachInitialDelayInMs determines initial delay in milliseconds for batch disk attach/detach
 	AttachDetachInitialDelayInMs int
 	ForceDetachBackoff           bool
-	WaitForDetach                bool
 }
 
 // ExtendedLocation contains additional info about the location of resources.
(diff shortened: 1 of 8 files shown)


pkg/azuredisk/azure_controller_common.go
```diff
@@ -190,38 +188,42 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
...
 	c.lockMap.LockEntry(node)
 	defer c.lockMap.UnlockEntry(node)
 
-	if !waitForDetachHappened && c.AttachDetachInitialDelayInMs > 0 && requestNum == 1 {
+	if c.AttachDetachInitialDelayInMs > 0 && requestNum == 1 {
 		klog.V(2).Infof("wait %dms for more requests on node %s, current disk attach: %s", c.AttachDetachInitialDelayInMs, node, diskURI)
 		time.Sleep(time.Duration(c.AttachDetachInitialDelayInMs) * time.Millisecond)
 	}
 
+	_, instanceType, err := getNodeInfoFromLabels(ctx, string(nodeName), c.cloud.KubeClient)
```
nearora-msft, line 199: @andyzhangx  do you see any issues with this simple implementation to adjust the batch size as needed?

pkg/azuredisk/azure_controller_common.go
```diff
@@ -190,38 +188,42 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
...
+	// Remove some disks from the batch if the number is more than the max number of disks allowed
+	removeDisks := len(diskMap) - numDisksAllowed
+	if removeDisks > 0 {
+		klog.V(2).Infof("azureDisk - too many disks to attach, remove %d disks from the request", removeDisks)
+		for diskURI, options := range diskMap {
+			if removeDisks == 0 {
+				break
+			}
+			if options != nil {
+				delete(diskMap, diskURI)
```
andyzhangx, line 221: pls also logging the removed disk URI

andyzhangx (commented):
thanks for the PR, I have left some comments.

pkg/azuredisk/azure_controller_common.go
andyzhangx: - pls keep original logic about making disk detach as higher priority than disk attach, from the logs, I could see it already has mitigated the issue a lot since even it could reduce the `MaximumDataDisksExceeded` error num, and even when this error happens, it could recover very soon. Your PR would eliminate this error but use would hit disk attach error for the removed disk attach in the end.
 - `getNodeInfoFromLabels` only works on AKS, may not work on other clusters, so if there is error or `instanceType` is empty, then skip this adjustment
 - `getMaxDataDiskCount` depends on the static max disk account table which is also not reliable, if instanceType is not in maxDataDiskCountMap[vmsize], then should also skip this adjustment
 - pls provide a feature flag (enabled by default), then we could disable this adjustment with this flag in csi controller pod

andyzhangx:
btw, the original Elastic issue is mainly contributed by retry-after value returned by CRP during polling, this `MaximumDataDisksExceeded` error still happens some times, and from the CRP logs, I could see the next `VMScaleSetVMs.VMScaleSetVMsOperation.PUT` would succeed very soon in a few seconds, that means it could recover very soon.

On the other hand, I think this PR is still valuable which could make most disk attach succeed in the beginning. Thanks for the fix.

pkg/azuredisk/azure_controller_common.go
nearora-msft: Thanks for your comments @andyzhangx. 
Even if the user gets the disk attach error for the removed disk, it will be retried by the CSI driver, right? And it would succeed in the next few calls? 
Do we still need waitFordetach changes? Don't they add to the overall attach latency?

andyzhangx: > Even if the user gets the disk attach error for the removed disk, it will be retried by the CSI driver, right? And it would succeed in the next few calls?

from my previous observation, sometimes the remove disk attach would be stuck for long time since disk detach does not happen immediately (preempted by other disk attach). that would add to the attach latency of remove disks.
I think in this PR, you could still keep the waitFordetach changes, after your PR merged, I could refine the original waitFordetach logic, make it only happen when there is MaximumDataDisksExceeded error(make sure in the next few disk attach/detach cycles, detach would have higher priority than attach), I think that could have the least impact.

pkg/azuredisk/azure_controller_common.go
```diff
@@ -72,7 +72,7 @@ const (
 )
 
 var defaultBackOff = kwait.Backoff{
-	Steps:    7,
+	Steps:    20,
```
edreed, line 75: This will cause us to retry from almost 2 hours. Should we be retrying that long?

pkg/azuredisk/azure_controller_common.go
```diff
@@ -190,38 +190,54 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
...
+		}
+		if instanceType != "" {
+			maxNumDisks, instanceExists := getMaxDataDiskCount(instanceType)
+			if instanceExists {
+				attachedDisks, _, err := c.GetNodeDataDisks(ctx, nodeName, azcache.CacheReadTypeDefault)
+				if err != nil {
+					return -1, err
+				}
+				numDisksAttached := len(attachedDisks)
+				numDisksAllowed = int(maxNumDisks) - numDisksAttached
```
edreed, line 215: If `numDisksAttached` > `maxNumDisks`, `numDisksAllowed` would go negative. We should be safe here and set `numDisksAllowed` to zero in this case.

```go
numDisksAllowed = int(maxNumDisks) - math.MaxInt(maxNumDisks, numDisksAttached)
```

pkg/azuredisk/azure_controller_common.go
```diff
@@ -190,38 +190,54 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
...
+	removeDisks := len(diskMap) - numDisksAllowed
+	if removeDisks > 0 {
+		klog.V(2).Infof("too many disks to attach, remove %d disks from the request", removeDisks)
+		for diskURI, options := range diskMap {
+			if removeDisks == 0 {
+				break
+			}
+			if options != nil {
+				klog.V(2).Infof("remove disk(%s) from attach request", diskURI)
+				delete(diskMap, diskURI)
```
edreed, line 235: If a disk is removed from the map, the error that will be returned is "disk(<pvc>) could not be found on node" which is confusing. We should emit a clearer error message, e.g. "disk <pvc> could not be attached to node <node> because it would exceed the maximum number of disks".

pkg/azuredisk/azure_controller_common.go
andyzhangx: > While with the changes in this PR, I haven't seen that error occurring.

@nearora-msft 
Your PR would eliminate this MaxDataDisksExceeded error but user would hit different disk attach error for the removed disk attach in the end. When there is MaxDataDisksExceeded error, we should make sure in the next few disk attach/detach cycles, detach would have higher priority than attach.

Pls keep the original waitForDetach changes, after your PR merged, I will refine the logic to make it happen **only** when there is MaximumDataDisksExceeded error, I think that could have the least impact.

pkg/azuredisk/azure_controller_common.go
andyzhangx: this is actually not related to this fix, there is a timeout set in the outside which would timeout sooner before 20th backoff, and set as 20 steps does not make sense though it won't happen in the beginning.

pkg/azuredisk/azure_controller_common.go
andyzhangx: this kind of clear error msg may be difficult to set since in this AttachDisk operation, it would return success since other disks are removed from the list.

pkg/azuredisk/azure_controller_common.go
```diff
@@ -190,38 +190,54 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
...
+	if c.AttachDetachInitialDelayInMs > 0 && requestNum == 1 {
 		klog.V(2).Infof("wait %dms for more requests on node %s, current disk attach: %s", c.AttachDetachInitialDelayInMs, node, diskURI)
 		time.Sleep(time.Duration(c.AttachDetachInitialDelayInMs) * time.Millisecond)
 	}
 
+	numDisksAllowed := math.MaxInt
+	if c.CheckDiskCountForBatching {
+		_, instanceType, err := getNodeInfoFromLabels(ctx, string(nodeName), c.cloud.KubeClient)
+		if err != nil {
+			return -1, err
```
andyzhangx, line 205: if there is error, just error log here, and exit CheckDiskCountForBatching since `getNodeInfoFromLabels` may fail in many ways, e.g. lable does not exist or kubeclient may not have permission, etc.

pkg/azuredisk/azure_controller_common.go
nearora-msft: As per the logs, this the error that would be seen for the removed disk:

GRPC error: rpc error: code = Internal desc = Attach volume /subscriptions/d64ddb0c-7399-4529-a2b6-037b33265372/resourceGroups/MC_nearora-rg_disk-csi-driver-testing_westus2/providers/Microsoft.Compute/disks/pvc-f0ea4a65-df82-4ef2-ae06-a18cc4030995 to instance aks-storagepool-30416368-vmss000002 failed with could not find disk(/subscriptions/d64ddb0c-7399-4529-a2b6-037b33265372/resourcegroups/mc_nearora-rg_disk-csi-driver-testing_westus2/providers/microsoft.compute/disks/pvc-f0ea4a65-df82-4ef2-ae06-a18cc4030995) in current disk list(len: 0) nor in diskMap(map[])

The attach gets retried by AKS and eventually succeeds.

Is waiting for detach going to ensure faster attach? Sorry, just trying to understand the reason that we need to keep both the changes.

pkg/azuredisk/azure_controller_common.go
nearora-msft: This came in when I reverted Andy's change for waitForDetach commit. @andyzhangx , what should be the correct retry steps here?

andyzhangx: original value is more reasonable, you could just revert the revert PR.

pkg/azuredisk/azure_controller_common.go
andyzhangx: yes, in this case, the driver should wait for detach complete first, otherwise the remaining disk attach would be stuck there for more time (that also depends), pls revert the revert commit, and I will make corresponding changes after this PR merge. thanks.

nearora-msft: Reverted

pkg/azuredisk/azure_controller_common.go
nearora-msft: The disk that is removed may have come from a different attach request, so we'd need to find a way to let that request know that the disk from another batch attach request has dropped the disk. 

One way to do that would be to cache all the removed disks in a map and if we find the disk from the current request in that map, we return the appropriate error. But that would require maintaining another map and using locks for concurrency control. Please lmk if you have any better ideas.

andyzhangx: that's too complex and error prone, I think we could just keep current error msg if there is no better way, it's nice to have anyway.

pkg/azuredisk/nodeserver.go
```diff
@@ -444,16 +445,16 @@ func (d *Driver) NodeGetInfo(ctx context.Context, _ *csi.NodeGetInfoRequest) (*c
 	}, nil
 }
 
-func getMaxDataDiskCount(instanceType string) int64 {
+func getMaxDataDiskCount(instanceType string) (int64, bool) {
```
andyzhangx, line 448: can you add ut for returned bool value in another PR? thx

andyzhangx (approved):
/lgtm

nearora-msft:
> /cherrypick release-1.32

Thanks for creating cherry-pick to releases @andyzhangx, when would these be available on customer clusters?

pkg/azuredisk/nodeserver.go
nearora-msft: Added : https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3080

PR #3071, opened 2025-05-06
fix: bump mount-utils to fix volume resizing failure when there are nfs unresponsive volumes

gnufied:
This fixes an issue with azure-disk volume resizing during mounting if worker node also has nfs volumes mounted and nfs mount points are unresponsive.

See - https://github.com/kubernetes/kubernetes/pull/131568 for more information.


andyzhangx (approved):
/retest
/lgtm

gnufied:
@andyzhangx lets wait for 0.33 release before we cherry-pick.

PR #3112, opened 2025-05-27
doc: add linked permissions

jsafrane:
**What type of PR is this?**
/kind documentation

I admit I don't fully understand Azure permissions. I think that when the CSI driver patches a VM to attach a disk, Azure evaluates *all* fields of the VM and not only the changed ones. Azure then evaluates all "link"/"action" permissions of anything that is linked to the VM - load balancers, security groups, network interfaces etc, even though the CSI driver never modifies these VM properties.

I'm adding the permissions found in an internal audit. However, I don't know _what else_ can be linked to a VM. The list may be quite incomplete. And it's not very future proof, as Azure may link other objects to the VMs in the future and we will need to update the driver permissions. It would be great if the permissions could say that the driver is allowed to modify only the disks attached to the VM + needs permissions for them, ignoring all other linked objects.

**Requirements**:
  
- [x] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (approved):
/lgtm

PR #3156, opened 2025-06-26
fix: removing crc from mount option for xfs.

pawanpraka1:
**What type of PR is this?**
/kind bug
/kind regression

AzureLinuxV3 does not support crc option and causing the failure while mounting the device.
    
Dmesg error :
Deprecated V4 format (crc=0) not supported by kernel.

https://portal.microsofticm.com/imp/v5/incidents/details/647022560

Fixes #

https://portal.microsofticm.com/imp/v5/incidents/details/647022560

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

https://github.com/microsoft/azurelinux/commit/a1cef2314400f4f22926386d25f26244cf7dc388

### DIFF
--- a/pkg/azurediskplugin/Dockerfile
+++ b/pkg/azurediskplugin/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.20.6
+FROM alpine:3.18.12
 RUN apk upgrade --available --no-cache && \
     apk add --no-cache util-linux e2fsprogs e2fsprogs-extra ca-certificates udev xfsprogs xfsprogs-extra btrfs-progs btrfs-progs-extra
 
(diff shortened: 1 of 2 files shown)


pkg/azurediskplugin/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM alpine:3.20.6
+FROM alpine:3.18.12
```
andyzhangx, line 15: /hold
`alpine:3.18.12` is already EOL now, that's the reason we want to upgrade to 3.20

andyzhangx: let's use this PR to fix the issue: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3160

andyzhangx:
addressed by https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3160, feel free to propose new PR for long term fix, thanks.

PR #3162, opened 2025-06-30
chore: avoid k8s retry for successful attach/detach

hccheng72:
**What type of PR is this?**
/kind bug

This issue originated from a user scenario where both LRS and ZRS disks were being detached in the same batch. When the zone was down, the detach operation for ZRS disks could still succeed, but the LRS disks would cause the batch request to fail. As a result, CSI incorrectly assumed that no disks were successfully detached. To avoid repeated retries leading to increasing backoff delays, we now check the attach/detach status before returning an error.

no retry if the attach/detach batch fails but the attach/detach for the disk is successful.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azure_controller_common.go
+++ b/pkg/azuredisk/azure_controller_common.go
@@ -302,7 +302,14 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 			err = vmset.UpdateVM(ctx, nodeName)
 		}
 		if err != nil {
-			return -1, err
+			// re-check if attach fails, some other disks might fail in the batch but the attach for the disk could be successful
+			// to avoid delay for the k8s retry
+			diskLun, _, errGetLun := c.GetDiskLun(ctx, diskName, diskURI, nodeName)
+			if errGetLun != nil {
+				return -1, err
+			}
+			klog.Errorf("attaching batch to node %s failed, diskMap len:%d, %+v, error: %v", nodeName, len(diskMap), diskMap, err)
+			return diskLun, nil
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/azuredisk/azure_controller_common.go
```diff
@@ -302,7 +302,14 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 			err = vmset.UpdateVM(ctx, nodeName)
 		}
 		if err != nil {
-			return -1, err
+			// re-check if attach fails, some other disks might fail in the batch but the attach for the disk could be successful
+			// to avoid delay for the k8s retry
+			diskLun, _, errGetLun := c.GetDiskLun(ctx, diskName, diskURI, nodeName)
+			if errGetLun != nil {
+				return -1, err
```
nearora-msft, line 309: The error message could be more descriptive here, like that on line 320.

nearora-msft: Do we still need the lun check starting from line 316?

pkg/azuredisk/azure_controller_common.go
```diff
@@ -302,7 +302,14 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 			err = vmset.UpdateVM(ctx, nodeName)
 		}
 		if err != nil {
-			return -1, err
+			// re-check if attach fails, some other disks might fail in the batch but the attach for the disk could be successful
```
andyzhangx, line 305: can you only check when there is no throttling error?

andyzhangx, line 305: that means disk attach fails partially in one batch? how is that possible? if that's very edge case, I would suggest keep original logic

hccheng72: @andyzhangx I added the user's case from an IcM into the description.

andyzhangx: thanks, that looks like edge case, I would suggest keeping original logic otherwise this PR would introduce extra get vmss call in every attach/detach failure, waiting for retry should be ok in such edge case.

hccheng72: @landreasyan What do you think?

I also observed another case where the calling disk isn’t included in the diskMap, so the returned error isn’t actually related to that disk. I believe this scenario is more common.

andyzhangx: that's a batch operation which means all disks in that batch attach would fail unless it's LRS+ZRS combo and in zone outage scenario, for other cases, just wait for disk attach retry should be reasonable since there is already disk attach failure on the same node. and even in LRS+ZRS combo and in zone outage scenario, I think wait for one more disk attach retry should be also ok since finally attached disk would be considered as attached.

My major concern is that this PR would introduce extra get vmss call in every attach/detach failure, e.g. when there is throttling or other errors, it would make extra get vmss list call, that would lead to more throttling as long as there is disk attach failure (disk attach failure is very common)

/hold

PR #3178, opened 2025-07-03
support custom Azure disk naming using PVC name and namespace [#3174]

### LINKED ISSUE #3174: Support Custom Azure Disk Naming Using PVC Name and Namespace in Dynamic Provisioning
&nbsp;
In many Kubernetes environments, particularly in multi-tenant or enterprise clusters, it's important to **track Azure disks** back to the originating **PersistentVolumeClaim (PVC)** and **namespace**. Currently, the Azure Disk CSI driver generates disk names based on PVC UUIDs, which makes it difficult to identify resources by workload.

This is especially painful when:
- Trying to map Azure resources back to Kubernetes without cross-referencing volume attachments or PVC annotations
- Analyzing cost reports in Azure without meaningful disk names
- Managing lifecycle of disks across namespaces or teams

We’d like the Azure Disk CSI driver to **support custom disk naming patterns** when dynamically provisioning disks, such as:

 `Namespace name + PVC Name`  Or `pvcNamespace + pvcName`

This could be enabled via a new **optional StorageClass parameter**, such as:

```yaml
parameters:
  useCustomDiskNameFormat: "true"
  diskNameFormat: "pvc-$(NAMESPACE)-$(PVC_NAME)"
```

The driver should:

- Interpolate well-known variables like `$(PVC_NAME)`, `$(NAMESPACE)`
- Truncate the name if it exceeds Azure’s 80-character limit
- Sanitize invalid characters
&nbsp;

- **Using a webhook or static PVs**: This approach allows control over disk names, but introduces significant complexity. It requires pre-provisioning Azure disks manually or through automation, and bypasses native CSI dynamic provisioning features like topology awareness, access modes, and automated cleanup.

- **Relying on Azure resource tags**: While we can tag disks with `pvcName` and `namespace`, the **disk name itself remains a UUID-based string**. This makes it difficult to quickly identify which Kubernetes workload a disk belongs to when viewing resources in the Azure Portal, CLI, or cost reports.

- **Post-provision renaming outside CSI driver**: Azure doesn’t allow renaming managed disks after creation, so there's no way to update names after provisioning. This limits any external workaround after the volume is created.

**Additional context**

This feature would significantly improve operational visibility and manageability for teams working in large or multi-tenant clusters.

Currently, identifying Azure disks based on dynamically provisioned volumes requires either:
- Deep inspection of Kubernetes PVCs and PVs
- Manual cross-referencing of disk URIs
- Custom tooling for mapping PVC metadata to Azure resources

Having meaningful, human-readable disk names (e.g., `pvc-<namespace>-<name>`) directly at provisioning time would:
- Simplify cost tracking and resource auditing in Azure
- Improve troubleshooting workflows (especially in incident response)
- Help infrastructure teams associate cloud resources with workloads faster

This capability aligns with common naming practices used by cloud teams and has been requested in other CSI implementations (e.g., [AWS EBS CSI #996](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/996)).

We believe adding optional support for customizable disk naming—using standard PVC metadata and a safe format—would benefit many users and could be implemented without disrupting default behaviors.

yugannkt:
**What type of PR is this?**

/kind feature
/kind documentation

This PR introduces a custom disk naming logic to the Azure Disk CSI driver:

- When both the PVC namespace and name tags are present and non-empty, the Azure disk will be named as `<namespace>-<pvcname>`.
- If either tag is missing, the default disk naming logic is used.
- The disk name is always sanitized to meet Azure requirements.

This improves traceability and operational clarity by making disk names more meaningful and directly linked to their originating PVCs.

Fixes #3174

**Requirements**:
- [x] includes documentation
- [x] adds unit tests

- The logic is implemented in the `CreateVolume` method in `controllerserver.go`.
- Example: If a PVC has namespace `prod` and name `data`, the resulting disk name will be `prod-data`.

**Release note**:
```
```release-note
Added custom disk naming logic: When both PVC namespace and name tags are present, Azure disks are named as <namespace>-<pvcname>. This improves traceability and operational clarity. If either tag is missing, the default naming logic is used. The disk name is always sanitized to meet Azure requirements. (#3174)
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -153,13 +153,26 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 
 	if diskParams.DiskName == "" {
-		// Custom PVC naming logic: use pvcNamespace-pvcName if both tags are present, else fallback to default name
 		pvcName, pvcNameOk := diskParams.Tags[consts.PvcNameTag]
 		pvcNamespace, pvcNamespaceOk := diskParams.Tags[consts.PvcNamespaceTag]
-		if pvcNameOk && pvcNamespaceOk && pvcName != "" && pvcNamespace != "" {
-			diskParams.DiskName = pvcNamespace + "-" + pvcName
+		if val, ok := params["diskName"]; ok && val != "" {
+			// Template substitution for ${pvc.metadata.namespace} and ${pvc.metadata.name}
+			diskName := val
+			if pvcNamespaceOk {
+				diskName = strings.ReplaceAll(diskName, "${pvc.metadata.namespace}", pvcNamespace)
+			}
+			if pvcNameOk {
+				diskName = strings.ReplaceAll(diskName, "${pvc.metadata.name}", pvcName)
+			}
+			if diskName == "-" || diskName == "" {
+				return nil, status.Error(codes.InvalidArgument, "CreateVolume diskName template must resolve to a non-empty value")
+			}
+			diskParams.DiskName = diskName
+			klog.V(2).Infof("Disk name (template) will be: %s, the name length is %d.", diskName, len(diskName))
 		} else {
+			// Fallback to default name (original upstream behavior)
 			diskParams.DiskName = name
+			klog.V(2).Infof("Disk name (default) will be: %s, the name length is %d.", name, len(name))
 		}
 	}
 	diskParams.DiskName = azureutils.CreateValidDiskName(diskParams.DiskName)
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -153,7 +153,14 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 
 	if diskParams.DiskName == "" {
-		diskParams.DiskName = name
+		// Custom PVC naming logic: use pvcNamespace-pvcName if both tags are present, else fallback to default name
+		pvcName, pvcNameOk := diskParams.Tags[consts.PvcNameTag]
+		pvcNamespace, pvcNamespaceOk := diskParams.Tags[consts.PvcNamespaceTag]
+		if pvcNameOk && pvcNamespaceOk && pvcName != "" && pvcNamespace != "" {
+			diskParams.DiskName = pvcNamespace + "-" + pvcName
+		} else {
+			diskParams.DiskName = name
+		}
 	}
 	diskParams.DiskName = azureutils.CreateValidDiskName(diskParams.DiskName)
 


pkg/azuredisk/controllerserver.go
```diff
@@ -153,7 +153,14 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 
 	if diskParams.DiskName == "" {
-		diskParams.DiskName = name
+		// Custom PVC naming logic: use pvcNamespace-pvcName if both tags are present, else fallback to default name
+		pvcName, pvcNameOk := diskParams.Tags[consts.PvcNameTag]
```
andyzhangx, line 156: these two tags always exist, that would change the disk name by default, e.g.
```
I0703 13:21:50.354802       1 utils.go:105] GRPC call: /csi.v1.Controller/CreateVolume
I0703 13:21:50.354824       1 utils.go:106] GRPC request: {"accessibility_requirements":{"preferred":[{"segments":{"topology.disk.csi.azure.com/zone":""}}],"requisite":[{"segments":{"topology.disk.csi.azure.com/zone":""}}]},"capacity_range":{"required_bytes":549755813888},"name":"pvc-769b844f-9f08-4a97-ac04-54a1b3cb4cf0","parameters":{"csi.storage.k8s.io/pv/name":"pvc-769b844f-9f08-4a97-ac04-54a1b3cb4cf0","csi.storage.k8s.io/pvc/name":"pvc-qv7vf","csi.storage.k8s.io/pvc/namespace":"azuredisk-5903","skuName":"Standard_LRS"},"volume_capabilities":[{"AccessType":{"Mount":{"mount_flags":["barrier=1","acl"]}},"access_mode":{"mode":7}}]}
```

andyzhangx: what about adding a new parameter, e.g. `diskName`, if you set as `diskName: ${pvc.metadata.namespace}-${pvc.metadata.name}`, then it would have the same behavior as your PR, and that would be more flexible, e.g. if you want to add extra prefix or suffix, could you just change the `diskName` value, I have implemented similar functionality in smb csi driver: https://github.com/kubernetes-csi/csi-driver-smb/blob/master/docs/driver-parameters.md#subdir-parameter-supports-following-pvpvc-metadata-conversion, I could port the change to this driver if you agree with this solution.

yugannkt: Thank you for the suggestion!

I have updated the implementation to support an opt-in diskName parameter, as you recommended. Now, if diskName is set in the StorageClass parameters, it is used as a template and supports variable substitution for ${pvc.metadata.namespace} and ${pvc.metadata.name}. This allows users to customize disk names flexibly, including adding prefixes or suffixes as needed.

If diskName is not set, the driver falls back to the default CSI-generated name, so the change is fully backward compatible.

Example usage:
 ```
parameters:
 diskName: "${pvc.metadata.namespace}-${pvc.metadata.name}"
```

andyzhangx:
addressed by https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3189

PR #3196, opened 2025-07-07
fix: use host binary for xfs

pawanpraka1:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/charts/latest/azuredisk-csi-driver/templates/node-configmap.yaml
+++ b/charts/latest/azuredisk-csi-driver/templates/node-configmap.yaml
@@ -1,0 +1,14 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: host-xfs
+data:
+  xfs: |
+    #!/bin/sh
+    if [ -x /host/sbin/mkfs.xfs ]; then
+      chroot /host /sbin/mkfs.xfs "$@"
+    elif [ -x /host/usr/sbin/mkfs.xfs ]; then
+      chroot /host /usr/sbin/mkfs.xfs "$@"
+    else
+      chroot /host mkfs.xfs "$@"
+    fi
--- a/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node.yaml
+++ b/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node.yaml
@@ -210,6 +210,10 @@ spec:
               name: mountpoint-dir
             - mountPath: /etc/kubernetes/
               name: azure-cred
+            - mountPath: /host
+              name: host-dir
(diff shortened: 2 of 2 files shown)


charts/latest/azuredisk-csi-driver/templates/node-configmap.yaml
```diff
@@ -0,0 +1,14 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: host-xfs
+data:
+  xfs: |
+    #!/bin/sh
+    if [ -x /host/sbin/mkfs.xfs ]; then
```
andyzhangx, line 8: Thanks for the fix, this open source disk driver supports multiple OS, e.g. Ubuntu, Azure Linux, CoreOS, it even works with SUSE, this PR may break on other OS, since we have already used debian-base image which uses a relatively old version of mkfs.xfs (compared to version used in former base image alpine:3.20), that should already fix the issue.

pawanpraka1: >  this PR may break on other OS

Could you elaborate where it will not work?

andyzhangx: I don't know where mkfs.xfs exists in CoreOS, SUSE and other distro, this **may** break in those distro, and also mounting with `/` (root dir) would introduce a security hole, it could be probably challenged by a few customers.

pawanpraka1: > I don't know where mkfs.xfs exists in CoreOS, SUSE and other distro, this may break in those distro

There is a way we can install xfsprogs on those distros.

> also mounting with / (root dir) would introduce a security hole, it could be probably challenged by a few customers.

There are opensource repositories where host filesystem is mounted to solve the same issue. It is a generic solution for this kind of problem where driver mounts the host filesystem as readonly and uses the host binaries.

>  we have already used debian-base image which uses a relatively old version of mkfs.xfs

When we update the image and there is some compatibility issue in future, we will hit the issue again.

pawanpraka1: >also mounting with / (root dir) would introduce a security hole, it could be probably challenged by a few customers.

Another way we can solve this is by mounting the mkfs.xfs and dependent libraries inside the node daemonset pod and we can use the mkfs.xfs binary inside the pod and LD_PRELOAD the libraries, this way we should be able to use host binaries inside the daemonset pod without mounting the whole root filesystem.

PR #3227, opened 2025-07-25
feat: notify user using events on sku migration complete

hasethuraman:
**What type of PR is this?**
/kind feature

In Azure when user modifies the PersistentVolume sku from Premium_LRS to PremiumV2_LRS, it takes time to change the effect though the SKU shows PremiumV2_LRS immediately in portal. The completionPercentage in disk properties indicate same. Though this window does not stop the user from attaching the PV but may see some performance drop. So, users may want to know how long this will take and may plan the maintenance window in a planned manner that would avoid any unforeseen/unnecessary disruption. Example, in order to reduce any impact or due to any restrictions, user wanted to plan to batch the volumes (> x10s of PVs) for using PremiumV2_LRS. Without this PR, users may to use other way to monitor the completionPercentage outside AKS. With this, user can wait for the events on PersitentVolume for completion event and batch next set of volumes.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
+++ b/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
@@ -76,7 +76,11 @@ spec:
           image: "{{ .Values.image.csiProvisioner.repository }}:{{ .Values.image.csiProvisioner.tag }}"
 {{- end }}
           args:
+{{- if .Values.feature.modifyVolume }}
+            - "--feature-gates=Topology=true,HonorPVReclaimPolicy=true,VolumeAttributesClass=true"
+{{- else }}
             - "--feature-gates=Topology=true,HonorPVReclaimPolicy=true"
+{{- end }}
             - "--csi-address=$(ADDRESS)"
             - "--v=2"
             - "--timeout=30s"
@@ -163,7 +167,11 @@ spec:
             - "-leader-election"
             - "--leader-election-namespace={{ .Release.Namespace }}"
             - '-handle-volume-inuse-error=false'
-            - '-feature-gates=RecoverVolumeExpansionFailure=true'
+{{- if .Values.feature.modifyVolume }}
+            - "-feature-gates=RecoverVolumeExpansionFailure=true,VolumeAttributesClass=true"
+{{- else }}
+            - "-feature-gates=RecoverVolumeExpansionFailure=true"
+{{- end }}
             - "-timeout=240s"
             - "--retry-interval-max=30m"
           env:
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -40,6 +40,9 @@ import (
 	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/volume/util/hostutil"
 	"k8s.io/mount-utils"
@@ -96,6 +99,8 @@ type Driver struct {
 	cloud                        *azure.Cloud
 	clientFactory                azclient.ClientFactory
 	diskController               *ManagedDiskController
+	eventRecorder                record.EventRecorder
+	migrationMonitor             *MigrationProgressMonitor
 	mounter                      *mount.SafeFormatAndMount
 	deviceHelper                 optimization.Interface
 	nodeInfo                     *optimization.NodeInfo
@@ -263,6 +268,27 @@ func NewDriver(options *DriverOptions) *Driver {
 		driver.diskController.ForceDetachBackoff = driver.forceDetachBackoff
 		driver.diskController.WaitForDetach = driver.waitForDetach
 		driver.diskController.CheckDiskCountForBatching = driver.checkDiskCountForBatching
+
+		if kubeClient != nil {
+			eventBroadcaster := record.NewBroadcaster()
+			eventBroadcaster.StartStructuredLogging(0)
+			eventBroadcaster.StartRecordingToSink(&clientcorev1.EventSinkImpl{
+				Interface: kubeClient.CoreV1().Events(""),
+			})
+			eventRecorder := eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{
+				Component: driver.Name,
+			})
+			driver.eventRecorder = eventRecorder
+			driver.migrationMonitor = NewMigrationProgressMonitor(kubeClient, eventRecorder, driver.diskController)
+
+			// Recover any ongoing migrations after restart
+			go func() {
+				time.Sleep(30 * time.Second) // Wait for controller to fully start
+				if err := driver.recoverMigrationMonitorsFromAnnotations(context.Background()); err != nil {
+					klog.Errorf("Failed to recover migration monitors: %v", err)
+				}
+			}()
+		}
 	}
 
 	driver.deviceHelper = optimization.NewSafeDeviceHelper()
@@ -356,6 +382,12 @@ func (d *Driver) Run(ctx context.Context) error {
 	go func() {
 		//graceful shutdown
 		<-ctx.Done()
+
+		// Stop migration monitor if it exists
+		if d.migrationMonitor != nil {
+			d.migrationMonitor.Stop()
+		}
+
 		s.GracefulStop()
 	}()
 	// Driver d act as IdentityServer, ControllerServer and NodeServer
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -424,6 +424,19 @@ func (d *Driver) ControllerModifyVolume(ctx context.Context, req *csi.Controller
 		skuName = ""
 	}
 
+	// Get current disk state before modification
+	currentDisk, err := d.checkDiskExists(ctx, diskURI)
+	if err != nil {
+		return nil, status.Error(codes.NotFound, fmt.Sprintf("Volume not found, failed with error: %v", err))
+	}
+	// Check if this is a SKU migration
+	var fromSKU armcompute.DiskStorageAccountTypes
+	var monitorSKUMigration bool
+	if currentDisk != nil && currentDisk.SKU != nil && currentDisk.SKU.Name != nil {
+		fromSKU = *currentDisk.SKU.Name
+		monitorSKUMigration = skuName == armcompute.DiskStorageAccountTypesPremiumV2LRS && fromSKU == armcompute.DiskStorageAccountTypesPremiumLRS
+	}
+
 	klog.V(2).Infof("begin to modify azure disk(%s) account type(%s) rg(%s) location(%s)",
 		diskParams.DiskName, skuName, diskParams.ResourceGroup, diskParams.Location)
 
@@ -451,6 +464,27 @@ func (d *Driver) ControllerModifyVolume(ctx context.Context, req *csi.Controller
 	}
 
 	isOperationSucceeded = true
+
+	// Start migration monitoring if this is a SKU change
+	if monitorSKUMigration && d.migrationMonitor != nil {
+		pvName := diskParams.DiskName
+		if pvName == "" {
+			// Extract disk name from URI if not provided
+			_, _, diskName, parseErr := azureutils.GetInfoFromURI(diskURI)
+			if parseErr != nil {
+				klog.Warningf("Failed to extract disk name from URI %s: %v", diskURI, parseErr)
+			} else {
+				pvName = diskName
+			}
+		}
+
+		if pvName != "" {
+			if monitorErr := d.migrationMonitor.StartMigrationMonitoring(ctx, diskURI, pvName, fromSKU, skuName); monitorErr != nil {
+				klog.Warningf("Failed to start migration monitoring for disk %s: %v", diskURI, monitorErr)
+			}
+		}
+	}
+
 	klog.V(2).Infof("modify azure disk(%s) account type(%s) rg(%s) location(%s) successfully", diskParams.DiskName, skuName, diskParams.ResourceGroup, diskParams.Location)
 
 	return &csi.ControllerModifyVolumeResponse{}, err
--- a/pkg/azuredisk/migration_monitor.go
+++ b/pkg/azuredisk/migration_monitor.go
@@ -1,0 +1,364 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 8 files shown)


hasethuraman:
> thanks for the contribution. some comments:
> 
> * can you emit events on PVC instead of PV since ModifyVolume happens on PVC first, and customer usually get pvc events to check whether ModifyVolume is completed or not
> * this PR also changes the annotation of PV? why it's needed? is event not enough?

Annotation is to avoid creating duplicate events when CSI disk driver pod restarts. I can remove the annotations if we think it's ok to allow duplication

pkg/azuredisk/controllerserver.go
```diff
@@ -424,6 +424,19 @@ func (d *Driver) ControllerModifyVolume(ctx context.Context, req *csi.Controller
 		skuName = ""
 	}
 
+	// Get current disk state before modification
+	currentDisk, err := d.checkDiskExists(ctx, diskURI)
+	if err != nil {
+		return nil, status.Error(codes.NotFound, fmt.Sprintf("Volume not found, failed with error: %v", err))
+	}
```
mittachaitu: Do we really need to recheck disk existence? We are already doing at line 409, Can we reuse it?

pkg/azuredisk/migration_monitor.go
```diff
@@ -0,0 +1,364 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/klog/v2"
+)
+
+const (
+	// Migration monitoring constants
+	migrationCheckInterval     = 30 * time.Second
```
mittachaitu: Based on migration timeout good to have migrationCheckInterval higher than 30sec like 2min/5min

pkg/azuredisk/migration_monitor.go
```diff
@@ -0,0 +1,364 @@
...
+		kubeClient:     kubeClient,
+		eventRecorder:  eventRecorder,
+		diskController: diskController,
+		activeTasks:    make(map[string]*MigrationTask),
+	}
+}
+
+// StartMigrationMonitoring starts monitoring a disk migration with progress updates
+func (m *MigrationProgressMonitor) StartMigrationMonitoring(ctx context.Context, diskURI, pvName string, fromSKU, toSKU armcompute.DiskStorageAccountTypes) error {
+	m.mutex.Lock()
```
mittachaitu, line 194: Should we take lock at volume level instead instance level?

pkg/azuredisk/migration_monitor.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+// removeMigrationAnnotations removes migration-related annotations from the PersistentVolume
+func (m *MigrationProgressMonitor) removeMigrationAnnotations(ctx context.Context, pvName string) error {
+	pv, err := m.kubeClient.CoreV1().PersistentVolumes().Get(ctx, pvName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	if pv.Annotations != nil {
```
mittachaitu: nit: We can explicitly make delete without checking for non nil

andyzhangx:
> > thanks for the contribution. some comments:
> > 
> > * can you emit events on PVC instead of PV since ModifyVolume happens on PVC first, and customer usually get pvc events to check whether ModifyVolume is completed or not
> > * this PR also changes the annotation of PV? why it's needed? is event not enough?
> 
> Annotation is to avoid creating duplicate events when CSI disk driver pod restarts. I can remove the annotations if we think it's ok to allow duplication

@hasethuraman duplicated events should be ok and also CSI disk driver pod restarts happens very infrequently, patch annotation would have side effects, e.g. annotation never removed due to CSI disk driver pod restart?

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
```diff
@@ -76,7 +76,11 @@ spec:
           image: "{{ .Values.image.csiProvisioner.repository }}:{{ .Values.image.csiProvisioner.tag }}"
 {{- end }}
           args:
+{{- if .Values.feature.modifyVolume }}
```
andyzhangx: `VolumeAttributesClass` is a GA feature from k8s 1.34, so I think it's ok set as true by default, `.Values.feature.modifyVolume` is not necessary.

pkg/azuredisk/azuredisk.go
```diff
@@ -263,6 +268,27 @@ func NewDriver(options *DriverOptions) *Driver {
...
+			eventBroadcaster.StartRecordingToSink(&clientcorev1.EventSinkImpl{
+				Interface: kubeClient.CoreV1().Events(""),
+			})
+			eventRecorder := eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{
+				Component: driver.Name,
+			})
+			driver.eventRecorder = eventRecorder
+			driver.migrationMonitor = NewMigrationProgressMonitor(kubeClient, eventRecorder, driver.diskController)
+
+			// Recover any ongoing migrations after restart
```
landreasyan, line 288: This error happens in the go routine. Do we need to communicate it back to the main thread through channels?

pkg/azuredisk/migration_monitor.go
```diff
@@ -0,0 +1,364 @@
...
+
+// Recovery function using annotations
+func (d *Driver) recoverMigrationMonitorsFromAnnotations(ctx context.Context) error {
+	pvList, err := d.cloud.KubeClient.CoreV1().PersistentVolumes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	recoveredCount := 0
+	for _, pv := range pvList.Items {
```
landreasyan, line 566: Could we instead list PVC-s that have VAC bound to them and either rely on the status (which might actually not be very reliable as events will expire after some time) or add the annotation to those PVC-s during the migration?

pkg/azuredisk/migration_monitor.go
```diff
@@ -0,0 +1,364 @@
...
+)
+
+const (
+	// Migration monitoring constants
+	migrationCheckInterval     = 30 * time.Second
+	migrationTimeout           = 24 * time.Hour
+	progressReportingThreshold = 20 // Report every 20% completion
+
+	// Event types and reasons
+	EventTypeNormal  = "Normal"
```
jmclong: You can use 

k8s.io/api/core/v1

https://github.com/kubernetes/api/blob/b40c1cacbb902b21a7e0c7bf0967321860c1a632/core/v1/types.go#L7167

These constants exist here

hasethuraman:
> e.g. annotation never removed due to CSI disk driver pod restart?

When driver restart & later when the migration complete, we will cleanup the annotation. I am thinking to have a label for PVCs undergoing migration that can be filtered using label selector to resume the monitor upon driver restart. Instead of annotation, we will have label and remove it when migration complete

pkg/azuredisk/migration_monitor.go
hasethuraman: Yes, we have annotation but that cannot be used to filter I see. I will move to label & filter.

pkg/azuredisk/migration_monitor.go
hasethuraman: Increasing to 5 mins is high considering as the purpose of this to ease the maintenance and dont want to delay 5 mins in case of any delay in critical application volumes migration. Increased to 1 minute

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
hasethuraman: Removed

pkg/azuredisk/migration_monitor.go
```diff
@@ -0,0 +1,512 @@
...
+
+// GetActiveMigrations returns currently active migration tasks
+func (m *MigrationProgressMonitor) GetActiveMigrations() map[string]*MigrationTask {
+	m.mutex.RLock()
+	defer m.mutex.RUnlock()
+
+	result := make(map[string]*MigrationTask)
+	for k, v := range m.activeTasks {
+		taskCopy := *v
+		result[k] = &taskCopy
```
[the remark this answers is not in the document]
hasethuraman: It is added for testing purpose

pkg/azuredisk/migration_monitor.go
```diff
@@ -0,0 +1,512 @@
...
+				continue
+			}
+
+			if timeout, err = time.ParseDuration(timeoutStr); err != nil {
+				klog.Warningf("Invalid migration timeout duration: %s", timeoutStr)
+				continue
+			}
+
+			migrationTimeouts[size.Value()] = timeout
+			sortedMigrationSlabArray = append(sortedMigrationSlabArray, size.Value())
```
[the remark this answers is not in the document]
hasethuraman: Intention is to have that mentioned behaviour in case if we wanted override the behaviour through env without modifying code. Moreover, it is done for testing purpose.

pkg/azuredisk/azuredisk.go
```diff
@@ -263,6 +268,27 @@ func NewDriver(options *DriverOptions) *Driver {
...
+			go func() {
```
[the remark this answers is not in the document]
hasethuraman: 30 seconds a good timeout to continue with routine and large enough. Customizing is too much, even if it fails there is a bigger issue underneath. Will revisit later.

pkg/azuredisk/migration_monitor.go
```diff
@@ -0,0 +1,520 @@
...
+	// Label for migration status on PVC
+	LabelMigrationInProgress = "disk.csi.azure.com/migration-in-progress"
+
+	// Volume size thresholds (in bytes)
+	volumeSize2TB = 2 * 1024 * 1024 * 1024 * 1024 // 2TB
+	volumeSize4TB = 4 * 1024 * 1024 * 1024 * 1024 // 4TB
+
+	// Timeout durations based on volume size
+	migrationTimeoutSmall  = 5 * time.Hour  // Below 2TB
+	migrationTimeoutMedium = 9 * time.Hour  // 2TB to 4TB
```
andyzhangx: migrate a PV1 disk to PV2 would cost 9 hours? any data here why it's so long? the csi driver controller pod may not run for such long time without restart, so what if controller pod restarts?

pkg/azuredisk/migration_monitor.go
```diff
@@ -0,0 +1,520 @@
...
+		m.mutex.Lock()
+		delete(m.activeTasks, task.DiskURI)
+		m.mutex.Unlock()
+		task.CancelFunc()
+	}()
+
+	ticker := time.NewTicker(migrationCheckInterval)
+	defer ticker.Stop()
+
+	for {
```
andyzhangx: what about using kwait lib func, e.g. `kwait.PollUntilContextTimeout` to make logic more clear?

pkg/azuredisk/migration_monitor.go
```diff
@@ -0,0 +1,520 @@
...
+// Returns true if label already existed, false if it was newly added
+func (m *MigrationProgressMonitor) addMigrationLabelIfNotExists(ctx context.Context, pvcName, pvcNamespace string, fromSKU, toSKU armcompute.DiskStorageAccountTypes) (bool, error) {
+	pvc, err := m.kubeClient.CoreV1().PersistentVolumeClaims(pvcNamespace).Get(ctx, pvcName, metav1.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+
+	// Check if migration label already exists
+	if pvc.Labels != nil {
+		if value, exists := pvc.Labels[LabelMigrationInProgress]; exists && value == "true" {
```
andyzhangx: pls use annotations instead of lable for updating status since annotation value update is also used by pv controller, lable is specified by user, it's better not touching that value.

andyzhangx (commented):
btw, can you also provide an example events in PR description when you have done manual tests?

pkg/azuredisk/migration_monitor.go
```diff
@@ -0,0 +1,520 @@
...
+
+	pvcName := pv.Spec.ClaimRef.Name
+	pvcNamespace := pv.Spec.ClaimRef.Namespace
+
+	// Get volume size from PV capacity
+	var volumeSize int64
+	// Use the resource.Quantity methods explicitly to avoid import not recognized by linter/ide/goimports
+	var storageQuantity resource.Quantity
+	var exists bool
+	if storageQuantity, exists = pv.Spec.Capacity[corev1.ResourceStorage]; exists {
```
andyzhangx: write like following which is more clear:
```
storageQuantity, exists := pv.Spec.Capacity[corev1.ResourceStorage]
if exists {
```

pkg/azuredisk/controllerserver.go
```diff
@@ -406,7 +406,9 @@ func (d *Driver) ControllerModifyVolume(ctx context.Context, req *csi.Controller
 		return nil, status.Errorf(codes.Internal, "invalid modify volume req: %v", req)
 	}
 	diskURI := volumeID
-	if _, err := d.checkDiskExists(ctx, diskURI); err != nil {
+	var currentDisk *armcompute.Disk
+	var err error
+	if currentDisk, err = d.checkDiskExists(ctx, diskURI); err != nil {
```
andyzhangx:
```
currentDisk, err := d.checkDiskExists(ctx, diskURI)
if err != nil {
```

pkg/azuredisk/controllerserver.go
```diff
@@ -451,6 +461,27 @@ func (d *Driver) ControllerModifyVolume(ctx context.Context, req *csi.Controller
 	}
 
 	isOperationSucceeded = true
+
+	// Start migration monitoring if this is a SKU change
+	if monitorSKUMigration && d.migrationMonitor != nil {
+		pvName := diskParams.DiskName
+		if pvName == "" {
+			// Extract disk name from URI if not provided
+			_, _, diskName, parseErr := azureutils.GetInfoFromURI(diskURI)
```
andyzhangx: we could only rely on diskURI to get disk name (as the only source), that could make the logic more clear (remove L467, 468)

pkg/azuredisk/migration_monitor.go
hasethuraman: The sku migration can do other relevant data movement that could possibly take time. When controller pod restarts, we will use the label form PV (the latest update) to know the migration that is in progress & resume. I have also corrected some timeouts based on size range

pkg/azuredisk/migration_monitor.go
hasethuraman: Thanks, I have moved the label to PV now & PVC labels untouched.

hasethuraman: This is mainly to filter the PVs when controller pod restarts, we will start the migration monitor for the disks migrating

hasethuraman:
> btw, can you also provide an example events in PR description when you have done manual tests?

Done & addressed other comments too

andyzhangx:
@hasethuraman are you able to fix the race condition issue in ModifyVolume ?

```
==================
WARNING: DATA RACE
Read at 0x00c000e85528 by goroutine 1192:
  sigs.k8s.io/azuredisk-csi-driver/pkg/azuredisk.TestControllerModifyVolume.func3()
```

hasethuraman:
> @hasethuraman are you able to fix the race condition issue in ModifyVolume ?
> 
> ```
> ==================
> WARNING: DATA RACE
> Read at 0x00c000e85528 by goroutine 1192:
>   sigs.k8s.io/azuredisk-csi-driver/pkg/azuredisk.TestControllerModifyVolume.func3()
> ```

I had fixed all races, but the go routine is still causing data races here. I am resuming on this now to fix.

PR #3237, opened 2025-08-01
chore: only remove disks that can be processed from the batch

landreasyan:
**What type of PR is this?**
/kind bug

Only the disks that can be processed in a single batch request should be removed from the batching map and per CSI Spec, when node disk attach limit is reached, the driver should return ResourceExhausted errors. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


nearora-msft:
Does this fix any issues, or is this just to ensure that we return the right error to K8s?

landreasyan:
> Does this fix any issues, or is this just to ensure that we return the right error to K8s?

It fixes the logic of removing the disks queued for batching. Disks that cannot be processed in the current batch will be left in the queue to be processed with the next request if there is capacity or otherwise, report the right error.

andyzhangx:
@landreasyan  `ResourceExhausted` error won't make reschedule happen, that won't mitigate the original MaxDataDiskNumExcceeded issue unless user deletes the pod manually to make reschedule happen

pkg/azuredisk/azure_controller_common.go
```diff
@@ -232,13 +232,13 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 				if int(maxNumDisks) > numDisksAttached {
 					numDisksAllowed = int(maxNumDisks) - numDisksAttached
 				} else {
-					numDisksAllowed = 0
+					return -1, fmt.Errorf("Maximum number of disks %s %d", util.MaximumDataDiskExceededMsg, maxNumDisks)
```
andyzhangx: this is actually the original behavior, it just returns MaximumDataDiskExceededMsg when calling azure API, but behavior is exactly the same as before

landreasyan: Yes, this part just makes the code nicer. The changes are made in the refactored code in the `retrieveAttachBatchedDiskRequests`.

landreasyan:
> @landreasyan `ResourceExhausted` error won't make reschedule happen, that won't mitigate the original MaxDataDiskNumExcceeded issue unless user deletes the pod manually to make reschedule happen

Yes, it won't but regardless, we should return the right error. This is called out in the csi spec as well: https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerpublishvolume-errors

pkg/azuredisk/azure_controller_common.go
```diff
@@ -350,7 +334,26 @@ func (c *controllerCommon) retrieveAttachBatchedDiskRequests(nodeName, diskURI s
 		klog.V(2).Infof("no attach disk(%s) request on node(%s), diskMap len:%d, %+v", diskURI, nodeName, len(diskMap), diskMap)
 		return nil, nil
 	}
-	c.attachDiskMap.Store(nodeName, make(map[string]*provider.AttachDiskOptions))
```
andyzhangx, line 353: original logic is always clearing up the queue since if removeDisks > 0, keep the remaining disk in the queue would not succeed in the end. and CSI driver has retry logic. clearing up the queue would make the logic more straightfoward.

pkg/azuredisk/azure_controller_common.go
andyzhangx: thanks for the explanation, I think this makes sense since if `numDisksAllowed = 0`, it does not need to go to next steps, just return the error directly

pkg/azuredisk/azure_controller_common.go
landreasyan: I really wanted to avoid dropping the disks that haven't been processed, even though I agree that k8s will eventually retry but more changes need to be made to do this safely, so I am adding the cleaning up the queue part back.

nearora-msft: With this approach, the next attach request that comes, would add to this batch of already existing disks in the attachDiskMap? So, we won't create a fresh batch, but reuse the originally dropped disks?
Is that the correct understanding?

andyzhangx:
this PR has made following test fail:
```
Summarizing 1 Failure:
  [FAIL] Dynamic Provisioning [single-az] [It] should succeed without MaximumDataDisksExceeded
  /home/prow/go/src/sigs.k8s.io/azuredisk-csi-driver/test/e2e/testsuites/dynamically_provisioned_job_attach_batch_tester.go:97
```

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_azuredisk-csi-driver/3237/pull-azuredisk-csi-driver-e2e-capz/1952867569819979776

k-benzarti:
/reopen
/remove-lifecycle rotten

I think we should we take another look at this.

At the time this was evaluated, the feedback from @andyzhangx was that returning ResourceExhausted wouldn't trigger a reschedule. However, now that KEP-4876 has graduated to Beta in K8s 1.34/1.35 (https://kubernetes.io/blog/2025/09/11/kubernetes-v1-34-mutable-csi-node-allocatable-count/), the kubelet does actively listen for ResourceExhausted (gRPC code 8) and will instantly transition the pod to Failed so the scheduler can move it. Because the driver currently returns Internal (gRPC code 13), AKS users are currently locked out of this upstream Kubernetes safety net and pods are still getting permanently stuck in ContainerCreating state.

PR #3247, opened 2025-08-11
chore: add an option to configure instance-view param in VMSS VM list API

landreasyan:
… API

**What type of PR is this?**
/kind feature

There are 2 API calls for [VMSS VM List](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machine-scale-set-vms/list?view=rest-compute-2025-02-01&tabs=HTTP); however, the sdk only implements the one with instance-view parameter.

VMSS VM List API with instance-view parameter times out when a cluster zone is not available and is not reliable for scenarios that require zonal redundance. I am adding an option of selecting which API to call to populate the cache in order to accommodate all scenarios.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
+++ b/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
@@ -235,6 +235,7 @@ spec:
             - "--user-agent-suffix={{ .Values.driver.userAgentSuffix }}"
             - "--allow-empty-cloud-config={{ .Values.controller.allowEmptyCloudConfig }}"
             - "--vmss-cache-ttl-seconds={{ .Values.controller.vmssCacheTTLInSeconds }}"
+            - "--vmss-cache-with-instance-view={{ .Values.controller.vmssCacheWithInstanceView }}"
             - "--enable-traffic-manager={{ .Values.controller.enableTrafficManager }}"
             - "--traffic-manager-port={{ .Values.controller.trafficManagerPort }}"
             - "--enable-otel-tracing={{ .Values.controller.otelTracing.enabled }}"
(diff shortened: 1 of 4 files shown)


landreasyan:
/hold until PR in cloud-provider is merged and new vendored in the PR: https://github.com/kubernetes-sigs/cloud-provider-azure/pull/9339 

Will adjust the flags once the PR on cloud-provider is finalized and merged.

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
```diff
@@ -235,6 +235,7 @@ spec:
             - "--user-agent-suffix={{ .Values.driver.userAgentSuffix }}"
             - "--allow-empty-cloud-config={{ .Values.controller.allowEmptyCloudConfig }}"
             - "--vmss-cache-ttl-seconds={{ .Values.controller.vmssCacheTTLInSeconds }}"
+            - "--vmss-cache-with-instance-view={{ .Values.controller.vmssCacheWithInstanceView }}"
```
andyzhangx: this is only for open source install, let's revert this change to make it simple now since in the end, we don't need ListInstanceView so we don't need to expose this in chart config

charts/latest/azuredisk-csi-driver/values.yaml
```diff
@@ -53,6 +53,7 @@ controller:
   provisionerWorkerThreads: 100
   attacherWorkerThreads: 1000
   vmssCacheTTLInSeconds: -1
+  vmssCacheWithInstanceView: false
```
andyzhangx: remove this

pkg/azuredisk/azuredisk_option.go
```diff
@@ -101,6 +102,7 @@ func (o *DriverOptions) AddFlags() *flag.FlagSet {
 	fs.Int64Var(&o.TrafficManagerPort, "traffic-manager-port", 7788, "default traffic manager port")
 	fs.Int64Var(&o.AttachDetachInitialDelayInMs, "attach-detach-initial-delay-ms", 1000, "initial delay in milliseconds for batch disk attach/detach")
 	fs.Int64Var(&o.VMSSCacheTTLInSeconds, "vmss-cache-ttl-seconds", -1, "vmss cache TTL in seconds (600 by default)")
+	fs.BoolVar(&o.VMSSCacheWithInstanceView, "vmss-cache-with-instance-view", true, "boolean flag to enable vmss cache with instance view")
```
andyzhangx: `ListVMSSWithInstanceView` and set as false by default

landreasyan:
> I have created a PR to upgrade the azure cloud provider lib first: #3252

Thanks for merging my PR in the cloud provider lib and creating a new release so promptly.

andyzhangx (approved):
/lgtm

PR #3286, opened 2025-08-27
feat: monitor migration for premiumv2 disk created from snapshot and edge case handling

hasethuraman:
**What type of PR is this?**

/kind feature

When a backing Azure Disk migrates from one SKU to another, we generate events to indicate the start, progress, and completion of the migration. This event generation is currently supported only when the migration is triggered via the ModifyVolume path.
However, there are scenarios where a disk is created from a snapshot of a different SKU. In such cases, event generation is also necessary. To support this, we have enhanced the CreateVolume path to reuse the same infrastructure for emitting migration events.
image

Since the CreateVolume path may not allow labeling the Persistent Volume (PV) to track migrating disks, this has been handled accordingly, while ensuring that the behavior in the ModifyVolume path remains intact.

Additionally, we addressed an edge case in the ModifyVolume path where the controller might restart after the disk SKU is updated but before the operation returns success.

Fixes #

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azureconstants/azure_constants.go
+++ b/pkg/azureconstants/azure_constants.go
@@ -115,7 +115,6 @@ const (
 var (
 	// ManagedDiskPath is described here: https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate#create-a-managed-disk-from-an-existing-managed-disk-in-the-same-or-different-subscription.
 	ManagedDiskPath    = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/disks/%s"
-	SnapshotPath       = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/snapshots/%s"
 	ManagedDiskPathRE  = regexp.MustCompile(`(?i).*/subscriptions/(?:.*)/resourceGroups/(?:.*)/providers/Microsoft.Compute/disks/(.+)`)
 	DiskSnapshotPathRE = regexp.MustCompile(`(?i).*/subscriptions/(?:.*)/resourceGroups/(?:.*)/providers/Microsoft.Compute/snapshots/(.+)`)
 )
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -317,7 +317,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 			// if migration monitoring is enabled and if target SKU is PremiumV2LRS, get the snapshot SKU
 			// the snapshot SKU is marked as source SKU
 			if skuName == armcompute.DiskStorageAccountTypesPremiumV2LRS && d.migrationMonitor != nil {
-				sourceSKU = d.getSnapshotSKU(ctx, sourceID)
+				sourceSKU, _ = d.getSnapshotSKU(ctx, sourceID)
 			}
 			metricsRequest = "controller_create_volume_from_snapshot"
 		} else {
@@ -1442,25 +1442,25 @@ func (d *Driver) GetSourceDiskSize(ctx context.Context, subsID, resourceGroup, d
 }
 
 // getSnapshotSKU retrieves the SKU of the snapshot and returns the SKU or if any error occurs
-func (d *Driver) getSnapshotSKU(ctx context.Context, sourceID string) string {
+func (d *Driver) getSnapshotSKU(ctx context.Context, sourceID string) (string, error) {
 	subsID, resourceGroup, snapshotName, err := azureutils.GetInfoFromURI(sourceID)
 	if err != nil {
 		klog.Warningf("could not get subscription id, resource group from snapshot uri (%s) with error(%v)", sourceID, err)
-		return ""
+		return "", err
 	}
 	snapClient, err := d.clientFactory.GetSnapshotClientForSub(subsID)
 	if err != nil {
 		klog.Warningf("could not get snapshot client for subscription(%s) with error(%v)", subsID, err)
-		return ""
+		return "", err
 	}
 	result, err := snapClient.Get(ctx, resourceGroup, snapshotName)
 	if err != nil {
 		klog.Warningf("get snapshot %s from rg(%s) error: %v", snapshotName, resourceGroup, err)
-		return ""
+		return "", err
 	}
 	if result == nil || result.SKU == nil || result.SKU.Name == nil {
 		klog.Warningf("Snapshot or Snapshot property not found for snapshot (%s) in resource group (%s)", snapshotName, resourceGroup)
-		return ""
+		return "", status.Error(codes.NotFound, fmt.Sprintf("Snapshot or Snapshot property not found for snapshot (%s) in resource group (%s)", snapshotName, resourceGroup))
 	}
-	return string(*result.SKU.Name)
+	return string(*result.SKU.Name), nil
 }
--- a/pkg/azuredisk/controllerserver_test.go
+++ b/pkg/azuredisk/controllerserver_test.go
@@ -326,7 +326,7 @@ func TestCreateVolume(t *testing.T) {
 				mp := make(map[string]string)
 				mp["tags"] = "unit=test"
 				volumeSnapshotSource := &csi.VolumeContentSource_SnapshotSource{
-					SnapshotId: fmt.Sprintf(consts.SnapshotPath, "subs", "rg", "unit-test"),
+					SnapshotId: fmt.Sprintf(diskSnapshotPath, "subs", "rg", "unit-test"),
 				}
 				volumeContentSourceSnapshotSource := &csi.VolumeContentSource_Snapshot{
 					Snapshot: volumeSnapshotSource,
@@ -4075,11 +4075,13 @@ func TestGetSourceDiskSize(t *testing.T) {
 
 func TestGetSnapshotSKU(t *testing.T) {
 	type testCase struct {
-		name          string
-		snapshotURI   string
-		setupMocks    func(factory *mock_azclient.MockClientFactory, snap *mock_snapshotclient.MockInterface)
-		expectedSKU   string
-		expectFactory bool
+		name            string
+		snapshotURI     string
+		setupMocks      func(factory *mock_azclient.MockClientFactory, snap *mock_snapshotclient.MockInterface)
+		expectedSKU     string
+		expectFactory   bool
+		expectErrSubstr string
+		expectGRPCCode  codes.Code
 	}
 	tests := []testCase{
 		{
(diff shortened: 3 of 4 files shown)


pkg/azuredisk/controllerserver.go
```diff
@@ -225,6 +225,20 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+			}
+			snap, err := d.GetSourceSnapshot(ctx, subsID, resourceGroup, snapName)
+			if err == nil {
+				if snap.SKU != nil && snap.SKU.Name != nil {
+					sourceSKU = string(*snap.SKU.Name)
+				} else {
+					klog.Warningf("snapshot sku is nil, snapshotName: %s", snapName)
+				}
+			} else {
+				klog.Warningf("failed to get source snapshot(%s) size, err: %v", sourceID, err)
```
mittachaitu: IMO, If we are unable to get snapshot details better to return error instead of skipping it...

pkg/azuredisk/controllerserver.go
```diff
@@ -341,6 +358,28 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Errorf(codes.Internal, "%v", err)
 	}
 
+	// After disk creation success and you have diskObj (armcompute.Disk) & req.VolumeContentSource != nil && snapshot != nil
+	if skuName == armcompute.DiskStorageAccountTypesPremiumV2LRS && sourceSKU != string(skuName) {
```
mittachaitu:
```suggestion
	if skuName == armcompute.DiskStorageAccountTypesPremiumV2LRS && sourceSKU != string(skuName) {
```
Should we need to keep strict check for `sourceSKU`? Instead of comparing with current `skuName`? Is migration even supported from UltraDisk to premiumV2?

pkg/azuredisk/controllerserver.go
```diff
@@ -429,9 +468,19 @@ func (d *Driver) ControllerModifyVolume(ctx context.Context, req *csi.Controller
...
+	if currentDisk != nil {
+		if currentDisk.SKU != nil && currentDisk.SKU.Name != nil {
+			fromSKU = *currentDisk.SKU.Name
+			monitorSKUMigration = skuName == armcompute.DiskStorageAccountTypesPremiumV2LRS && fromSKU == armcompute.DiskStorageAccountTypesPremiumLRS
+		}
+
+		// modifyVolume will be reattempted if controller restarts. In case if the controller restarted after updating disk sku to PremiumV2LRS & before we label,
+		// we may not be able to detect the migration post restart, hence during re-attempt, we can check this condition and set monitorSKUMigration accordingly
+		if !monitorSKUMigration {
+			if currentDisk.Properties != nil && currentDisk.Properties.CompletionPercent != nil && *currentDisk.Properties.CompletionPercent < float32(100.0) {
```
mittachaitu: Do we need to have check for CompletionPercent? By the time we receive modifyVolume request if migration is completed then we will miss to generate event, right? Can we remove that check so that existing flow will generate that migration completed event?
```suggestion
			if currentDisk.Properties != nil && currentDisk.Properties.CompletionPercent != nil  {
```

pkg/azuredisk/migration_monitor.go
```diff
@@ -190,7 +191,7 @@ func NewMigrationProgressMonitor(kubeClient kubernetes.Interface, eventRecorder
 }
 
 // StartMigrationMonitoring starts monitoring a disk migration with progress updates
-func (m *MigrationProgressMonitor) StartMigrationMonitoring(ctx context.Context, diskURI, pvName string, fromSKU, toSKU armcompute.DiskStorageAccountTypes) error {
+func (m *MigrationProgressMonitor) StartMigrationMonitoring(ctx context.Context, provisioningFlow bool, diskURI, pvName string, fromSKU, toSKU armcompute.DiskStorageAccountTypes, volumeSize int64) error {
```
mittachaitu: nit:
```suggestion
func (m *MigrationProgressMonitor) StartMigrationMonitoring(ctx context.Context, isProvisioningFlow bool, diskURI, pvName string, fromSKU, toSKU armcompute.DiskStorageAccountTypes, volumeSize int64) error {
```

pkg/azuredisk/migration_monitor.go
```diff
@@ -244,24 +239,27 @@ func (m *MigrationProgressMonitor) StartMigrationMonitoring(ctx context.Context,
...
 		MigrationTimeout:     migrationTimeout,
+		PVLabeled:            false,
 	}
 
 	m.activeTasks[diskURI] = task
 
-	// Start async monitoring
-	go m.monitorMigrationProgress(task)
-
 	// Add label to PVC to track migration state and check if it was already present
```
mittachaitu:
```suggestion
	// Add label to PV to track migration state and check if it was already present
```

mittachaitu (commented):
Provided few comments PTAL!!

pkg/azuredisk/controllerserver.go
andyzhangx: this is an optional feature, that would break existing volume creation workflow if returns error

pkg/azuredisk/controllerserver.go
```diff
@@ -341,6 +358,28 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		return nil, status.Errorf(codes.Internal, "%v", err)
 	}
 
+	// After disk creation success and you have diskObj (armcompute.Disk) & req.VolumeContentSource != nil && snapshot != nil
+	if skuName == armcompute.DiskStorageAccountTypesPremiumV2LRS && sourceSKU != string(skuName) {
+		if d.migrationMonitor != nil {
+			// convert sourceSKU which may be in SnapshotStorageAccountTypes to DiskStorageAccountTypes
+			var srcSkuName armcompute.DiskStorageAccountTypes
```
andyzhangx: can you write a standalone func to wrap this convert logic, this CreateVolume func is becoming longer and longer now

andyzhangx: and wrap the function would make writing ut easier

pkg/azuredisk/controllerserver.go
```diff
@@ -225,6 +225,20 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 					},
 				},
 			}
+			subsID, resourceGroup, snapName, err := azureutils.GetInfoFromURI(sourceID)
+			if err != nil {
+				return nil, status.Errorf(codes.NotFound, "%v", err)
+			}
+			snap, err := d.GetSourceSnapshot(ctx, subsID, resourceGroup, snapName)
```
andyzhangx: this would make an extra get snapshot API call only to get the snapshot sku name, and `CreateManagedDisk` would could return the created disk properties, is it possible to get the sku name using that function? that may removing the need of get snapshot API calls

pkg/azuredisk/controllerserver.go
hasethuraman: Today expected source SKUs are Standard_ZRS (snapshot) and Premium_LRS (managed disk). But the managed disk can be HDDs too. If backend supports the UltraDisk as sourceSKU for creating PremiiumV2_LRS, then this code path will take care of starting the monitor. If backend do not support, it wont hit this place at all. So as long as we ensure that we focus the monitor code only for the PremiumV2_LRS (as targetSKU), it is sufficient & avoid cluttering this condition.

pkg/azuredisk/controllerserver.go
hasethuraman: Agreed to @andyzhangx 's comment, in fact I will skip API calls if the feature is not enabled. Will use this comment to track it.

pkg/azuredisk/controllerserver.go
hasethuraman: This code is designed to handle the edge condition where SKU change monitoring is expected to take several minutes. During this period, the completion percentage will remain below 100%. There is no scenario where the migration completes so quickly that event generation is missed. In such cases, monitoring is not required, as the migration is conducted under planned supervision—specifically when the user applies VolumeAttributeClass—and is aware of whether to check for events.
Additionally, removing this condition from the ModifyVolume path would incorrectly trigger events when a user modifies only QoS parameters for a migrated disk, which is not the intended behavior

pkg/azuredisk/controllerserver.go
hasethuraman: @andyzhangx The path is creating a PremiumV2_LRS disk using a snapshot originally taken from a Premium_LRS disk. Given that, the snapshot must be of SKU Standard_ZRS—should I presume simply, because the source SKU (Standard_ZRS) wouldn’t be available in the createManagedDisk response, correct?

hasethuraman: Retaining the same, not assuming Standard_ZRS and using the GetSourceSnapshot to retrieve the SKU of snapshot

pkg/azuredisk/controllerserver.go
```diff
@@ -63,6 +63,35 @@ type listVolumeStatus struct {
...
 
+func (d *Driver) startSKUMigrationMonitor(
+	ctx context.Context,
+	isProvisioningFlow bool,
+	fromSKUStr string,
+	toSKU armcompute.DiskStorageAccountTypes,
+	diskURI, pvName string,
+	sizeBytes int64,
+) {
+	if d.migrationMonitor == nil || fromSKUStr == "" {
```
mittachaitu, line 134: nit: How about moving these checks into function called `shouldStartMigrationMonitoring`?

pkg/azuredisk/controllerserver_test.go
```diff
@@ -1227,6 +1419,296 @@ func TestControllerModifyVolume(t *testing.T) {
 	}
 }
 
+func TestControllerModifyVolume_MigrationLifecycleAndTimeout(t *testing.T) {
```
mittachaitu, line 1423: Should we also need to have testcase for not emitting the event?

mittachaitu (approved):
LGTM

pkg/azuredisk/controllerserver_test.go
hasethuraman: @mittachaitu What is the flow you have in mind?

pkg/azuredisk/controllerserver.go
hasethuraman: For now, not needed @mittachaitu until we have more conditions coming in.

pkg/azuredisk/controllerserver_test.go
hasethuraman: We will discuss further & have a follow up PR if more coverage is required.

pkg/azuredisk/migration_monitor.go
```diff
@@ -275,6 +273,12 @@ func (m *MigrationProgressMonitor) StartMigrationMonitoring(ctx context.Context,
 // monitorMigrationProgress monitors the progress of a disk migration using PollUntilContextTimeout
 func (m *MigrationProgressMonitor) monitorMigrationProgress(task *MigrationTask) {
 
+	isPvcInfoEmpty := func() bool {
+		m.mutex.RLock()
```
andyzhangx: `m.mutex` is MigrationProgressMonitor wide field, but task exists for every pv, what about add `mutex` in task field, otherwise when there are lots of PVs, it could be locked frequently since there is only one mutex for the whole driver

hasethuraman: Added the mutex at the task level, which has reduced contention by over 90%. I'm currently retaining the monitor-level mutex, as it's only used to serialize add/remove/read operations on the task list—whose content is minimal. Given its limited scope, the impact on performance is negligible.

pkg/azuredisk/controllerserver.go
```diff
@@ -225,6 +254,22 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 					},
 				},
 			}
+			if d.migrationMonitor != nil {
```
andyzhangx: pls check skuName first, if it's not PremiumV2LRS, then skip, that would save one get snapshot api call

pkg/azuredisk/controllerserver.go
```diff
@@ -225,6 +254,22 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+				subsID, resourceGroup, snapName, err := azureutils.GetInfoFromURI(sourceID)
```
andyzhangx: wrap this code as a new func like getSnapshotSKU(sourceID)

pkg/azuredisk/controllerserver.go
andyzhangx: and it's possible that skuName is empty, which means skuName is same as sourceSKU

pkg/azuredisk/controllerserver.go
```diff
@@ -63,6 +63,35 @@ type listVolumeStatus struct {
 	err           error
 }
 
+func (d *Driver) startSKUMigrationMonitor(
+	ctx context.Context,
+	isProvisioningFlow bool,
```
andyzhangx, line 128: add a comment for this func, especially for isProvisioningFlow parameter

pkg/azuredisk/controllerserver.go
```diff
@@ -463,16 +524,9 @@ func (d *Driver) ControllerModifyVolume(ctx context.Context, req *csi.Controller
...
-		_, _, diskName, parseErr := azureutils.GetInfoFromURI(diskURI)
-		if parseErr != nil {
-			klog.Warningf("Skipping monitor, failed to extract disk name from URI %s: %v", diskURI, parseErr)
-		} else {
-			if monitorErr := d.migrationMonitor.StartMigrationMonitoring(ctx, diskURI, diskName, fromSKU, skuName); monitorErr != nil {
-				klog.Warningf("Failed to start migration monitoring for disk %s: %v", diskURI, monitorErr)
-			}
-		}
+	if monitorSKUMigration {
+		volSizeBytes := int64(*currentDisk.Properties.DiskSizeGB) * 1024 * 1024 * 1024
```
andyzhangx, line 578: pls add nil pointer check for currentDisk and Properties

pkg/azuredisk/controllerserver.go
```diff
@@ -1336,3 +1390,19 @@ func (d *Driver) GetSourceDiskSize(ctx context.Context, subsID, resourceGroup, d
 	}
 	return (*result.Properties).DiskSizeGB, result, nil
 }
+
+// GetSourceSnapshot recursively searches for the sourceSnapshot and returns: sourceSnapshot snapshot size, error
+func (d *Driver) GetSourceSnapshot(ctx context.Context, subsID, resourceGroup, snapshotName string) (*armcompute.Snapshot, error) {
```
andyzhangx: pls add ut

pkg/azuredisk/migration_monitor.go
```diff
@@ -168,6 +168,8 @@ type MigrationTask struct {
 	MigrationTimeout     time.Duration // Calculated timeout based on volume size
 	Timedout             bool          // Indicates if the task has timed out
 	Cancelled            atomic.Bool   // Indicates if the task was cancelled
+	PVLabeled            bool          // Indicates if the PV has been labelled for migration
+	mutex                sync.RWMutex
```
andyzhangx: add comment about how is this mutex used

pkg/azuredisk/controllerserver.go
andyzhangx: so `startSKUMigrationMonitor` is used on both CreateVolume and ModifyVolume, comment the workflow for the function, e.g. how it labels the pvc, etc.

pkg/azuredisk/controllerserver.go
andyzhangx: L258 `subsID, resourceGroup, snapName, err := azureutils.GetInfoFromURI(sourceID)` is actually not necessary since in L262, it also takes subsID, so just use a new func like `GetSnapshotSKU(sourceID)` to get snapshotSKU, that would be more clear

pkg/azuredisk/controllerserver.go
hasethuraman: Added skuName == PremiumLRS and likewise we ensure targetSKU is PremiumV2LRS inside the helper function. Hence, this way the empty skuName case is taken care too.

pkg/azuredisk/controllerserver.go
hasethuraman: Added getSnapshotSKU and moved GetInfFromURI into getSnapshotSKU as we need the resource group and snap name as the snapshot client expects it.

pkg/azuredisk/controllerserver.go
hasethuraman: Instead added the properties nil check above and so monitorSKUMigration is set only if disk + properties are not nil

pkg/azuredisk/controllerserver.go
hasethuraman: @andyzhangx addressed, keeping this latest comment open to not to merge until I re-run some manual unit test. Updated the PR to review meanwhile.

Will resolve once done with re-runs

pkg/azuredisk/controllerserver.go
```diff
@@ -225,6 +314,20 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 					},
 				},
 			}
+			// if migration monitoring is enabled and if target SKU is PremiumV2LRS, get the snapshot SKU
+			// the snapshot SKU is marked as source SKU
+			if skuName == armcompute.DiskStorageAccountTypesPremiumV2LRS && d.migrationMonitor != nil {
+				snapSku, err := d.getSnapshotSKU(ctx, sourceID)
```
andyzhangx: what about return the string directly? and move the warning logs into getSnapshotSKU func

hasethuraman: We can, but the UT for testing will never know if the appropriate errors are returned. I am changing it (in order to reduce the lines create volume) and test "" & presume the errors are appropriate give there are few calls.

pkg/azureconstants/azure_constants.go
```diff
@@ -115,6 +115,7 @@ const (
 var (
 	// ManagedDiskPath is described here: https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate#create-a-managed-disk-from-an-existing-managed-disk-in-the-same-or-different-subscription.
 	ManagedDiskPath    = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/disks/%s"
+	SnapshotPath       = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/snapshots/%s"
```
andyzhangx: this var is not used?

pkg/azuredisk/controllerserver.go
```diff
@@ -1336,3 +1440,27 @@ func (d *Driver) GetSourceDiskSize(ctx context.Context, subsID, resourceGroup, d
 	}
 	return (*result.Properties).DiskSizeGB, result, nil
 }
+
+// getSnapshotSKU retrieves the SKU of the snapshot and returns the SKU or if any error occurs
+func (d *Driver) getSnapshotSKU(ctx context.Context, sourceID string) string {
+	subsID, resourceGroup, snapshotName, err := azureutils.GetInfoFromURI(sourceID)
+	if err != nil {
+		klog.Warningf("could not get subscription id, resource group from snapshot uri (%s) with error(%v)", sourceID, err)
```
andyzhangx, line 1448: this func would log warning everywhere, more common practices is `return "", fmt.Errorf(...)' in this func, and then in the caller, it could print out warning msg in outsider, and you could also write ut easier since you could get detailed error msg when error happens.

pkg/azureconstants/azure_constants.go
hasethuraman: Thanks, noticing there is one already. Removed this and using the existing one.

pkg/azuredisk/controllerserver.go
hasethuraman: Since it is getSnapshotSKU & so local only within azuredisk, I have made to return the err that we get from other functions and print the readable message to help while debugging as it is optional path without requiring createVolume to print. Now the UT is reverted slightly back to be compatible to check the errors.

PR #3294, opened 2025-09-02
fix: detach call should not use the full context deadline and should leave an active context for force detach on timeout

landreasyan:
**What type of PR is this?**
/bug

When Detach call times out, it currently uses the full context deadline to wait for a timeout from the platform. This blocks CSI from being able to issue a Force Detach call on timeouts. Detach call should not use the full context deadline for the operation and should leave an active context for force detach. 

This PR allows to define a min context timeout a Detach call should use, if the context has a longer timeout, then the operation would use half of the context's deadline and leave the half for Force Detach, otherwise it will use it's min timeout. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azure_controller_common.go
+++ b/pkg/azuredisk/azure_controller_common.go
@@ -61,6 +61,9 @@ const (
 	// default initial delay in milliseconds for batch disk attach/detach
 	defaultAttachDetachInitialDelayInMs = 1000
 
+	// default detach operation timeout
+	defaultDetachOperationMinTimeoutInSeconds = 240
+
 	// WriteAcceleratorEnabled support for Azure Write Accelerator on Azure Disks
 	// https://docs.microsoft.com/azure/virtual-machines/windows/how-to-enable-write-accelerator
 	WriteAcceleratorEnabled = "writeacceleratorenabled"
@@ -90,6 +93,8 @@ type controllerCommon struct {
 	detachDiskMap sync.Map
 	// DisableDiskLunCheck whether disable disk lun check after disk attach/detach
 	DisableDiskLunCheck bool
+	// DetachOperationMinTimeoutInSeconds determines timeout for waiting on detach operation before a force detach
+	DetachOperationMinTimeoutInSeconds int
 	// AttachDetachInitialDelayInMs determines initial delay in milliseconds for batch disk attach/detach
 	AttachDetachInitialDelayInMs int
 	ForceDetachBackoff           bool
@@ -394,6 +399,8 @@ func (c *controllerCommon) DetachDisk(ctx context.Context, diskName, diskURI str
 
 	if len(diskMap) == 0 {
 		// disk was already processed in the batch, return the result
+		// returns success when the disk has toBeDetached flag set to true
+		// we could consider issuing a force detach if the disk is stuck in ToBeDetached state for too long
 		return c.verifyDetach(ctx, diskName, diskURI, nodeName)
 	}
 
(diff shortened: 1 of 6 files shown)


charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
```diff
@@ -110,7 +110,7 @@ spec:
           args:
             - "-v=2"
             - "-csi-address=$(ADDRESS)"
-            - "-timeout=1200s"
+            - "-timeout=8m"
```
andyzhangx: why changing the default timeout from 20min to 8min? I think you only need to set 10min timeout for common disk detach, and set another 10min for force detach timeout.

landreasyan: I don't want to wait 20mins for a VM updates to complete and ideally want to timeout within 2mins. This is still a higher limit than what I was hoping for. I would rather have faster timeouts and have k8s retry instead of holding a node lock for more than 2-3mins. 

We can leave the limit in AKS to 20mins as is and change it later.

pkg/azuredisk/azure_controller_common.go
```diff
@@ -401,7 +406,20 @@ func (c *controllerCommon) DetachDisk(ctx context.Context, diskName, diskURI str
 	c.diskStateMap.Store(formattedDiskURI, "detaching")
 	defer c.diskStateMap.Delete(formattedDiskURI)
 
-	if err = vmset.DetachDisk(ctx, nodeName, diskMap, false); err != nil {
+	// Calculate timeout for regular detach operation, if operation fails or times out,
+	// we need to have an active context to issue a force detach
+	detachContextTimeout := time.Duration(c.DetachOperationMinTimeoutInMs) * time.Millisecond
```
andyzhangx: `DetachOperationMinTimeoutInSeconds`
it's better using seconds which is easier to read

pkg/azuredisk/azure_controller_common.go
```diff
@@ -394,14 +399,29 @@ func (c *controllerCommon) DetachDisk(ctx context.Context, diskName, diskURI str
...
 	}
 
 	klog.V(2).Infof("Trying to detach volume %s from node %s, diskMap len:%d, %s", diskURI, nodeName, len(diskMap), diskMap)
 	c.diskStateMap.Store(formattedDiskURI, "detaching")
 	defer c.diskStateMap.Delete(formattedDiskURI)
 
-	if err = vmset.DetachDisk(ctx, nodeName, diskMap, false); err != nil {
+	// Calculate timeout for regular detach operation, if operation fails or times out,
+	// we need to have an active context to issue a force detach
+	detachContextTimeout := time.Duration(c.DetachOperationMinTimeoutInSeconds) * time.Millisecond
```
andyzhangx: it should be `* time.Second`?

charts/latest/azuredisk-csi-driver/values.yaml
```diff
@@ -72,6 +72,7 @@ controller:
   vmType: ""
   provisionerWorkerThreads: 100
   attacherWorkerThreads: 1000
+  detachOperationMinTimeoutInSeconds: 120
```
andyzhangx: should be 240 here? and actually it's not necessary to expose every value and setting in helm chart config, you could revert all changes in helm chart config in this PR, that would be easier for cherrypick.

landreasyan: yeah, I can revert the changes for this PR.

PR #3344, opened 2025-09-22
fix: reconcile loop to watch migrations if earlier submission had failed

hasethuraman:
**What type of PR is this?**

/kind bug
/kind feature

If the label was never applied, migration won’t be monitored (this PR does not add retries for the initiation start migration failure due to any kubernetes transient issue).

Added the label‑driven recovery in a loop that re-establishes migration monitoring (emitting SKUMigrationStarted, SKUMigrationProgress, SKUMigrationCompleted events) for Premium_LRS → PremiumV2_LRS disk migrations after a controller restart or an earlier transient failure that prevented the in-memory monitor from being created.

With this the background goroutine (controller instances only, i.e. NodeID == "") sleeps 30s after startup, then every 10 minutes calls recoverMigrationMonitorsFromLabels(...).
Any PersistentVolume with label disk.csi.azure.com/migration-in-progress=true and CSI driver = this driver gets a new in-memory monitoring task (unless one is already active).

Manual recovery path:
If users see no events due to a transient start failure, they can add the label themselves and the next scan will attach monitoring.

Fixes #
Migration monitor failed to start due to transient issue and so events dont show up
In below test result, we can see after 10 mins the recovery found a volume with label as I manually added to test the behaviour post transient failure.
image

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -289,7 +289,6 @@ func NewDriver(options *DriverOptions) *Driver {
 
 		if kubeClient != nil && driver.NodeID == "" && driver.enableMigrationMonitor {
 			eventBroadcaster := record.NewBroadcaster()
-			eventBroadcaster.StartStructuredLogging(0)
 			eventBroadcaster.StartRecordingToSink(&clientcorev1.EventSinkImpl{
 				Interface: kubeClient.CoreV1().Events(""),
 			})
--- a/pkg/azuredisk/migration_monitor.go
+++ b/pkg/azuredisk/migration_monitor.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
 )
 
 const (
@@ -632,13 +633,13 @@ func (d *Driver) recoverMigrationMonitorsFromLabels(ctx context.Context) error {
 			toSKU := armcompute.DiskStorageAccountTypesPremiumV2LRS
 
 			if pv.Spec.CSI.VolumeAttributes != nil {
-				if sku, exists := pv.Spec.CSI.VolumeAttributes["storageAccountType"]; exists {
-					if sku != string(toSKU) {
+				if sku, exists := azureutils.ParseDiskParametersForKey(pv.Spec.CSI.VolumeAttributes, "storageAccountType"); exists {
+					if !strings.EqualFold(sku, string(toSKU)) {
 						continue
 					}
 				}
-				if sku, exists := pv.Spec.CSI.VolumeAttributes["skuName"]; exists {
-					if sku != string(toSKU) {
+				if sku, exists := azureutils.ParseDiskParametersForKey(pv.Spec.CSI.VolumeAttributes, "skuName"); exists {
+					if !strings.EqualFold(sku, string(toSKU)) {
 						continue
 					}
 				}
@@ -656,6 +657,6 @@ func (d *Driver) recoverMigrationMonitorsFromLabels(ctx context.Context) error {
 		}
 	}
 
-	klog.V(2).Infof("Recovered %d migration monitors from labels", recoveredCount)
+	klog.V(4).Infof("Recovered %d migration monitors from labels", recoveredCount)
 	return nil
 }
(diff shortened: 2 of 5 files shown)


pkg/azuredisk/azuredisk.go
```diff
@@ -302,8 +302,12 @@ func NewDriver(options *DriverOptions) *Driver {
 			// Recover any ongoing migrations after restart
 			go func() {
 				time.Sleep(30 * time.Second) // Wait for controller to fully start
-				if err := driver.recoverMigrationMonitorsFromLabels(context.Background()); err != nil {
-					klog.Errorf("Failed to recover migration monitors: %v", err)
+				// Periodically check every 10 minutes - in worst case kubernetes administrator can add the label to subscribe the events
+				for {
+					if err := driver.recoverMigrationMonitorsFromLabels(context.Background()); err != nil {
```
landreasyan, line 306: Could you add an option to enable and disable this feature? The monitor lists all the PVs on the cluster and this can cause client side throttling for larger scale clusters. 

Could you also see if the logging level could be increased for the logs in the monitor because I've noticed that it overwhelms the logs on larger clusters.

hasethuraman: > Could you add an option to enable and disable this feature?

Yes, enableMigrationMonitor is present and this code kick-in based on that.

>The monitor lists all the PVs on the cluster and this can cause client side throttling for larger scale clusters.

It does not, we have a label selector and there wont be more than 50 PVs that will be selected for migration

>Could you also see if the logging level could be increased for the logs in the monitor

There is only one log that get printed at the start up which I even bumped up now - but that will get logged only once per start

And remaining logs will happen only when migration is going on, do you see any migration?

landreasyan: The additional logs are coming from here independent on the migration status or progress: https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/c3aa6c8a25ceb748b6738ec6db9acbbde904f8f9/pkg/azuredisk/azuredisk.go#L292 

I think even 50 could cause significant throttling on larger clusters because we are not using a cache. Maybe switching to cached client would make sense.

hasethuraman: Thanks @landreasyan for providing, removed that line. It is not needed

hasethuraman:
@landreasyan if we are good, can we merge this?

pkg/azuredisk/migration_monitor.go
```diff
@@ -625,12 +625,25 @@ func (d *Driver) recoverMigrationMonitorsFromLabels(ctx context.Context) error {
...
 			// For recovery, we need to determine fromSKU and toSKU from the disk properties
-			klog.V(2).Infof("Recovering migration monitor for PV: %s", pv.Name)
+			klog.V(3).Infof("Recovering migration monitor for PV: %s", pv.Name)
 
 			// For now, we'll use placeholders - when more SKUs involve for migration, need to get this from disk properties
 			fromSKU := string(armcompute.DiskStorageAccountTypesPremiumLRS) // we may incorrectly say PremiumLRS instead of StandardZRS, lets keep it simple for now
 			toSKU := armcompute.DiskStorageAccountTypesPremiumV2LRS
 
+			if pv.Spec.CSI.VolumeAttributes != nil {
+				if sku, exists := pv.Spec.CSI.VolumeAttributes["storageAccountType"]; exists {
```
andyzhangx: both `storageAccountType` and `skuName` are not case sensitive

hasethuraman:
@andyzhangx @landreasyan any further concerns?

pkg/azureutils/azure_disk_utils.go
```diff
@@ -548,6 +548,19 @@ func ValidateDataAccessAuthMode(dataAccessAuthMode string) error {
 	return fmt.Errorf("dataAccessAuthMode(%s) is not supported", dataAccessAuthMode)
 }
 
+func ParseDiskParametersForKey(parameters map[string]string, key string) (string, bool) {
+	if parameters == nil {
```
andyzhangx: this is not necessary in for ... range since nil would be skipped in L556, other part LGTM, thanks.

hasethuraman: Removed it.

Copied it from ParseDiskParameters() :) I didnt remove from ParseDiskParameters & not sure of the impact as it is assigned to VolumeContext.

PR #3355, opened 2025-09-25
Check bd size before calling fs resize to avoid any potential fs corruption 

landreasyan:
**What type of PR is this?**
/kind bug

This PR addresses a critical issue where resize2fs may return success even when the underlying block device size is not properly updated, leading to silent filesystem corruption and blkid returning an empty result for the fs check. The fix introduces a pre-check using blockdev --getsize64 to ensure the device size reflects the expected expansion before invoking resize2fs. This validation step helps avoid resizing against stale or incorrect metadata, improving reliability and preventing data loss in local CSI environments.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx (approved):
/lgtm

PR #3081, opened 2025-05-07
feat: add option to override ARM retry-after header if the value is lower than the configured minimum

sunpa93:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Due to CRP extension bug, ARM polling GET responses are being given a high retry-after value after a certain number of retries. To mitigate the impact of the impact of this bug, this PR adds an option to override the retry-after header in http response when the value is lower than the configured minimum. This would allow the driver to poll for GET with a longer delay in between, lessening the frequency of large retry-after.

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:
- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #3358, opened 2025-09-27
fix: disable enable-minimum-retry-after by default

andyzhangx:
**What type of PR is this?**
/kind bug

fix: disable enable-minimum-retry-after by default

```
during the polling phase while waiting for CRP to finalize attaching disks, Azure tells us to retry  after X seconds (4/2/2/2), which is less than the expected 5 according to the logic. In doing so, we remove the header... And according to the Azure SDK, For long running operations and respective polling: [General Guidelines: API Design | Azure SDKs](https://azure.github.io/azure-sdk/general_design.html#long-running-operations).

DO support the following polling configuration options:

pollInterval
Polling configuration may be used only in the absence of relevant retry-after headers from service, and otherwise should be ignored.


And from the poller source code: [azure-sdk-for-go/sdk/azcore/runtime/poller.go at main · Azure/azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/poller.go)

[...]

type PollUntilDoneOptions struct {
// Frequency is the time to wait between polling intervals in absence of a Retry-After header. Allowed minimum is one second.
// Pass zero to accept the default value (30s).
Frequency time.Duration
}

// PollUntilDone will poll the service endpoint until a terminal state is reached, an error is received, or the context expires.
// It internally uses Poll(), Done(), and Result() in its polling loop, sleeping for the specified duration between intervals.
// options: pass nil to accept the default values.
// NOTE: the default polling frequency is 30 seconds which works well for most operations.  However, some operations might
// benefit from a shorter or longer duration.
func (p *Poller[T]) PollUntilDone(ctx context.Context, options *PollUntilDoneOptions) (res T, err error) {
if options == nil {
options = &PollUntilDoneOptions{}
}
cp := *options
if cp.Frequency == 0 {
cp.Frequency = 30 * time.Second
}

[...]

TLDR - CSI driver retryafter implementation removes the RetryAFter header, causing the default polling interval to fallback to 30 secs. This explains the delay you've witnessed in publishing the volume.
```

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azuredisk_option.go
+++ b/pkg/azuredisk/azuredisk_option.go
@@ -91,7 +91,7 @@ func (o *DriverOptions) AddFlags() *flag.FlagSet {
 	fs.StringVar(&o.UserAgentSuffix, "user-agent-suffix", "", "userAgent suffix")
 	fs.BoolVar(&o.UseCSIProxyGAInterface, "use-csiproxy-ga-interface", true, "boolean flag to enable csi-proxy GA interface on Windows")
 	fs.BoolVar(&o.EnableOtelTracing, "enable-otel-tracing", false, "If set, enable opentelemetry tracing for the driver. The tracing is disabled by default. Configure the exporter endpoint with OTEL_EXPORTER_OTLP_ENDPOINT and other env variables, see https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration.")
-	fs.BoolVar(&o.EnableMinimumRetryAfter, "enable-minimum-retry-after", true, "boolean flag to enable minimum retry after policy in azclient")
+	fs.BoolVar(&o.EnableMinimumRetryAfter, "enable-minimum-retry-after", false, "boolean flag to enable minimum retry after policy in azclient")
 	//only used in v1
 	fs.BoolVar(&o.EnableDiskOnlineResize, "enable-disk-online-resize", true, "boolean flag to enable disk online resize")
 	fs.BoolVar(&o.AllowEmptyCloudConfig, "allow-empty-cloud-config", true, "Whether allow running driver without cloud config")


landreasyan:
@andyzhangx, this was added specifically to address a latency bug from VMSS. 

VMSS estimates a retry after for single disk operations but its estimates are less accurate when multiple disks are involved. In such cases, VMSS would often return very short retry-after values (e.g., 2s), causing CSI to retry too quickly. After several retries, this would trigger a longer backoff (e.g., 37s) from VMSS, significantly increasing latency.

Sung Jun resolved this in this PR: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3081 by allowing CSI to respect its own retry-after logic when VMSS returns unusually low values. This prevents excessive GET operations and avoids hitting the 37s backoff threshold.
This change has been confirmed to reduce overall disk operation latency by avoiding unnecessary retries and long backoffs.

andyzhangx:
> @andyzhangx, this was added specifically to address a latency bug from VMSS.
> 
> VMSS estimates a retry after for single disk operations but its estimates are less accurate when multiple disks are involved. In such cases, VMSS would often return very short retry-after values (e.g., 2s), causing CSI to retry too quickly. After several retries, this would trigger a longer backoff (e.g., 37s) from VMSS, significantly increasing latency.
> 
> Sung Jun resolved this in this PR: #3081 by allowing CSI to respect its own retry-after logic when VMSS returns unusually low values. This prevents excessive GET operations and avoids hitting the 37s backoff threshold. This change has been confirmed to reduce overall disk operation latency by avoiding unnecessary retries and long backoffs.

@landreasyan the original issue and workaround is due to a CRP extension bug, and that bug was fixed and rolled out complete a few months ago, we should avoid using such workaround (removing the header) since this CRP extension bug is already fixed, otherwise it would create extra issue as mentioned in this PR description.

PR #3379, opened 2025-10-07
fix: create volume failure when requested pvc size is smaller than snapshot disk size

priyansh17:
In cases where the requested size is lesser than snapshot size, we want the Disk RP to handle the size automatically from snapshot.
This is the process we use in our backend which has worked seamlessly for years. We want to integrate the same behavior here.

/kind bug

Currently if the snapshot size is more than the requested size in PVC it causes failure to provision the Disk, this change will help to check if snapshot size is larger than requested size in PVC, provision disk based on that size.

Example:
```
PVC.YAML:
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pv2-pvc-restored
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: managed-csi-premium-v2
  resources:
    requests:
      storage: 1024Gi
  dataSource:
    name: pv2-restore-vs
    kind: VolumeSnapshot
    apiGroup: snapshot.storage.k8s.io
```

Here the Volume Snapshot referred if has size of 1024.25Gi we would need a disk of exactly 1024.25Gi, instead current flow fails to do so as it tries to provision it for 1024Gi.

Adding refernece to the 256MiB Increment Behaviour.

image

https://learn.microsoft.com/en-us/azure/virtual-machines/windows/disks-upload-vhd-to-managed-disk-powershell#create-an-empty-managed-disk

- [ ] includes documentation
- [X] adds unit tests
- [X] tested upgrade from previous version

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azure_managedDiskController.go
+++ b/pkg/azuredisk/azure_managedDiskController.go
@@ -34,6 +34,7 @@ import (
 	volumehelpers "k8s.io/cloud-provider/volume/helpers"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
+	csidriverconsts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 
 	azureconsts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
@@ -153,10 +154,20 @@ func (c *ManagedDiskController) CreateManagedDisk(ctx context.Context, options *
 	if err != nil {
 		return "", err
 	}
-	diskProperties := armcompute.DiskProperties{
-		DiskSizeGB:      &diskSizeGB,
-		CreationData:    &creationData,
-		BurstingEnabled: options.BurstingEnabled,
+
+	diskProperties := armcompute.DiskProperties{}
+	// when creating from snapshot, and diskSizeGB is 0, let disk RP calculate size from snapshot bytes size.
+	if diskSizeGB == 0 && options.SourceType == csidriverconsts.SourceSnapshot {
+		diskProperties = armcompute.DiskProperties{
+			CreationData:    &creationData,
+			BurstingEnabled: options.BurstingEnabled,
+		}
+	} else {
+		diskProperties = armcompute.DiskProperties{
+			CreationData:    &creationData,
+			BurstingEnabled: options.BurstingEnabled,
+			DiskSizeGB:      &diskSizeGB,
+		}
 	}
 
 	if options.PublicNetworkAccess != "" {
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -49,10 +49,11 @@ import (
 )
 
 const (
-	waitForSnapshotReadyInterval = 5 * time.Second
-	waitForSnapshotReadyTimeout  = 10 * time.Minute
-	maxErrMsgLength              = 990
-	checkDiskLunThrottleLatency  = 1 * time.Second
+	waitForSnapshotReadyInterval     = 5 * time.Second
+	waitForSnapshotReadyTimeout      = 10 * time.Minute
+	maxErrMsgLength                  = 990
+	checkDiskLunThrottleLatency      = 1 * time.Second
+	maxSnapshotSizeDifferenceAllowed = 50 // in GiB
 )
 
 // listVolumeStatus explains the return status of `listVolumesByResourceGroup`
@@ -196,6 +197,7 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 
 	capacityBytes := req.GetCapacityRange().GetRequiredBytes()
 	volSizeBytes := int64(capacityBytes)
+	requestSizeToBeSupplied := true
 	requestGiB := int(volumehelper.RoundUpGiB(volSizeBytes))
 
 	if diskParams.PerformancePlus != nil && *diskParams.PerformancePlus && requestGiB < consts.PerformancePlusMinimumDiskSizeGiB {
@@ -314,11 +316,31 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 					},
 				},
 			}
-			// if migration monitoring is enabled and if target SKU is PremiumV2LRS, get the snapshot SKU
-			// the snapshot SKU is marked as source SKU
-			if skuName == armcompute.DiskStorageAccountTypesPremiumV2LRS && d.migrationMonitor != nil {
-				sourceSKU, _ = d.getSnapshotSKU(ctx, sourceID)
+
+			// Get snapshot once and extract both SKU and disk size info
+			snapshot, err := d.getSnapshot(ctx, sourceID)
+			if err == nil {
+				// fetch snapshot info and compare size with requested size
+				// when snapshot size is larger than requested size, do not supply the size in the request
+				// to allow Azure to create a disk with exact snapshot size in bytes.
+				diskSizeInBytes, err := getDiskSizeInBytesFromSnapshot(snapshot)
+				if err == nil {
+					requestedGiBfromSnapshot := int(volumehelper.RoundUpGiB(diskSizeInBytes))
+					differenceSize := requestedGiBfromSnapshot - requestGiB
+					if requestedGiBfromSnapshot > requestGiB && differenceSize <= maxSnapshotSizeDifferenceAllowed {
+						klog.V(4).Infof("snapshot size (%d GiB) is larger than requested size (%d GiB) but difference (%d GiB) is within the allowed limit (%d GiB), will not supply the size in the create disk request", requestedGiBfromSnapshot, requestGiB, differenceSize, maxSnapshotSizeDifferenceAllowed)
+						requestSizeToBeSupplied = false
+					}
+				}
+
+				// Get SKU if migration monitoring is enabled and target SKU is PremiumV2LRS
+				if skuName == armcompute.DiskStorageAccountTypesPremiumV2LRS && d.migrationMonitor != nil {
+					sourceSKU, _ = getSnapshotSKUFromSnapshot(snapshot)
+				}
+			} else {
+				return nil, status.Errorf(codes.NotFound, "%v", err)
 			}
+
 			metricsRequest = "controller_create_volume_from_snapshot"
 		} else {
 			sourceID = content.GetVolume().GetVolumeId()
@@ -392,6 +414,11 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}
 
 	diskParams.VolumeContext[consts.RequestedSizeGib] = strconv.Itoa(requestGiB)
+
+	if !requestSizeToBeSupplied && sourceType == consts.SourceSnapshot {
+		requestGiB = 0
+	}
+
 	volumeOptions := &ManagedDiskOptions{
 		AvailabilityZone:    diskZone,
 		BurstingEnabled:     diskParams.EnableBursting,
@@ -1441,26 +1468,48 @@ func (d *Driver) GetSourceDiskSize(ctx context.Context, subsID, resourceGroup, d
 	return (*result.Properties).DiskSizeGB, result, nil
 }
 
-// getSnapshotSKU retrieves the SKU of the snapshot and returns the SKU or if any error occurs
-func (d *Driver) getSnapshotSKU(ctx context.Context, sourceID string) (string, error) {
+// getSnapshot retrieves the Snapshot and returns the Snapshot or the error if any error occurs
+func (d *Driver) getSnapshot(ctx context.Context, sourceID string) (*armcompute.Snapshot, error) {
 	subsID, resourceGroup, snapshotName, err := azureutils.GetInfoFromURI(sourceID)
 	if err != nil {
 		klog.Warningf("could not get subscription id, resource group from snapshot uri (%s) with error(%v)", sourceID, err)
-		return "", err
+		return nil, err
 	}
 	snapClient, err := d.clientFactory.GetSnapshotClientForSub(subsID)
 	if err != nil {
 		klog.Warningf("could not get snapshot client for subscription(%s) with error(%v)", subsID, err)
-		return "", err
+		return nil, err
 	}
-	result, err := snapClient.Get(ctx, resourceGroup, snapshotName)
+	snapshotRetrieved, err := snapClient.Get(ctx, resourceGroup, snapshotName)
 	if err != nil {
 		klog.Warningf("get snapshot %s from rg(%s) error: %v", snapshotName, resourceGroup, err)
-		return "", err
+		return nil, err
 	}
-	if result == nil || result.SKU == nil || result.SKU.Name == nil {
-		klog.Warningf("Snapshot or Snapshot property not found for snapshot (%s) in resource group (%s)", snapshotName, resourceGroup)
-		return "", status.Error(codes.NotFound, fmt.Sprintf("Snapshot or Snapshot property not found for snapshot (%s) in resource group (%s)", snapshotName, resourceGroup))
+	return snapshotRetrieved, nil
+}
+
+// getSnapshotSKU retrieves the SKU of the snapshot and returns the SKU or if any error occurs
+func getSnapshotSKUFromSnapshot(computeSnapshot *armcompute.Snapshot) (string, error) {
+	if computeSnapshot == nil {
+		klog.Warningf("Snapshot is nil")
+		return "", status.Error(codes.NotFound, "Snapshot is nil")
 	}
-	return string(*result.SKU.Name), nil
+	if computeSnapshot.SKU == nil || computeSnapshot.SKU.Name == nil {
+		klog.Warningf("Snapshot or Snapshot Properties SKU not found for snapshot")
+		return "", status.Error(codes.NotFound, "Snapshot SKU property not found")
+	}
+	return string(*computeSnapshot.SKU.Name), nil
+}
+
+// getDiskSizeInBytes retrieves the size of the disk and returns the size or if any error occurs
+func getDiskSizeInBytesFromSnapshot(computeSnapshot *armcompute.Snapshot) (int64, error) {
+	if computeSnapshot == nil {
+		klog.Warningf("Snapshot is nil")
+		return 0, status.Error(codes.NotFound, "Snapshot is nil")
+	}
+	if computeSnapshot.Properties == nil || computeSnapshot.Properties.DiskSizeBytes == nil {
+		klog.Warningf("Snapshot or Snapshot Properties.DiskSizeBytes not found for snapshot")
+		return 0, status.Error(codes.NotFound, "Snapshot size not found")
+	}
+	return *computeSnapshot.Properties.DiskSizeBytes, nil
 }
--- a/pkg/azuredisk/fake_azuredisk.go
+++ b/pkg/azuredisk/fake_azuredisk.go
@@ -96,7 +96,7 @@ type FakeDriver interface {
 	checkDiskExists(ctx context.Context, diskURI string) (*armcompute.Disk, error)
 	waitForSnapshotReady(context.Context, string, string, string, time.Duration, time.Duration) error
 	getSnapshotByID(context.Context, string, string, string, string) (*csi.Snapshot, error)
-	getSnapshotSKU(context.Context, string) (string, error)
+	getSnapshot(context.Context, string) (*armcompute.Snapshot, error)
 	ensureMountPoint(string) (bool, error)
 	ensureBlockTargetFile(string) error
 	getDevicePathWithLUN(lunStr string) (string, error)
--- a/pkg/azuredisk/controllerserver_test.go
+++ b/pkg/azuredisk/controllerserver_test.go
@@ -354,7 +354,9 @@ func TestCreateVolume(t *testing.T) {
 					SKU: &armcompute.SnapshotSKU{
 						Name: ptr.To(armcompute.SnapshotStorageAccountTypesStandardZRS),
 					},
-					Properties: &armcompute.SnapshotProperties{},
+					Properties: &armcompute.SnapshotProperties{
+						DiskSizeBytes: ptr.To(int64(1073741824)), // 1GB in bytes
+					},
 				}, nil).AnyTimes()
 				_, err := d.CreateVolume(context.Background(), req)
 				expectedErr := status.Errorf(codes.Internal, "test")
@@ -381,7 +383,9 @@ func TestCreateVolume(t *testing.T) {
 					VolumeContentSource: &volumecontensource,
 				}
 				disk := &armcompute.Disk{
-					Properties: &armcompute.DiskProperties{},
+					Properties: &armcompute.DiskProperties{
+						DiskSizeBytes: ptr.To(int64(1073741824)), // 1GB in bytes
+					},
 				}
 				diskClient := mock_diskclient.NewMockInterface(cntl)
 				d.getClientFactory().(*mock_azclient.MockClientFactory).EXPECT().GetDiskClientForSub(gomock.Any()).Return(diskClient, nil).AnyTimes()
@@ -684,6 +688,7 @@ func TestCreateVolume_SnapshotPremiumLRS_ToPremiumV2_EmitsMigrationEvents(t *tes
 		},
 		Properties: &armcompute.SnapshotProperties{
 			DiskSizeGB:        &sizeGB,
+			DiskSizeBytes:     ptr.To(int64(1073741824)), // 1GB in bytes
 			ProvisioningState: &provState,
 		},
 	}, nil).AnyTimes()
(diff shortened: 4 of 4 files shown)


andyzhangx (commented):
/ok-to-test

pkg/azuredisk/controllerserver.go
```diff
@@ -320,6 +320,23 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 				sourceSKU, _ = d.getSnapshotSKU(ctx, sourceID)
 			}
 			metricsRequest = "controller_create_volume_from_snapshot"
+
+			// Fetch snapshot info and check size
+			subsID, resourceGroup, snapshotName, err := azureutils.GetInfoFromURI(sourceID)
```
andyzhangx: can you leverage`getSnapshotSKU` to get snapshot size? and rename as `getSnapshotInfo`

andyzhangx (commented):
pls sign the cla first (select individual contributor)

pkg/azuredisk/controllerserver.go
priyansh17: Sure, updated for the same.

priyansh17:
@andyzhangx I have called out the issue in description. Let me know if you don't agree to this approach.

Other alternatives to this approach could be:
- supplying the size in bytes instead of GiB. (Disk RP Supports that argument)

pkg/azuredisk/controllerserver.go
```diff
@@ -314,11 +315,24 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
 			// if migration monitoring is enabled and if target SKU is PremiumV2LRS, get the snapshot SKU
 			// the snapshot SKU is marked as source SKU
 			if skuName == armcompute.DiskStorageAccountTypesPremiumV2LRS && d.migrationMonitor != nil {
 				sourceSKU, _ = d.getSnapshotSKU(ctx, sourceID)
 			}
+
+			// fetch snapshot info and compare disk size with requested size
+			// when disk size is larger than requested size, do not supply the size in the request
+			// to allow Azure to create a disk with exact disk size in bytes from snapshot.
+			diskSizeInBytes, err := d.getDiskSizeInBytes(ctx, sourceID)
```
andyzhangx: both `d.getSnapshotSKU` and `d.getDiskSizeInBytes` would call `d.getComputeSnapshot`, that would make extra get snapshot call, pls combine these two functions to make only one call, e.g. `getSnapshotInfo`

priyansh17: Sure, I was going with that flow initially where we make one call to get info and extract fields according to need.

Are you fine with this overall approach to handle sizes for creation via snapshot.

andyzhangx: are there other options?

andyzhangx: >supplying the size in bytes instead of GiB. (Disk RP Supports that argument)

this option does not make much difference

priyansh17: Yes correct, if we don't make this change, we are falling into problem during AKS Restores.

I verified that for Disk workload we don't supply this size when option is copy via snapshot. So, this should work seamlessly.

pkg/azuredisk/controllerserver.go
```diff
@@ -1441,26 +1459,48 @@ func (d *Driver) GetSourceDiskSize(ctx context.Context, subsID, resourceGroup, d
...
 	if err != nil {
 		klog.Warningf("get snapshot %s from rg(%s) error: %v", snapshotName, resourceGroup, err)
-		return "", err
+		return nil, err
+	}
+	return snapshotRetrieved, nil
+}
+
+// getSnapshotSKU retrieves the SKU of the snapshot and returns the SKU or if any error occurs
+func (d *Driver) getSnapshotSKUFromSnapshot(ctx context.Context, computeSnapshot *armcompute.Snapshot) (string, error) {
```
andyzhangx: fix the golint error and add ut for this func

pkg/azuredisk/controllerserver.go
```diff
@@ -1441,26 +1459,48 @@ func (d *Driver) GetSourceDiskSize(ctx context.Context, subsID, resourceGroup, d
 	return (*result.Properties).DiskSizeGB, result, nil
 }
 
-// getSnapshotSKU retrieves the SKU of the snapshot and returns the SKU or if any error occurs
-func (d *Driver) getSnapshotSKU(ctx context.Context, sourceID string) (string, error) {
+// getSnapshot retrieves the Snapshot and returns the Snapshot or the error if any error occurs
```
andyzhangx, line 1471: add ut

hasethuraman:
@priyansh17 when I tried to create a premium v2 disk from the existing snapshot through azure cli, I dont see that mentioned .25GiB bump & it must be the same for CSI flow too. This .25GiB is something will be specific to upload option. Do you see that applies for CSI flow? Not sure if tweaking the PVC request is fine because it could be the user input & failing is the best way for user to decide in case if that chosen snapshot could be a wrong one (and if higher, we may end up creating a higher one). Should this be selectively enforced rather being a default behaviour?

andyzhangx:
> @priyansh17 when I tried to create a premium v2 disk from the existing snapshot through azure cli, I dont see that mentioned .25GiB bump & it must be the same for CSI flow too. This .25GiB is something will be specific to upload option. Do you see that applies for CSI flow? Not sure if tweaking the PVC request is fine because it could be the user input & failing is the best way for user to decide in case if that chosen snapshot could be a wrong one (and if higher, we may end up creating a higher one). Should this be selectively enforced rather being a default behaviour?

that's good point, what's the error msg now when this issue happen?

hasethuraman:
> > @priyansh17 when I tried to create a premium v2 disk from the existing snapshot through azure cli, I dont see that mentioned .25GiB bump & it must be the same for CSI flow too. This .25GiB is something will be specific to upload option. Do you see that applies for CSI flow? Not sure if tweaking the PVC request is fine because it could be the user input & failing is the best way for user to decide in case if that chosen snapshot could be a wrong one (and if higher, we may end up creating a higher one). Should this be selectively enforced rather being a default behaviour?
> 
> that's good point, what's the error msg now when this issue happen?

@priyansh17 could you share the failure log/event?

priyansh17:
> > > @priyansh17 when I tried to create a premium v2 disk from the existing snapshot through azure cli, I dont see that mentioned .25GiB bump & it must be the same for CSI flow too. This .25GiB is something will be specific to upload option. Do you see that applies for CSI flow? Not sure if tweaking the PVC request is fine because it could be the user input & failing is the best way for user to decide in case if that chosen snapshot could be a wrong one (and if higher, we may end up creating a higher one). Should this be selectively enforced rather being a default behaviour?
> > 
> > 
> > that's good point, what's the error msg now when this issue happen?
> 
> @priyansh17 could you share the failure log/event?

Hello, let me explain with an example how it affects the CSI flow.

We have 2 scenarios: Operational backups and Vaulted backups.
Operational backups use the flow where you take a snapshot store it and during restore use the same snapshot URL as source in PVC.
In Vaulted backups we store them as blobs in vaults, and during restore re-create disks using Upload option and then follow the same operational flow as above.

Suppose you have backed up the PVCs using Azure AKS Backup solution, if the backups are vaulted and stored as pageblobs, when converted back to disks we notice the size of these disks are 256MiB more than original. 

Note** this is only in cases of Managed disks which have **Logical sector size of 4096**.

Let the original disk be of **100GiB. (1,07,37,41,82,400 Bytes)**
Post restores the size is **100.25GiB or (1,07,64,26,17,856 Bytes)** 

Now when requesting PV your PVC would be referring to this disk's snapshot URL, and the snapshot size is also 100.25GiB. 
Here we notice that CSI throws the error to create disks with size mismatch.

This error is same for **CLI and Portal** as well not just CSI Driver if you mentioned requested size less than the snapshot's size.
Attaching the error at the end.


> Not sure if tweaking the PVC request is fine because it could be the user input & failing is the best way for user to decide in case if that chosen snapshot could be a wrong one (and if higher, we may end up creating a higher one). Should this be selectively enforced rather being a default behavior?

Regarding this point, we are only using the behavior of not supplying size and auto-adjusting when user has mentioned they want the PV DataSource from the snapshot and there's a mismatch.






Error from events:
```
Name:          premium-v2-4k-pvc
Namespace:     dd4kd
StorageClass:  managed-csi-premium-v2-4k
Status:        Pending
Volume:        
Labels:        velero.io/backup-name=bkp.f2b012c9-033a-47d4-ba0c-3361d7c107bd.202509290001335090106
               velero.io/restore-name=res.693fc9e4-97f1-4a06-aa4d-dc98c71a0063.202509300849392616914
               velero.io/volume-snapshot-name=velero-premium-v2-4k-pvc-5dhp9
Annotations:   backup.velero.io/must-include-additional-items: true
               volume.beta.kubernetes.io/storage-provisioner: disk.csi.azure.com
               volume.kubernetes.io/selected-node: aks-agentpool-27570098-vmss000001
               volume.kubernetes.io/storage-provisioner: disk.csi.azure.com
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      
Access Modes:  
VolumeMode:    Filesystem
DataSource:
  APIGroup:  snapshot.storage.k8s.io
  Kind:      VolumeSnapshot
  Name:      velero-premium-v2-4k-pvc-5dhp9
Used By:     premium-v2-4k-pod
Events:
  Type     Reason                Age                 From                                                                                              Message
  ----     ------                ----                ----                                                                                              -------
  Normal   WaitForFirstConsumer  16m (x2 over 16m)   persistentvolume-controller                                                                       waiting for first consumer to be created before binding
  Warning  ProvisioningFailed    11m (x10 over 16m)  disk.csi.azure.com_csi-azuredisk-controller-c7c75d6dc-nhbfl_4e864351-ced6-499f-896e-581391b7fcb6  failed to provision volume with StorageClass "managed-csi-premium-v2-4k": rpc error: code = Internal desc = PUT http://localhost:7788/subscriptions/2c6832bf-90ef-457a-a663-9aeae88a7f80/resourceGroups/MC_priyansh-rg_priyansh-k8s_eastasia/providers/Microsoft.Compute/disks/pvc-54c5185b-12ab-47e0-8a99-83d48186e594
--------------------------------------------------------------------------------
RESPONSE 200: 200 OK
ERROR CODE: InvalidVHDXFile
--------------------------------------------------------------------------------
{
  "startTime": "2025-09-30T09:02:43.330425+00:00",
  "endTime": "2025-09-30T09:02:49.6274452+00:00",
  "status": "Failed",
  "error": {
    "code": "InvalidVHDXFile",
    "message": "Source is not a valid VHDX file."
  },
  "name": "ee7119f9-bad2-4531-a66b-c0ac9dcdfe5a"
}
--------------------------------------------------------------------------------
  Normal  ExternalProvisioning  77s (x63 over 16m)  persistentvolume-controller  Waiting for a volume to be created either by the external provisioner 'disk.csi.azure.com' or manually by the system administrator. If volume creation is delayed, please verify that the provisioner is running and correctly registered.
  ```

File: pkg/azuredisk/controllerserver.go (hunk 4 above)
Comment (priyansh17) [reply pkg/azuredisk/controllerserver.go:1471]: Added UTs

File: pkg/azuredisk/controllerserver.go (hunk 3 above)
Comment (priyansh17) [reply pkg/azuredisk/controllerserver.go]: Fixed and Added UTs

**Comment by priyansh17** [comment]
> pls also squash all commits by `git rebase -i HEAD~12` and force push the commits, thx

Updated to squash all commits into 1, thanks.

**Comment by priyansh17** [comment]
@k8s-ci-robot /retest

File: pkg/azuredisk/controllerserver.go [hunk 5]
```diff
@@ -314,11 +315,25 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+
+			// Get snapshot once and extract both SKU and disk size info
+			snapshot, err := d.getSnapshot(ctx, sourceID)
+			if err == nil {
+				// Always get disk size for size comparison
+				diskSizeInBytes, err := d.getDiskSizeInBytesFromSnapshot(snapshot)
+				if err == nil {
+					requestedGiBfromSnapshot := int(volumehelper.RoundUpGiB(diskSizeInBytes))
+					if requestedGiBfromSnapshot > requestGiB {
+						requestSizeToBeSupplied = false
```
andyzhangx, line 332: add a logging here, `klog.V(4).Infof(..` to log the requestedGiBfromSnapshot and requestGiB in logs

pkg/azuredisk/controllerserver.go
```diff
@@ -314,11 +315,25 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
-				sourceSKU, _ = d.getSnapshotSKU(ctx, sourceID)
+
+			// Get snapshot once and extract both SKU and disk size info
+			snapshot, err := d.getSnapshot(ctx, sourceID)
+			if err == nil {
+				// Always get disk size for size comparison
+				diskSizeInBytes, err := d.getDiskSizeInBytesFromSnapshot(snapshot)
+				if err == nil {
+					requestedGiBfromSnapshot := int(volumehelper.RoundUpGiB(diskSizeInBytes))
+					if requestedGiBfromSnapshot > requestGiB {
```
hasethuraman: > Regarding this point, we are only using the behavior of not supplying size and auto-adjusting when user has mentioned they want the PV DataSource from the snapshot and there's a mismatch.

@priyansh17 Along with the check can we see if requestGiB is not 1GB/some threshold short of snapshot size? This will ensure that we dont end up creating a larger disk when users use incorrect snapshot?

Also, what would happen when this cycle repeats, we end up accumulating 250MiB every backup restore flow?

pkg/azuredisk/azure_managedDiskController.go
```diff
@@ -153,12 +154,19 @@ func (c *ManagedDiskController) CreateManagedDisk(ctx context.Context, options *
...
 	diskProperties := armcompute.DiskProperties{
-		DiskSizeGB:      &diskSizeGB,
 		CreationData:    &creationData,
 		BurstingEnabled: options.BurstingEnabled,
 	}
 
+	// Supply size if sourceType is not snapshot or
+	// if sourceType is snapshot and diskSizeGB is not 0.
+	// When creating from snapshot with size 0, let disk RP calculate size from snapshot bytes size.
+	if options.SourceType != csidriverconsts.SourceSnapshot || diskSizeGB != 0 {
```
mayankagg9722: we can remove "options.SourceType != csidriverconsts.SourceSnapshot" as you have already handled this below while calculating the size.

pkg/azuredisk/controllerserver.go
priyansh17: Yes, sure we can have that change to check if requestGiB is in 1GiB threshold, I was thinking of this initially, but it seemed too specific of a case to be generalized. @andyzhangx do you agree with this?

 For the second point, our understanding is that restore is a rare operation. backup is frequent. Hence it should not affect much, but going with this flow it would take 4 restores and backup cycles to increment the size by 1GiB. 
 
 Disk D1 (x bytes)
 1st Backup- > Snapshot S1 
 Vaulted Restore of S1 -> D2 (x bytes+ 256 MiB)
 2nd Backup -> S2
  Vaulted Restore of S2 -> D3 (x bytes+ 512 MiB)
  ..
  ..
  ..
  Vaulted Restore of S4 -> D5 (x bytes + 1 GiB)
  
  
  When user is shown the disk on portal in GiB, it rounds it down (i.e. 100.25 is shown as 100 GiB) 
  So, until 4th Vaulted restore (which is extremely rare cycle) he would see same size.

pkg/azuredisk/azure_managedDiskController.go
priyansh17: Hi Mayank, giving second thought to this I believe we should keep this check. 

In a flow where diskSizeGB alone is considered we have no way to tell if we wanted to ignore it or not. i.e. if we had made it 0 or it was supplied 0 from the start.

priyansh17:
> pls submit a PR against master branch, and then we could cherrypick to release-1.33 automatically. usually you don't need to submit a PR against release-xx branch unless automatic cherrypick does not work.

Sure, thank you, will update target branch.

pkg/azuredisk/controllerserver.go
priyansh17: Added logger

pkg/azuredisk/controllerserver.go
andyzhangx: the 5th snapshot would be 1.25GB bigger than original disk size, right?
if we only check requestGiB is in 1GiB threshold,  it may exceeds 1GB when lots of snapshots are made?

priyansh17: No, we will check in comparison to D4 when making D5. so it would still be 256MiB larger only.

As during this restore backed up disk was D4 not D1.

priyansh17:
@k8s-ci-robot /retest

pkg/azuredisk/controllerserver.go
hasethuraman: Thanks @priyansh17 for laying out the example

Would not the restore using S4 will create the disk size equal to x+1GiB (where x is D1 size)?

hasethuraman: If yes, can we skip the previous 256MiB when new one is backed up (of disk that was restored), so that we ensure it is 256MiB when there is a backup, restore loop. Since you agreed to the other part of my first comment, this can go in subsequent iteration IMO

priyansh17: Hello @hasethuraman, Yes restore from S4 will create a size of x+1GiB.

But my point here was D1 is nowhere to be found now, the customer was using D4 when he took backup and restored, and D4 was already x + 768MiB. So, the diff is still of 256MiB.
D4 now needs to be looked at as original disk for this restore.

hasethuraman: Agreed @priyansh17, but the size is slowly growing from 0 to 1GiB in this example. We will track this & revisit later since restore using upload will be infrequent operation. The CSI create from snapshot is anyway not blocked.

hasethuraman: @priyansh17 for my first comment - is it possible to add a parameter in StorageClass in restore path and enforce selectively?

hasethuraman: Also @priyansh17 will this 256MiB data after restoration will be visible to process?

priyansh17: Hello, to answer your questions:

> Agreed @priyansh17, but the size is slowly growing from 0 to 1GiB in this example. We will track this & revisit later since restore using upload will be infrequent operation. The CSI create from snapshot is anyway not blocked.

I would suggest let us take this PR in with the suggestion you made of comparing sizes to be in difference of 1 GiB. This should not cause any harm and make our Cross regions restores / any kind of vaulted restores work as required.
Do you see any harm/risks in not taking this now?

> is it possible to add a parameter in StorageClass in restore path and enforce selectively?

As of today, I do not believe this can be part of storage class to enforce selectively, as we do not manipulate any K8s resources. we create the YAMLs as-it-is / as-it-was when backup was taken. 
Hence the need is for us to handle how we call the create disk API.

> will this 256MiB data after restoration will be visible to process?

Yes, if the user checks the size of disk post restore which is attached to his PV it would be +256MiB from the original, this we will document from our side.

hasethuraman: >Yes, if the user checks the size of disk post restore which is attached to his PV it would be +256MiB from the original, this we will document from our side.

Sure. But what about the 256MiB visibility to the process or application?

priyansh17: We have added logs to see if there was a mismatch and we had relied on Snapshot size. That is one place where there is visibility to process. 

Maybe I am misunderstanding the question, by visibility to process what does it mean?

hasethuraman: Basically, the pod in kubernetes will be able to view, modify this 256MiB?

pkg/azuredisk/controllerserver.go
```diff
@@ -314,11 +316,31 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+			snapshot, err := d.getSnapshot(ctx, sourceID)
+			if err == nil {
+				// fetch snapshot info and compare size with requested size
+				// when snapshot size is larger than requested size, do not supply the size in the request
+				// to allow Azure to create a disk with exact snapshot size in bytes.
+				diskSizeInBytes, err := d.getDiskSizeInBytesFromSnapshot(snapshot)
+				if err == nil {
+					requestedGiBfromSnapshot := int(volumehelper.RoundUpGiB(diskSizeInBytes))
+					differenceSize := requestedGiBfromSnapshot - requestGiB
+					if requestedGiBfromSnapshot > requestGiB && differenceSize < maxSnapshotSizeDifferenceAllowed {
```
priyansh17: <=

pkg/azuredisk/azure_managedDiskController.go
```diff
@@ -154,11 +155,17 @@ func (c *ManagedDiskController) CreateManagedDisk(ctx context.Context, options *
 		return "", err
 	}
 	diskProperties := armcompute.DiskProperties{
-		DiskSizeGB:      &diskSizeGB,
 		CreationData:    &creationData,
 		BurstingEnabled: options.BurstingEnabled,
 	}
 
+	// Supply size if sourceType is not snapshot or
```
priyansh17: make this more readable

pkg/azuredisk/controllerserver.go
hasethuraman: Lets say when you have 10Gi, with that flow, it becomes 10.25Gi, would like to know the disk size from the pod and the checksum of last 256Mi if it has data or zeroes to ensure if we are good with above changes

priyansh17: As checked and confirmed with team, the data is zeroed and would not be modified.

pkg/azuredisk/controllerserver.go
```diff
@@ -49,10 +49,11 @@ import (
...
 const (
-	waitForSnapshotReadyInterval = 5 * time.Second
-	waitForSnapshotReadyTimeout  = 10 * time.Minute
-	maxErrMsgLength              = 990
-	checkDiskLunThrottleLatency  = 1 * time.Second
+	waitForSnapshotReadyInterval     = 5 * time.Second
+	waitForSnapshotReadyTimeout      = 10 * time.Minute
+	maxErrMsgLength                  = 990
+	checkDiskLunThrottleLatency      = 1 * time.Second
+	maxSnapshotSizeDifferenceAllowed = 50 // in GiB
```
priyansh17: Please review this @anshulahuja98 & @hasethuraman

anshulahuja98:
The overall change LGTM has been discussed with partner teams. Requested priyansh to add link to public docs of the 256mb round up in the description.

pkg/azuredisk/controllerserver.go
hasethuraman: LGTM

priyansh17:
@andyzhangx could you do a final review please. got it signed off from others.

pkg/azuredisk/controllerserver.go
```diff
@@ -1441,26 +1468,48 @@ func (d *Driver) GetSourceDiskSize(ctx context.Context, subsID, resourceGroup, d
...
 	if err != nil {
 		klog.Warningf("get snapshot %s from rg(%s) error: %v", snapshotName, resourceGroup, err)
-		return "", err
+		return nil, err
+	}
+	return snapshotRetrieved, nil
+}
+
+// getSnapshotSKU retrieves the SKU of the snapshot and returns the SKU or if any error occurs
+func (d *Driver) getSnapshotSKUFromSnapshot(computeSnapshot *armcompute.Snapshot) (string, error) {
```
andyzhangx: this should not be a function of `d *Driver`, it's a util func

pkg/azuredisk/controllerserver.go
```diff
@@ -1441,26 +1468,48 @@ func (d *Driver) GetSourceDiskSize(ctx context.Context, subsID, resourceGroup, d
...
+	}
+	if computeSnapshot.SKU == nil || computeSnapshot.SKU.Name == nil {
+		klog.Warningf("Snapshot or Snapshot Properties SKU not found for snapshot")
+		return "", status.Error(codes.NotFound, "Snapshot SKU property not found")
+	}
+	return string(*computeSnapshot.SKU.Name), nil
+}
+
+// getDiskSizeInBytes retrieves the size of the disk and returns the size or if any error occurs
+func (d *Driver) getDiskSizeInBytesFromSnapshot(computeSnapshot *armcompute.Snapshot) (int64, error) {
```
andyzhangx: this should not be a function of `d *Driver`, it's a util func

pkg/azuredisk/controllerserver.go
priyansh17: Updated

pkg/azuredisk/controllerserver.go
priyansh17: Updated

pkg/azuredisk/fake_azuredisk.go
```diff
@@ -96,7 +96,7 @@ type FakeDriver interface {
 	checkDiskExists(ctx context.Context, diskURI string) (*armcompute.Disk, error)
 	waitForSnapshotReady(context.Context, string, string, string, time.Duration, time.Duration) error
 	getSnapshotByID(context.Context, string, string, string, string) (*csi.Snapshot, error)
-	getSnapshotSKU(context.Context, string) (string, error)
+	getSnapshot(context.Context, string) (*armcompute.Snapshot, error)
```
priyansh17, line 99: do you want me to rename this as it's similar to getSnapshotByID in name. @andyzhangx

hasethuraman:
@priyansh17 this may need to be cherry-picked to release-1.33 branch & any backport required?

CC: @andyzhangx

priyansh17:
@hasethuraman / @andyzhangx we were supporting the CSI Snapshot handle change from 1.32, right? could we backport this as well to 1.32.

andyzhangx:
> @hasethuraman / @andyzhangx we were supporting the CSI Snapshot handle change from 1.32, right? could we backport this as well to 1.32.

@priyansh17 auto cherrypick won't work on release-1.32 since there is conflict, and you submit a PR against release-1.32 manually? thx

priyansh17:
> > @hasethuraman / @andyzhangx we were supporting the CSI Snapshot handle change from 1.32, right? could we backport this as well to 1.32.
> 
> @priyansh17 auto cherrypick won't work on release-1.32 since there is conflict, and you submit a PR against release-1.32 manually? thx

Yes, Sure. will update.

Issue #3448, opened 2025-12-16
Static provision crashed caused by segmentation fault on Azure stack hub clusters


Phaow:
#### What happened:
- Manual create a disk on ASH by azure command line.
- Use the manualy create disk create pvc and pv comsume by pod.
```console
$ oc apply -f - <<EOF
apiVersion: v1
kind: PersistentVolume
metadata:
  annotations:
    pv.kubernetes.io/provisioned-by: disk.csi.azure.com
  name: pv-azuredisk-static
spec:
  capacity:
    storage: 10Gi
  accessModes:
    - ReadWriteOnce
  persistentVolumeReclaimPolicy: Retain
  storageClassName: ""  # Important: empty string for static provisioning
  csi:
    driver: disk.csi.azure.com
    # Replace with your actual disk resource ID
    volumeHandle: /subscriptions/XXXXX/resourceGroups/XXXXXX/providers/Microsoft.Compute/disks/pvc-fc003bdf-cea2-49ea-8df7-c3941dfc6a6a
    volumeAttributes:
      fsType: ext4
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc-azuredisk-static
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi
  volumeName: pv-azuredisk-static  # Binds to the specific PV
  storageClassName: "" 
--- 
apiVersion: v1
kind: Pod
metadata:
  name: my-app
spec:
  containers:
  - name: my-app
    image: quay.io/openshifttest/hello-openshift@sha256:56c354e7885051b6bb4263f9faa58b2c292d44790599b7dde0e49e7c466cf339
    volumeMounts:
    - name: azuredisk
      mountPath: /mnt/adata
  volumes:
  - name: azuredisk
    persistentVolumeClaim:
      claimName: pvc-azuredisk-static
EOF  
```
- Check the static provisioned volume could be read and write by workload, while the pod stuck at ContainerCreating, ControllerPublishVolume failed of "panic: runtime error: invalid memory address or nil pointer dereference".   
```console
$ oc describe po/my-app
...
Events:
  Type     Reason              Age                 From                     Message
  ----     ------              ----                ----                     -------
  Normal   Scheduled           11m                 default-scheduler        Successfully assigned default/my-app to pewang-1216ash-cb9jm-worker-mtcazs-5b5n8
[...]

PR #3449, opened 2025-12-16
fix: ASH static provision segmentation fault

Phaow:
**What type of PR is this?**
/kind bug

- fix: ASH static provision segmentation fault issue, in ASH the `NetworkAccessPolicy` does not support so it will always nil while in static provision we need the `InsertDiskProperties ` which does not take it into consideration.

Fixes https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/3448

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azureutils/azure_disk_utils.go
+++ b/pkg/azureutils/azure_disk_utils.go
@@ -746,7 +746,7 @@ func checkDiskName(diskName string) bool {
 	return true
 }
 
-// InsertProperties: insert disk properties to map
+// InsertDiskProperties inserts disk properties to map
 func InsertDiskProperties(disk *armcompute.Disk, publishConext map[string]string) {
 	if disk == nil || publishConext == nil {
 		return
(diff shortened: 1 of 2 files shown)


Phaow:
Hi @andyzhangx , could you help take a look when you get a chacne? Thank you! ^^

pkg/azureutils/azure_disk_utils.go
```diff
@@ -757,7 +757,9 @@ func InsertDiskProperties(disk *armcompute.Disk, publishConext map[string]string
 	}
 	prop := disk.Properties
 	if prop != nil {
-		publishConext[consts.NetworkAccessPolicyField] = string(*prop.NetworkAccessPolicy)
+		if prop.NetworkAccessPolicy != nil {
```
theobarberbany, line 761: I don't like nil checks here - as it's not immediately obvious as to why we're doing them (new comer reading the code won't have context)

Also I'm not aware of how the rest of the CSI driver works, but could there potentially be a case where we expect nil outside of an azurestack environment, and want defaulting?

theobarberbany: IMHO I'd pass something down that lets this function know we're in azure stack not azure for this case.

theobarberbany: If we don't choose to do this, I'd add a comment explaining the field is not supported on ASH and so we shouldn't default it. (Also we expect it to be nil on ASH clusters)

Phaow: Agree, let me add a comment for this. the `InsertDiskProperties` only calls for no CreateVolume call (RequestedSizeGib is not set) - static provision, so the driver dynamic provision(the main use cases for csi driver) is not impacted.

PR #3650, opened 2026-06-02
fix: create volume leaks disk when throttled and PVC deleted

matthias-horne:
**What type of PR is this?**

/kind bug

When Azure throttles a `CreateVolume` request, the throttle commonly fires on a polling `GET` inside the SDK's async-operation poller — after the underlying `PUT` has already been accepted by ARM and the managed disk has begun provisioning. The driver currently returns `codes.Internal`, which the external-provisioner classifies as `ProvisioningFinished` (no retry, no cleanup). If the PVC is deleted before the next retry would have fired, the disk is leaked in Azure.
 
Fix: At the `CreateVolume` gRPC boundary, detect throttling errors via `azureutils.IsThrottlingError` and return `codes.Aborted` instead of `codes.Internal`. The CSI spec defines `Aborted` for `CreateVolume` as "Operation pending for the specified volume" — exactly this case. The external-provisioner's [`checkError()`](https://github.com/kubernetes-csi/external-provisioner/blob/6d3927887fda9105daa6317b1f64a5f88a3cbe65/pkg/controller/controller.go#L2035-L2070) then maps `Aborted` to `ProvisioningInBackground`, the work stays on the queue, and CSI idempotency converges on retry.

Side fix: `azureutils.IsThrottlingError` is broadened to recognize the cloud-provider-azure error from `vendor/sigs.k8s.io/cloud-provider-azure/pkg/azclient/policy/retryrepectthrottled`.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
fix: create volume leaks disk when throttled and PVC deleted
```

### DIFF
--- a/pkg/azuredisk/controllerserver.go
+++ b/pkg/azuredisk/controllerserver.go
@@ -466,6 +466,20 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		if strings.Contains(err.Error(), consts.NotFound) {
 			return nil, status.Error(codes.NotFound, err.Error())
 		}
+		// On Azure ARM throttling, the underlying PUT may have already been
+		// accepted by ARM (the throttle commonly fires on a polling GET inside
+		// the SDK poller after the disk creation has started). Returning
+		// codes.Internal here causes external-provisioner's checkError() to
+		// classify the operation as ProvisioningFinished — no retry, no
+		// cleanup — which leaks the disk in Azure if the PVC is meanwhile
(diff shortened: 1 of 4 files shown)


andyzhangx (commented):
/ok-to-test

matthias-horne:
Failing tests look unrelated to my change, because the volume has already been created successfully at the time of the error. Previous executions of the tests have shown similar errors, so I assume retesting is the way to go.

/retest

andyzhangx (approved):
/lgtm

PR #3668, opened 2026-06-12
fix: use node informer to get node labels

nearora-msft:
**What type of PR is this?**
/kind bug

GetNodeInfoFromLabels makes a direct API server call (Nodes().Get()) every time it needs to read zone and instance type labels from a node. This adds unnecessary load on the API server, especially in the controller path (CheckDiskCountForBatching).

Solution
Introduce a metadata-only node informer that caches node objects locally, stripping Spec, Status, and ManagedFields to minimize memory overhead. The cached NodeLister is passed to GetNodeInfoFromNodeLister and used as the primary lookup path. The nodeLister is only initialized for csi-controller though as there's no need to cache the labels for all the nodes at the csi-node level. The csi-node level call to GetNodeInfoFromLabels would still go to the API server.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azure_controller_common.go
+++ b/pkg/azuredisk/azure_controller_common.go
@@ -31,6 +31,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/types"
 	kwait "k8s.io/apimachinery/pkg/util/wait"
+	corelisters "k8s.io/client-go/listers/core/v1"
 	cloudprovider "k8s.io/cloud-provider"
 	volerr "k8s.io/cloud-provider/volume/errors"
 	"k8s.io/klog/v2"
@@ -92,6 +93,7 @@ type controllerCommon struct {
 	lockMap       *lockMap
 	cloud         *provider.Cloud
 	clientFactory azclient.ClientFactory
+	nodeLister    corelisters.NodeLister
 	// disk queue that is waiting for attach or detach on specific node
 	// <nodeName, map<diskURI, *provider.AttachDiskOptions/DetachDiskOptions>>
 	attachDiskMap sync.Map
@@ -234,7 +236,7 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 
 	numDisksAllowed := math.MaxInt
 	if c.CheckDiskCountForBatching {
-		_, instanceType, err := GetNodeInfoFromLabels(ctx, string(nodeName), c.cloud.KubeClient)
+		_, instanceType, err := GetNodeInfoFromLabels(ctx, string(nodeName), c.cloud.KubeClient, c.nodeLister)
 		if err != nil {
 			klog.Errorf("failed to get node info from labels: %v", err)
 		} else if instanceType != "" {
--- a/pkg/azuredisk/azure_controller_common.go
+++ b/pkg/azuredisk/azure_controller_common.go
@@ -31,7 +31,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/types"
 	kwait "k8s.io/apimachinery/pkg/util/wait"
-	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
 	cloudprovider "k8s.io/cloud-provider"
 	volerr "k8s.io/cloud-provider/volume/errors"
 	"k8s.io/klog/v2"
@@ -93,7 +93,7 @@ type controllerCommon struct {
 	lockMap       *lockMap
 	cloud         *provider.Cloud
 	clientFactory azclient.ClientFactory
-	nodeLister    corelisters.NodeLister
+	nodeLister    cache.GenericLister
 	// disk queue that is waiting for attach or detach on specific node
 	// <nodeName, map<diskURI, *provider.AttachDiskOptions/DetachDiskOptions>>
 	attachDiskMap sync.Map
--- a/pkg/azuredisk/azure_controller_common.go
+++ b/pkg/azuredisk/azure_controller_common.go
@@ -236,7 +236,13 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 
 	numDisksAllowed := math.MaxInt
 	if c.CheckDiskCountForBatching {
-		_, instanceType, err := GetNodeInfoFromLabels(ctx, string(nodeName), c.cloud.KubeClient, c.nodeLister)
+		var instanceType string
+		var err error
+		if c.nodeLister != nil {
+			_, instanceType, err = GetNodeInfoFromNodeLister(string(nodeName), c.nodeLister)
+		} else {
+			_, instanceType, err = GetNodeInfoFromLabels(ctx, string(nodeName), c.cloud.KubeClient)
+		}
 		if err != nil {
 			klog.Errorf("failed to get node info from labels: %v", err)
 		} else if instanceType != "" {
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -48,6 +48,10 @@ import (
 	"k8s.io/mount-utils"
 	"k8s.io/utils/ptr"
 
+	"k8s.io/client-go/informers"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+
 	consts "sigs.k8s.io/azuredisk-csi-driver/pkg/azureconstants"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
 	csicommon "sigs.k8s.io/azuredisk-csi-driver/pkg/csi-common"
@@ -156,6 +160,8 @@ type Driver struct {
 	enableMigrationMonitor      bool
 	// whether to convert ReadWrite cachingMode to ReadOnly for intree PVs to avoid issues
 	convertRWCachingModeForIntreePV bool
+	nodeLister                      corelisters.NodeLister
+	nodeInformerSynced              cache.InformerSynced
 }
 
 // NewDriver Creates a NewCSIDriver object. Assumes vendor version is equal to driver version &
@@ -332,6 +338,31 @@ func NewDriver(options *DriverOptions) *Driver {
 					time.Sleep(10 * time.Minute)
 				}
 			}()
+		}
+	}
+
+	if kubeClient != nil && driver.getNodeInfoFromLabels {
+		// Create a metadata-only node informer to cache node labels locally
+		nodeInformerFactory := informers.NewSharedInformerFactoryWithOptions(kubeClient, 0,
+			informers.WithTransform(func(obj interface{}) (interface{}, error) {
+				if node, ok := obj.(*corev1.Node); ok {
+					// Strip everything except metadata to reduce memory usage
+					node.Spec = corev1.NodeSpec{}
+					node.Status = corev1.NodeStatus{}
+					node.ManagedFields = nil
+					klog.V(4).Infof("node informer transform: node(%s) cached with metadata only - spec empty: %v, status empty: %v, managedFields nil: %v, labels: %v",
+						node.Name, reflect.DeepEqual(node.Spec, corev1.NodeSpec{}), reflect.DeepEqual(node.Status, corev1.NodeStatus{}), node.ManagedFields == nil, node.Labels)
+				}
+				return obj, nil
+			}),
+		)
+		nodeInformer := nodeInformerFactory.Core().V1().Nodes()
+		driver.nodeLister = nodeInformer.Lister()
+		driver.nodeInformerSynced = nodeInformer.Informer().HasSynced
+		nodeInformerFactory.Start(context.Background().Done())
+		klog.V(2).Infof("started node informer for GetNodeInfoFromLabels caching")
+		if driver.diskController != nil {
+			driver.diskController.nodeLister = driver.nodeLister
 		}
 	}
 
@@ -675,8 +706,25 @@ func (d *Driver) getUsedLunsFromNode(ctx context.Context, nodeName k8stypes.Node
 	return usedLuns, nil
 }
 
-// getNodeInfoFromLabels get zone, instanceType from node labels
-func GetNodeInfoFromLabels(ctx context.Context, nodeName string, kubeClient clientset.Interface) (string, string, error) {
+// GetNodeInfoFromLabels get zone, instanceType from node labels.
+// If nodeLister is non-nil, it uses the cached lister; otherwise falls back to an API call.
+// If the lister returns an error or a node with empty labels (e.g. cache not yet synced), it falls back to the API server.
+func GetNodeInfoFromLabels(ctx context.Context, nodeName string, kubeClient clientset.Interface, nodeLister corelisters.NodeLister) (string, string, error) {
+	if nodeLister != nil {
+		node, err := nodeLister.Get(nodeName)
+		if err == nil && len(node.Labels) > 0 {
+			zone := node.Labels[consts.WellKnownTopologyKey]
+			instanceType := node.Labels[consts.InstanceTypeKey]
+			klog.V(2).Infof("GetNodeInfoFromLabels: Node informer details for node(%s): zone=%s, instanceType=%s, labels=%v", nodeName, zone, instanceType, node.Labels)
+			return zone, instanceType, nil
+		}
+		if err != nil {
+			klog.V(4).Infof("get node(%s) from lister failed: %v, falling back to API server", nodeName, err)
+		} else {
+			klog.V(4).Infof("node(%s) labels from lister are empty, falling back to API server", nodeName)
+		}
+	}
+
 	if kubeClient == nil || kubeClient.CoreV1() == nil {
 		return "", "", fmt.Errorf("kubeClient is nil")
 	}
@@ -689,7 +737,10 @@ func GetNodeInfoFromLabels(ctx context.Context, nodeName string, kubeClient clie
 	if len(node.Labels) == 0 {
 		return "", "", fmt.Errorf("node(%s) label is empty", nodeName)
 	}
-	return node.Labels[consts.WellKnownTopologyKey], node.Labels[consts.InstanceTypeKey], nil
+	zone := node.Labels[consts.WellKnownTopologyKey]
+	instanceType := node.Labels[consts.InstanceTypeKey]
+	klog.V(2).Infof("GetNodeInfoFromLabels: API server node details for node(%s): zone=%s, instanceType=%s, labels=%v", nodeName, zone, instanceType, node.Labels)
+	return zone, instanceType, nil
 }
 
 // getDefaultDiskIOPSReadWrite according to requestGiB
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -249,9 +249,12 @@ func NewDriver(options *DriverOptions) *Driver {
 	if err != nil {
 		klog.Warningf("get kubeconfig(%s) failed with error: %v", options.Kubeconfig, err)
 	}
-	kubeClient, err := azureutils.GetKubeClient(options.Kubeconfig, options.KubeAPIQPS, options.KubeAPIBurst)
-	if err != nil {
-		klog.Warningf("get kubeclient failed with error: %v", err)
+	var kubeClient clientset.Interface
+	if kubeConfig != nil {
+		kubeClient, err = clientset.NewForConfig(kubeConfig)
+		if err != nil {
+			klog.Warningf("get kubeclient failed with error: %v", err)
+		}
 	}
 	driver.kubeClient = kubeClient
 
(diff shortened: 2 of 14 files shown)


andyzhangx (commented):
`getNodeInfoFromLabels` is false by default in AKS or upstream install, how did you find the problem?

https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/2e9b957a8c8d6845b6cd2633456c4529644a5913/charts/latest/azuredisk-csi-driver/values.yaml#L229

andyzhangx (commented):
Thanks for the fix! The goal of caching node labels to avoid repeated API calls makes sense.

However, I think a simpler approach would be to add a TTL cache directly inside `GetNodeInfoFromLabels` rather than introducing a full node informer + lister + factory. Node labels like zone and instance type rarely change, so a simple in-memory cache with expiration would suffice:

```go
var nodeInfoCache sync.Map // nodeName -> *cachedNodeInfo

type cachedNodeInfo struct {
    zone         string
    instanceType string
    expireAt     time.Time
}

func GetNodeInfoFromLabels(ctx context.Context, nodeName string, kubeClient clientset.Interface) (string, string, error) {
    // Check cache first
    if cached, ok := nodeInfoCache.Load(nodeName); ok {
        info := cached.(*cachedNodeInfo)
        if time.Now().Before(info.expireAt) {
            return info.zone, info.instanceType, nil
        }
    }

    if kubeClient == nil || kubeClient.CoreV1() == nil {
        return "", "", fmt.Errorf("kubeClient is nil")
    }

    node, err := kubeClient.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
    if err != nil {
        return "", "", fmt.Errorf("get node(%s) failed with %v", nodeName, err)
    }
    if len(node.Labels) == 0 {
        return "", "", fmt.Errorf("node(%s) label is empty", nodeName)
    }

    zone := node.Labels[consts.WellKnownTopologyKey]
    instanceType := node.Labels[consts.InstanceTypeKey]

    // Cache for 10 minutes - zone/instanceType labels never change
    nodeInfoCache.Store(nodeName, &cachedNodeInfo{zone, instanceType, time.Now().Add(10 * time.Minute)})
    return zone, instanceType, nil
}
```

Benefits over the informer approach:
1. **No function signature change** - doesn't break callers or require passing `nodeLister` everywhere
2. **Much less code** - ~15 lines vs 50+ lines of informer setup, transform, sync waiting, fallback logic
3. **No informer lifecycle management** - no factory, no `HasSynced`, no `Start(ctx.Done())`
4. **Lower memory** - only caches the 2 strings we actually need per node, not the full Node metadata object
5. **Simpler fallback** - cache miss just does the API call, no dual-path logic

The informer approach would make sense if we needed to watch for label changes reactively or needed many different label fields, but for just zone + instanceType (which essentially never change for a node), a simple TTL cache is sufficient and much easier to maintain.

nearora-msft:
> `getNodeInfoFromLabels` is false by default in AKS or upstream install, how did you find the problem?
> 
> https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/2e9b957a8c8d6845b6cd2633456c4529644a5913/charts/latest/azuredisk-csi-driver/values.yaml#L229

@landreasyan Do you know if the affected cluster had this set to true?

pkg/azuredisk/azuredisk.go
```diff
@@ -335,6 +341,31 @@ func NewDriver(options *DriverOptions) *Driver {
 		}
 	}
 
+	if kubeClient != nil && driver.getNodeInfoFromLabels {
+		// Create a metadata-only node informer to cache node labels locally
+		nodeInformerFactory := informers.NewSharedInformerFactoryWithOptions(kubeClient, 0,
+			informers.WithTransform(func(obj interface{}) (interface{}, error) {
+				if node, ok := obj.(*corev1.Node); ok {
```
landreasyan: I would recommend not caching the full node resource and creating informers for metadata only: https://pkg.go.dev/k8s.io/client-go/metadata/metadatainformer

This way, you won't have to zero any values.

pkg/azuredisk/azuredisk.go
```diff
@@ -675,8 +706,25 @@ func (d *Driver) getUsedLunsFromNode(ctx context.Context, nodeName k8stypes.Node
...
+		if err == nil && len(node.Labels) > 0 {
+			zone := node.Labels[consts.WellKnownTopologyKey]
+			instanceType := node.Labels[consts.InstanceTypeKey]
+			klog.V(2).Infof("GetNodeInfoFromLabels: Node informer details for node(%s): zone=%s, instanceType=%s, labels=%v", nodeName, zone, instanceType, node.Labels)
+			return zone, instanceType, nil
+		}
+		if err != nil {
+			klog.V(4).Infof("get node(%s) from lister failed: %v, falling back to API server", nodeName, err)
+		} else {
+			klog.V(4).Infof("node(%s) labels from lister are empty, falling back to API server", nodeName)
```
landreasyan: Why are we falling back to none cached client when the labels are empty? I actually don't understand why we need to fallback to none cached client at all.

pkg/azuredisk/azuredisk.go
```diff
@@ -335,6 +341,31 @@ func NewDriver(options *DriverOptions) *Driver {
 		}
 	}
 
+	if kubeClient != nil && driver.getNodeInfoFromLabels {
```
landreasyan: Isn't `getNodeInfoFromLabels` set to false by default? Unless we are changing the deployment, this doesn't seem like the right gate to address the client side throttling issue.

landreasyan: This needs to be or-ed with `CheckDiskCountForBatching`.

landreasyan: I don't think we need to setup the informers within it's consumer, i.e. the Driver. I would recommend moving the informers to main, waiting for them to sync and then passing them after `WaitForCacheSync` (currently missing in the PR) when we know the cache has been populated and can be used by the driver.

pkg/azuredisk/azuredisk.go
```diff
@@ -335,6 +341,31 @@ func NewDriver(options *DriverOptions) *Driver {
...
+		// Create a metadata-only node informer to cache node labels locally
+		nodeInformerFactory := informers.NewSharedInformerFactoryWithOptions(kubeClient, 0,
+			informers.WithTransform(func(obj interface{}) (interface{}, error) {
+				if node, ok := obj.(*corev1.Node); ok {
+					// Strip everything except metadata to reduce memory usage
+					node.Spec = corev1.NodeSpec{}
+					node.Status = corev1.NodeStatus{}
+					node.ManagedFields = nil
+					klog.V(4).Infof("node informer transform: node(%s) cached with metadata only - spec empty: %v, status empty: %v, managedFields nil: %v, labels: %v",
+						node.Name, reflect.DeepEqual(node.Spec, corev1.NodeSpec{}), reflect.DeepEqual(node.Status, corev1.NodeStatus{}), node.ManagedFields == nil, node.Labels)
```
landreasyan: I don't think it is worth doing the `reflect.DeepEqual` calls here. It is an expensive operation that would run regardless of the log level.

landreasyan:
Thanks for the PR, Neha! I left review comments but the tldr these are the blockers I see for this PR: 

1. Wrong gate. The informer is created under if `kubeClient` != nil && driver.getNodeInfoFromLabels`, but --get-node-info-from-labels is a node-side flag (used by `NodeGetInfo` for zone discovery, default false). The actual consumer of the lister is `controllerCommon.AttachDisk` on the controller, gated by `--check-disk-count-for-batching (default true)`

2. Use a metadata-only informer, not a full Node informer. 
3. Construct informers in main, WaitForCacheSync before passing to the driver and serving. 
4. Remove the apiserver fallback

In the future we should move all kube api calls to cache, so the updates we make here should align with that goal.

pkg/azuredisk/azuredisk.go
nearora-msft: Yes, I had added these for debugging purposes. Will remove.

pkg/azuredisk/azuredisk.go
nearora-msft: For scale out cases when the node hasn't synced and isn't available in the cache yet, it'd be good to use the API server fallback.

nearora-msft:
> Thanks for the fix! The goal of caching node labels to avoid repeated API calls makes sense.
> 
> However, I think a simpler approach would be to add a TTL cache directly inside `GetNodeInfoFromLabels` rather than introducing a full node informer + lister + factory. Node labels like zone and instance type rarely change, so a simple in-memory cache with expiration would suffice:
> 
> ```go
> var nodeInfoCache sync.Map // nodeName -> *cachedNodeInfo
> 
> type cachedNodeInfo struct {
>     zone         string
>     instanceType string
>     expireAt     time.Time
> }
> 
> func GetNodeInfoFromLabels(ctx context.Context, nodeName string, kubeClient clientset.Interface) (string, string, error) {
>     // Check cache first
>     if cached, ok := nodeInfoCache.Load(nodeName); ok {
>         info := cached.(*cachedNodeInfo)
>         if time.Now().Before(info.expireAt) {
>             return info.zone, info.instanceType, nil
>         }
>     }
> 
>     if kubeClient == nil || kubeClient.CoreV1() == nil {
>         return "", "", fmt.Errorf("kubeClient is nil")
>     }
> 
>     node, err := kubeClient.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
>     if err != nil {
>         return "", "", fmt.Errorf("get node(%s) failed with %v", nodeName, err)
>     }
>     if len(node.Labels) == 0 {
>         return "", "", fmt.Errorf("node(%s) label is empty", nodeName)
>     }
> 
>     zone := node.Labels[consts.WellKnownTopologyKey]
>     instanceType := node.Labels[consts.InstanceTypeKey]
> 
>     // Cache for 10 minutes - zone/instanceType labels never change
>     nodeInfoCache.Store(nodeName, &cachedNodeInfo{zone, instanceType, time.Now().Add(10 * time.Minute)})
>     return zone, instanceType, nil
> }
> ```
> 
> Benefits over the informer approach:
> 
> 1. **No function signature change** - doesn't break callers or require passing `nodeLister` everywhere
> 2. **Much less code** - ~15 lines vs 50+ lines of informer setup, transform, sync waiting, fallback logic
> 3. **No informer lifecycle management** - no factory, no `HasSynced`, no `Start(ctx.Done())`
> 4. **Lower memory** - only caches the 2 strings we actually need per node, not the full Node metadata object
> 5. **Simpler fallback** - cache miss just does the API call, no dual-path logic
> 
> The informer approach would make sense if we needed to watch for label changes reactively or needed many different label fields, but for just zone + instanceType (which essentially never change for a node), a simple TTL cache is sufficient and much easier to maintain.

@andyzhangx What are the downsides of using inbuilt caching instead of maintaining our own cache? If we are just storing the labels, there shouldn't be any memory overhead. The code still falls back to the API server logic anyway.
Do you see any other overhead?

pkg/azuredisk/azuredisk.go
landreasyan: Informers would update the cache within milliseconds once the node is added, so this shouldn't be an issue during scaling.

pkg/azuredisk/azuredisk.go
nearora-msft: Done.

pkg/azuredisk/azuredisk.go
nearora-msft: Ok, but what issue do you see in falling back to API server?

pkg/azuredisk/azuredisk.go
nearora-msft: This is just the initialization code, we are actually not starting the grpc servers until we call run. So, this should be fine?
I have also added WaitForCacheSync  in the new iteration.

andyzhangx:
@nearora-msft Good question. The informer approach works, but for this specific use case I think a simple TTL cache is a better fit:

1. **Overhead beyond memory** — even a metadata-only informer watches all node change events and maintains a long-lived connection. `GetNodeInfoFromLabels` is called once at `NodeGetInfo` startup and occasionally during `AttachDisk` — a watch-all-nodes informer is overkill for that frequency.

2. **Lifecycle complexity** — informer requires factory setup, `Start(ctx.Done())`, `HasSynced` sync-waiting, and fallback logic when the cache is not yet populated. A TTL cache is just a `sync.Map` load/store — no lifecycle management needed.

3. **No signature changes** — the TTL cache lives inside `GetNodeInfoFromLabels` itself, so no callers need to be updated (no `nodeLister` parameter threading). This is important since the function is called from both node-side and controller-side paths.

4. **These labels never change** — `topology.kubernetes.io/zone` and `node.kubernetes.io/instance-type` are immutable for a node's lifetime. A 10-minute TTL is just a safety net; the cached values will never go stale in practice.

The informer pattern makes sense when you need to react to label changes or access many different fields across many nodes. Here we just need 2 immutable strings, queried infrequently — a simple cache is sufficient and much easier to maintain.

nearora-msft:
> @nearora-msft Good question. The informer approach works, but for this specific use case I think a simple TTL cache is a better fit:
> 
> 1. **Overhead beyond memory** — even a metadata-only informer watches all node change events and maintains a long-lived connection. `GetNodeInfoFromLabels` is called once at `NodeGetInfo` startup and occasionally during `AttachDisk` — a watch-all-nodes informer is overkill for that frequency.
> 2. **Lifecycle complexity** — informer requires factory setup, `Start(ctx.Done())`, `HasSynced` sync-waiting, and fallback logic when the cache is not yet populated. A TTL cache is just a `sync.Map` load/store — no lifecycle management needed.
> 3. **No signature changes** — the TTL cache lives inside `GetNodeInfoFromLabels` itself, so no callers need to be updated (no `nodeLister` parameter threading). This is important since the function is called from both node-side and controller-side paths.
> 4. **These labels never change** — `topology.kubernetes.io/zone` and `node.kubernetes.io/instance-type` are immutable for a node's lifetime. A 10-minute TTL is just a safety net; the cached values will never go stale in practice.
> 
> The informer pattern makes sense when you need to react to label changes or access many different fields across many nodes. Here we just need 2 immutable strings, queried infrequently — a simple cache is sufficient and much easier to maintain.

@andyzhangx If we go with this approach, there's a large scale out event, we will end up making several GetNode calls because the nodes aren't in the local cache, which would again lead to api server throttling. And consider the case where the cluster consists of spot VMs which get created and deleted frequently, we may again face api server throttling. What do you think about those cases?

pkg/azuredisk/azuredisk.go
landreasyan: I really think it is unnecessary, complicates the code and opens up to the same risk of hitting client side throttling.

pkg/azuredisk/azuredisk.go
```diff
@@ -238,10 +245,14 @@ func NewDriver(options *DriverOptions) *Driver {
 	userAgent := GetUserAgent(driver.Name, driver.customUserAgent, driver.userAgentSuffix)
 	klog.V(2).Infof("driver userAgent: %s", userAgent)
 
-	kubeClient, err := azureutils.GetKubeClient(options.Kubeconfig, options.KubeAPIQPS, options.KubeAPIBurst)
+	kubeConfig, err := azureutils.GetKubeConfig(options.Kubeconfig, options.KubeAPIQPS, options.KubeAPIBurst)
 	if err != nil {
 		klog.Warningf("get kubeconfig(%s) failed with error: %v", options.Kubeconfig, err)
 	}
+	kubeClient, err := azureutils.GetKubeClient(options.Kubeconfig, options.KubeAPIQPS, options.KubeAPIBurst)
```
landreasyan: This method is redundant now as it calls the `GetKubeConfig` again. We should just call `NewForConfig(config)` similar to the metadata client.

pkg/azuredisk/azuredisk.go
```diff
@@ -675,8 +712,30 @@ func (d *Driver) getUsedLunsFromNode(ctx context.Context, nodeName k8stypes.Node
...
+			if ok && len(pom.Labels) > 0 {
+				zone := pom.Labels[consts.WellKnownTopologyKey]
+				instanceType := pom.Labels[consts.InstanceTypeKey]
+				klog.V(2).Infof("GetNodeInfoFromLabels: Node informer details for node(%s): zone=%s, instanceType=%s", nodeName, zone, instanceType)
+				return zone, instanceType, nil
+			}
+			if !ok {
+				klog.V(4).Infof("node(%s) from lister is not *metav1.PartialObjectMetadata, falling back to API server", nodeName)
+			} else {
+				klog.V(4).Infof("node(%s) labels from lister are empty, falling back to API server", nodeName)
```
landreasyan: The informer's cache is populated by a list/watch against the same API server that the fallback Nodes().Get() would hit. The lister and the direct call therefore read from the same source of truth. The only difference is potential cache staleness, which is irrelevant for `topology.kubernetes.io/zone` and `node.kubernetes.io/instance-type` since both are set at node registration and are immutable for the node's lifetime.

If the informer has the node cached with empty labels, that is the current state in the API server. A direct get will return the same object with the same empty labels the fallback to API to retrieve identical data. The only case where they'd differ is a brief race where the node was just created and labels haven't been applied yet but the watch will deliver that update within milliseconds, so the next reconcile/retry picks it up from cache. The direct read is subject to the same timing.

There are also no transient or network issues when retrieving data from cache. If you hit a cache miss, the node is likely not found or a brief moment where it was just added. There are no correctness concerns here. We risk hitting the same issue that the PR aims to fix if we keep the fallback. 

When a lister is configured, we need to trust it and if there is a cache miss or empty labels, return an empty result and let the caller proceed.

landreasyan:
> > @nearora-msft Good question. The informer approach works, but for this specific use case I think a simple TTL cache is a better fit:
> > 
> > 1. **Overhead beyond memory** — even a metadata-only informer watches all node change events and maintains a long-lived connection. `GetNodeInfoFromLabels` is called once at `NodeGetInfo` startup and occasionally during `AttachDisk` — a watch-all-nodes informer is overkill for that frequency.
> > 2. **Lifecycle complexity** — informer requires factory setup, `Start(ctx.Done())`, `HasSynced` sync-waiting, and fallback logic when the cache is not yet populated. A TTL cache is just a `sync.Map` load/store — no lifecycle management needed.
> > 3. **No signature changes** — the TTL cache lives inside `GetNodeInfoFromLabels` itself, so no callers need to be updated (no `nodeLister` parameter threading). This is important since the function is called from both node-side and controller-side paths.
> > 4. **These labels never change** — `topology.kubernetes.io/zone` and `node.kubernetes.io/instance-type` are immutable for a node's lifetime. A 10-minute TTL is just a safety net; the cached values will never go stale in practice.
> > 
> > The informer pattern makes sense when you need to react to label changes or access many different fields across many nodes. Here we just need 2 immutable strings, queried infrequently — a simple cache is sufficient and much easier to maintain.
> 
> @andyzhangx If we go with this approach, there's a large scale out event, we will end up making several GetNode calls because the nodes aren't in the local cache, which would again lead to api server throttling. And consider the case where the cluster consists of spot VMs which get created and deleted frequently, we may again face api server throttling. What do you think about those cases?

We have considered the sync.Map vs informer/lister before creating the PR and landed on keeping the lister. 

The motivating concern is API-server throttling during large scale-outs and on spot-heavy clusters with constant node churn. With a sync.Map, every cold miss is a direct Nodes().Get() and cold misses are concentrated exactly during scale-out/scale-in events, when API pressure is already highest. The informer replaces N unbounded point GETs with one LIST plus a single streamed watch, which is strictly cheaper and is the recommended, more api friendly pattern.

The cached objects are PartialObjectMetadata (small), and we can strip managedFields via a transform if we want them smaller. This is very close to the sync.Map without the cost of maintenance. 

Most importantly, I strongly feel that the driver should eventually serve all kube reads from informer-backed caches. A lister is a reusable, extensible primitive the same shared factory can back for future lookups. A one-off sync.Map is a single-purpose pattern we'd have to rip out the moment we cache a second resource.

I think we need to keep the lister, drop the API-server fallback paths, and (potentially) add a managedFields transform.

pkg/azuredisk/azuredisk.go
nearora-msft: Done, only falling back to api server if lister isn't available. This is to handle the csi-node driver, where I don't see a point of initializing the lister. getNodeInfoFromlabels is false by default so that path shouldn't generally be exercised. When it is set to true, it will fall back to API server. The idea soln would be to cache the per node labels for that node on the node, but that change could be made in another PR.

pkg/azuredisk/azure_controller_common.go
```diff
@@ -234,7 +236,7 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 
 	numDisksAllowed := math.MaxInt
 	if c.CheckDiskCountForBatching {
-		_, instanceType, err := GetNodeInfoFromLabels(ctx, string(nodeName), c.cloud.KubeClient)
+		_, instanceType, err := GetNodeInfoFromLabels(ctx, string(nodeName), c.cloud.KubeClient, c.nodeLister)
```
hccheng72:
```suggestion
		if c.nodeLister != nil {
          // use lister
        } else {
           _, instanceType, err := GetNodeInfoFromLabels(ctx, string(nodeName), c.cloud.KubeClient)
        }
```
can we keep the lister lookup inside `AttachDisk()` so `GetNodeInfoFromLabels()` signature stays unchanged and nodeserver doesn't need updates?

pkg/azureutils/azure_disk_utils.go
```diff
@@ -261,6 +262,15 @@ func GetKubeClient(kubeconfig string, qps float64, burst int) (clientset.Interfa
 		config.Burst = burst
 	}
 
+	return config, nil
+}
+
+func GetKubeClient(kubeconfig string, qps float64, burst int) (clientset.Interface, error) {
```
landreasyan, line 268: I don't think this method is used anymore?

pkg/azuredisk/azure_controller_common.go
nearora-msft: Then we would need 2 functions, one would be lister based and the other would be kubeconfig based. That makes sense, though.

pkg/azuredisk/azuredisk.go
```diff
@@ -675,8 +715,30 @@ func (d *Driver) getUsedLunsFromNode(ctx context.Context, nodeName k8stypes.Node
...
+		if len(pom.Labels) == 0 {
+			return "", "", fmt.Errorf("node(%s) labels are empty", nodeName)
+		}
+
+		zone := pom.Labels[consts.WellKnownTopologyKey]
+		instanceType := pom.Labels[consts.InstanceTypeKey]
+		klog.V(2).Infof("GetNodeInfoFromLabels: node(%s): zone=%s, instanceType=%s", nodeName, zone, instanceType)
+		return zone, instanceType, nil
+	}
+
```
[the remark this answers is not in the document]
nearora-msft: Not falling back to the API server is intentional. See this comment: https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3668#discussion_r3443991131

Also this change ensures that we wait for the cache to be synced in the driver initialization code.

pkg/azuredisk/azuredisk.go
```diff
@@ -335,6 +349,32 @@ func NewDriver(options *DriverOptions) *Driver {
 		}
 	}
 
+	if kubeConfig != nil && driver.checkDiskCountForBatching && driver.NodeID == "" {
+		// Create a metadata-only node informer to cache node labels locally (controller only)
+		metadataClient, err := metadata.NewForConfig(kubeConfig)
```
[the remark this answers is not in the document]
nearora-msft: Updated the PR description.

pkg/azureutils/azure_disk_utils.go
nearora-msft: It is getting used here : https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/3d4d99fd86f21e5d8e73f072c02e7a95446b10f0/pkg/azurediskplugin/main.go#L79

pkg/azuredisk/azuredisk.go
```diff
@@ -335,6 +349,32 @@ func NewDriver(options *DriverOptions) *Driver {
...
+			nodeInformerFactory := metadatainformer.NewSharedInformerFactory(metadataClient, 0)
+			nodeGVR := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "nodes"}
+			nodeInformer := nodeInformerFactory.ForResource(nodeGVR)
+			driver.nodeLister = nodeInformer.Lister()
+			driver.nodeInformerSynced = nodeInformer.Informer().HasSynced
+			nodeInformerFactory.Start(context.Background().Done())
+			syncCtx, syncCancel := context.WithTimeout(context.Background(), 30*time.Second)
+			defer syncCancel()
+			if !cache.WaitForCacheSync(syncCtx.Done(), driver.nodeInformerSynced) {
+				klog.Warningf("metadata node informer cache has not synced yet, will continue to sync in background")
```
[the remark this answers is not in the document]
nearora-msft: Moved the informer startuo code to Run() to ensure graceful shutdown.

pkg/azuredisk/azure_controller_common.go
nearora-msft: Done.

nearora-msft:
Merging the pull reuqest as the capz tests are failing for an unrelated reason. They were also failing for the PRs that have already merged.

PR #3672, opened 2026-06-19
feat(azuredisk): skip detach for already-unattached disk on ControllerUnpublishVolume

jmclong:
**What type of PR is this?**
/kind bug

During mass detach (e.g. when many spot/preemptible nodes are deleted at once), ControllerUnpublishVolume calls DetachDisk per volume, which serializes on the global VMManagementType mutex and can stall the controller.

Add a fast-path that does a cheap disk read first: if the disk no longer exists (HTTP 404) or exists but has no owning VM (ManagedBy and ManagedByExtended both empty), the volume is already detached from every node, so we return success without taking the node-centric DetachDisk path. checkDiskExists honors GetDisk throttling and a timeout, so any read error simply falls through to the normal detach.

The behavior is gated by a new CLI flag --skip-detach-for-unattached-disk (default true). Adds unit tests for diskHasNoOwner and isDiskNotFoundError.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azuredisk.go
+++ b/pkg/azuredisk/azuredisk.go
@@ -21,11 +21,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/http"
 	"reflect"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v7"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
@@ -138,6 +140,7 @@ type Driver struct {
 	forceDetachBackoff                 bool
 	waitForDetach                      bool
 	waitForDetachDiskComplete          bool
+	skipDetachForUnattachedDisk        bool
 	endpoint                           string
 	disableAVSetNodes                  bool
 	removeNotReadyTaint                bool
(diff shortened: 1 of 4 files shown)


NedAnd1:
In `DetachDisk`, we check if the instance is not found, and if the instance is not found we wait for the disk managed by property to go away using `c.waitForDiskManagedByTobeRemoved(...)` which says "it actually takes a while for disk property to be updated", so why are we expecting that a disk would no longer exist or no longer have an owning VM by the time ControllerUnpublishVolume is called?

jmclong:
> In `DetachDisk`, we check if the instance is not found, and if the instance is not found we wait for the disk managed by property to go away using `c.waitForDiskManagedByTobeRemoved(...)` which says "it actually takes a while for disk property to be updated", so why are we expecting that a disk would no longer exist or no longer have an owning VM by the time ControllerUnpublishVolume is called?

Thanks, that's fair. This scenario was to avoid going down the detach path when the node was long gone as a fast path to avoid going down some problematic cloud-provider-azure code. 

The spec says to return 0 OK when the volume or node no longer exists 

_This operation MUST be idempotent. If the volume corresponding to the volume_id is not attached to the node corresponding to the node_id, the Plugin MUST reply 0 OK. If the volume corresponding to the volume_id or the node corresponding to node_id cannot be found by the Plugin and the volume can be safely regarded as ControllerUnpublished from the node, the plugin SHOULD return 0 OK. If this operation failed, or the CO does not know if the operation failed or not, it can choose to call ControllerUnpublishVolume again._

But, as you pointed out there is some lag on ManagedBY being populated. I think it's find for this case but I don't think we have correctness after a quick attach/detach. For that reason, I don't think this code is correct.

hccheng72:
not related to your PR, but can we update the log to "azure instance id not found" to differentiate the real failure to get instance id?

https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/a021f4a3e74d5b08a38a7fd478bb432f058e15d3/pkg/azuredisk/azure_controller_common.go#L378

PR #3680, opened 2026-06-24
chore(lockMap): for cancelling wait on ctx cancellation

jmclong:
**What type of PR is this?**
/kind feature

This PR swaps to using `sync.Semaphore` instead of `sync.Mutex` to cancel waiting if context expires. `sync.Semaphore` and `sync.Mutex` have similar fairness mechanisms. This prevents requests from hanging on mutex while context has actually already expired and no progress will be possible when calling any method that requires a non-cancelled mutex.

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azure_controller_common.go
+++ b/pkg/azuredisk/azure_controller_common.go
@@ -200,7 +200,7 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 	}
 	node := strings.ToLower(string(nodeName))
 	diskuri := strings.ToLower(diskURI)
-	requestNum, err := c.batchAttachDiskRequest(diskuri, node, &options)
+	requestNum, err := c.batchAttachDiskRequest(ctx, diskuri, node, &options)
 	if err != nil {
 		return -1, err
 	}
@@ -208,8 +208,8 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 	var waitForDetachHappened bool
 	if c.WaitForDetach && c.isMaxDataDiskCountExceeded(ctx, string(nodeName)) {
 		// wait for disk detach to finish first on the same node
-		if err = kwait.PollUntilContextTimeout(ctx, 2*time.Second, 30*time.Second, true, func(context.Context) (bool, error) {
-			detachDiskReqeustNum, err := c.getDetachDiskRequestNum(node)
+		if err = kwait.PollUntilContextTimeout(ctx, 2*time.Second, 30*time.Second, true, func(ctx context.Context) (bool, error) {
+			detachDiskReqeustNum, err := c.getDetachDiskRequestNum(ctx, node)
 			if err != nil {
 				return false, err
 			}
@@ -224,7 +224,9 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 		}
 	}
 
-	c.lockMap.LockEntry(node)
+	if err := c.lockMap.LockEntry(ctx, node); err != nil {
+		return -1, fmt.Errorf("AttachDisk: failed to acquire node lock for node %s: %w", nodeName, err)
+	}
 	defer c.lockMap.UnlockEntry(node)
 
 	if !waitForDetachHappened && c.AttachDetachInitialDelayInMs > 0 && requestNum == 1 {
@@ -254,7 +256,7 @@ func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI str
 		}
 	}
 
-	diskMap, err := c.retrieveAttachBatchedDiskRequests(node, diskuri)
+	diskMap, err := c.retrieveAttachBatchedDiskRequests(ctx, node, diskuri)
 	if err != nil {
 		return -1, err
 	}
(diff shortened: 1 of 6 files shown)


andyzhangx (approved):
/lgtm

andyzhangx:
## Possible regression surfaced by the release-1.34 cherry-pick e2e (#3705)

Sharing an analysis of a Windows e2e failure that looks strongly correlated with this PR. Not asserting a revert is needed — but worth a second look before this ships more widely.

### Signal

On [pull-azuredisk-csi-driver-e2e-capz-windows-2022-hostprocess build 2076218685676064768](https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_azuredisk-csi-driver/3705/pull-azuredisk-csi-driver-e2e-capz-windows-2022-hostprocess/2076218685676064768/build-log.txt) (cherry-pick of this PR to `release-1.34`), 3 tests failed, all on the same node `capz-2q2t-jlhv5`:

- `should detach disk after pod deleted [Windows]` — pod stuck in `ContainerCreating` for 300s
- `should create a statefulset object, write and read to it, delete the pod and write and read to it again [Windows]` — 601s timeout
- `should use a pre-provisioned volume and mount it as readOnly [windows]` — 2112s timeout

### Why this looks tied to the PR (not a generic flake)

The controller log contains error strings that are **only produced by this PR's new code path**:

```
E 11:00:18  Could not detach volume ... context canceled
E 11:00:18  DetachDisk: failed to acquire node lock for node capz-2q2t-jlhv5:
            failed to acquire lock for entry "capz-2q2t-jlhv5":
            context deadline exceeded
```

The wrapped message `"DetachDisk: failed to acquire node lock for node %s: %w"` is added by this PR (`azure_controller_common.go` L397 in the diff); before the PR, `LockEntry` used `sync.Mutex` and could not return an error. So its appearance means the new context-cancelling code path was hit.

### Failure chain

Four disks tried to detach from the same node in the same window:

```
11:00:18  DetachDisk pre-provisioned-readOnly:          context canceled
11:00:18  DetachDisk pvc-ee97…:                         context canceled
11:00:18  DetachDisk pvc-af27…:                         ctx deadline exceeded (on lock)
11:00:18  DetachDisk pvc-df19…:                         ToBeDetached state / detach in progress
11:03:39  azureDisk – detach disk list failed:          ctx deadline exceeded (4 disks)
11:04:22  DetachDisk pre-provisioned-readOnly:          timed out, retry with force detach
11:09:12  AttachDisk reattach-disk-multiple-nodes:      already attached to other node,
                                                        could not be attached to node(capz-2q2t-jlhv5)
```

Old behavior: 4 detach requests would queue on `sync.Mutex`, execute in order, and either succeed or fail at the ARM call — but the lock itself never bailed out.
New behavior: the requests whose per-request ctx expires while waiting for the semaphore **return early with an error**, before any ARM call is issued. The batch (`batchDetachDiskRequest` / `retrieveDetachBatchedDiskRequests`) is broken up, kubelet sees detach failures, and the follow-up attach on another node then fails with "already attached to other node".

### Why Windows and not Linux

Windows HostProcess is meaningfully slower than Linux on the CSI node RPC path (PowerShell-based `Get-Disk`, format, mount via CSI proxy). That makes it far more likely to burn through the caller's ctx deadline before the semaphore is acquired, especially when several disks contend for the same per-node lock. On Linux the same code path exists but the per-op latency is small enough that the new "give up on ctx" branch is rarely reached.

### Semantic concern (independent of the failure above)

The concern isn't the implementation of `lockMap` — that looks correct (semaphore weight 1, FIFO, no leak on failed acquire, unit tests cover cancel/timeout). The concern is that **detach is an operation that must eventually complete for the node's disk state to stay consistent**. Letting a per-request gRPC ctx cancel the wait for the node lock, so we return before ever calling ARM, changes the guarantee from "the request will be issued" to "the request may be dropped". kubelet will retry, but:

- The `batch*DiskRequest` maps rely on all callers reaching `retrieve*BatchedDiskRequests` under the same lock to compose a single VM update. If some callers exit early on `ctx.Err()`, their disks are stuck in the batch map without a `retrieve` sweep, until another request for that node comes in.
- Once a disk goes into `ToBeDetached` state on the VM but the caller thinks detach "failed", the subsequent attach to another node races the still-in-progress detach → the "already attached" cascade above.

### Suggestions to consider

1. **Decouple the lock ctx from the caller ctx** for `DetachDisk` / `UpdateVM` — e.g. bound it with a large absolute timeout (`context.WithTimeout(context.Background(), 5m)`) so the semaphore is fair and cancellable but not steered by a short-lived kubelet-side ctx.
2. Alternatively, keep the old blocking semantics specifically for the outer node lock in `DetachDisk` and only make the batch-map locks (`attachDiskMapKey`/`detachDiskMapKey`) ctx-aware — those are short critical sections where ctx cancel is much safer.
3. Confirm the batch maps are cleaned up on `LockEntry` failure — right now if `batchDetachDiskRequest` fails to acquire the map lock, the entry may have already been inserted on a previous call and never drained.

### Not fully proven

I can't tell from a single Windows log whether:

- the same failure existed before this PR and is a pre-existing flake (the new error strings would just be surfacing an existing race more visibly), or
- it's a real regression introduced by this PR.

Would be useful to compare against the last few Windows e2e runs on `master` after this PR merged. Happy to help dig if that's useful.

cc @jmclong @andyzhangx

PR #3703, opened 2026-07-08
[WIP]feat: sdk call to check VM exist during DetachDisk

hccheng72:
**What type of PR is this?**
/kind feature

Replace the InstanceID pre-check in DetachDisk with a purpose-built verifyInstanceForDetach that decides whether a detach is still needed without triggering the subscription-wide ListVirtualMachines call that InstanceID can cause in mixed-VM-type clusters.

What changed:
DetachDisk used to call `c.cloud.InstanceID(ctx, nodeName)` purely as an existence check ("if the host is gone, skip detach"). It now calls `verifyInstanceForDetach(ctx, diskURI, nodeName)`, which resolves the answer in the cheapest List-free way available and returns cloudprovider.InstanceNotFound or errDiskAlreadyDetached when the detach can be skipped (VM gone or disk already detached).

Decision order:

- Pure VMSS-uniform cluster (DisableAvailabilitySetNodes && !EnableVmssFlexNodes) → call InstanceID directly. In this config getVMManagementTypeByNodeName short-circuits, so InstanceID is fully cache-backed and never lists.

- Otherwise, providerID fast path → read the node's `spec.providerID`, parse it, and issue a single targeted GET:
1.VMSS uniform → GetVirtualMachineScaleSetVMClient().Get(rg, vmss, instanceID)
2.standalone/flex → GetVirtualMachineClient().Get(rg, name)

- providerID unavailable (e.g. Node object already deleted after the VM was removed) → fetch the disk and use ManagedBy/ManagedByExtended (authoritative, List-free): both empty or disk not found ⇒ errDiskAlreadyDetached; otherwise targeted GET on the attached VM.

- Only if the disk can't be read / ManagedBy can't be parsed → fall back to InstanceID.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azure_controller_common.go
+++ b/pkg/azuredisk/azure_controller_common.go
@@ -88,6 +88,12 @@ var defaultBackOff = kwait.Backoff{
 	Jitter:   0.0,
 }
 
+// errDiskAlreadyDetached signals that a fresh disk read authoritatively confirmed the
+// disk is not attached to any VM (ManagedBy and ManagedByExtended are empty) or the
+// disk no longer exists. In these cases DetachDisk can return immediately without
+// polling waitForDiskManagedByTobeRemoved, since there is no stale ManagedBy to clear.
+var errDiskAlreadyDetached = errors.New("disk already detached")
+
 type controllerCommon struct {
 	diskStateMap  sync.Map // <diskURI, attaching/detaching state>
 	lockMap       *lockMap
@@ -386,10 +392,17 @@ func (c *controllerCommon) retrieveAttachBatchedDiskRequests(ctx context.Context
 
 // DetachDisk detaches a disk from VM
 func (c *controllerCommon) DetachDisk(ctx context.Context, diskName, diskURI string, nodeName types.NodeName) error {
-	if _, err := c.cloud.InstanceID(ctx, nodeName); err != nil {
+	if err := c.verifyInstanceForDetach(ctx, diskURI, nodeName); err != nil {
+		if errors.Is(err, errDiskAlreadyDetached) {
+			// a fresh disk read already confirmed the disk is detached (disk gone or
+			// ManagedBy/ManagedByExtended empty), so there is no stale ManagedBy to wait on.
+			klog.V(2).Infof("azureDisk - DetachDisk(%s) confirmed disk already detached from node %s, skip wait", diskURI, nodeName)
+			return nil
+		}
 		if errors.Is(err, cloudprovider.InstanceNotFound) {
-			// if host doesn't exist, no need to detach
-			klog.Warningf("azureDisk - failed to get azure instance id(%s), DetachDisk(%s) will assume disk is already detached",
+			// the instance is gone, but the disk may still carry a stale ManagedBy that
+			// takes a while to clear, so wait for it to be removed before assuming detached.
+			klog.Warningf("azureDisk - azure instance id(%s) not found, DetachDisk(%s) will assume disk is already detached",
 				nodeName, diskURI)
 			return c.waitForDiskManagedByTobeRemoved(ctx, diskURI, nodeName)
 		}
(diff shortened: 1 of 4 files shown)


pkg/azuredisk/azure_controller_common.go
```diff
@@ -796,6 +815,140 @@ func (c *controllerCommon) waitForDiskManagedByTobeRemoved(ctx context.Context,
...
+		info, parseErr := parseProviderID(providerID)
+		if parseErr == nil {
+			return c.instanceExistsByResourceInfo(ctx, info)
+		}
+		klog.V(2).Infof("verifyInstanceForDetach: cannot parse providerID %q of node %s: %v", providerID, nodeName, parseErr)
+	} else {
+		klog.V(2).Infof("verifyInstanceForDetach: providerID unavailable for node %s: %v", nodeName, nodeErr)
+	}
+
+	// Fallback: the node/providerID is unavailable (e.g. the Node object was already
```
NedAnd1, line 938: If we failed to get node provider id, maybe we should have an explicit arm call for the VM to ensure it no longer exists & immediately return InstanceNotFound if it does not?
After that, we already wait for the managed by property of the disk to go away in `DetachDisk` so the managed by checks in this function are unnecessary.

pkg/azuredisk/azure_controller_common.go
```diff
@@ -796,6 +815,140 @@ func (c *controllerCommon) waitForDiskManagedByTobeRemoved(ctx context.Context,
...
+// providerID is unavailable (e.g. the Node object was already deleted after the VM was
+// removed), it falls back to the disk's ManagedBy (authoritative Azure attachment
+// state), which is still List-free. Only when neither the providerID nor the disk can
+// be read does it fall back to cloud.InstanceID.
+func (c *controllerCommon) verifyInstanceForDetach(ctx context.Context, diskURI string, nodeName types.NodeName) error {
+	// Pure VMSS-uniform cluster: InstanceID is cache-backed and never lists VMs
+	// subscription-wide (getVMManagementTypeByNodeName short-circuits), so it is the
+	// cheapest option in the common case.
+	if c.cloud.DisableAvailabilitySetNodes && !c.cloud.EnableVmssFlexNodes {
+		_, err := c.cloud.InstanceID(ctx, nodeName)
```
NedAnd1, line 922: Still forces an expensive cache refresh under a lock AFAIK

hccheng72: No, it will return [here](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/58cdff1c592ad8c2c63d9fbf0c43e01bbfd2c454/vendor/sigs.k8s.io/cloud-provider-azure/pkg/provider/azure_vmss_cache.go#L397-L399), never reach cache refresh. As noted in the comment, this is the cheapest option for pure VMSS-uniform clusters (0 ARM on cache hit).

pkg/azuredisk/azure_controller_common.go
hccheng72: If we failed to get node provider id, we don't know which client to call (GetVirtualMachineScaleSetVMClient() or GetVirtualMachineClient()) for the VM, we only can fallback to original path which could be expensive, we try get Disk additionally to avoid that.

But we can try to avoid duplicated ManagedBy check. That's a good point.

pkg/azuredisk/azure_controller_common.go
NedAnd1: `az.VMSet.GetInstanceIDByNodeName` (relevant func called by `cloud.InstanceId`) doesn't end there.
After `ss.getVMManagementTypeByNodeName` returns, `ss.getVmssVM` is called for a VMSS node which in turn calls `ss.getNodeIdentityByNodeName` & if a VMSS node is not found the following code path is reached: https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/58cdff1c592ad8c2c63d9fbf0c43e01bbfd2c454/vendor/sigs.k8s.io/cloud-provider-azure/pkg/provider/azure_vmss.go#L827-L828

hccheng72: Thanks for pointing this out — I went back through the code carefully. In this pure VMSS-uniform branch, getVMManagementTypeByNodeName returns early (before touching nonVmssUniformNodesCache), so the subscription-wide ListVirtualMachines is never reached. The only possible ARM List calls are VMSS-scoped cache refreshes (the list of VMSSes, or a single VMSS's VMs), and only on a cold/stale cache. In the common (warm-cache) case, calling cache-backed InstanceID could be the cheapest option.

Furthermore, the throttling issue this PR addresses was observed with standalone/mixed VM types (under Spot VM churn specifically), where InstanceID triggers the subscription-wide VM List during management-type classification. Since we haven't seen the same issue on the other branches (pure VMSS-uniform). IMO, I'd prefer to keep InstanceID as-is there, with no regression risk.

PR #3711, opened 2026-07-09
fix:  avoid using formatAndMount from kubernetes mount utils

mittachaitu:
**What type of PR is this?**
/kind bug

This PR avoids using FormatAndMount function from kubernetes mount-utils
and got replaced with following changes:
- Check existence of filesystem through blkid, fsck and wipefs with specific `no-op` options
    - If filesystem exist then run fsck and perform mount it.
    - If filesystem doesn't exist read filesystem signatures from secondary superblock and fail if fs signature already exists.
    - If filesystem doesn't even exist in secondary superblock then format and mount the disk at target path.
    Note: This PR also removes the force option while formatting the device with filesystem.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

In specific scenario csi driver ends in re-formatting the device.
T0: Application scheduled for the first time on node, as part of NodeStageVolume request csi driver formated and mounted the disk.
T1: Application written good amount of data but before flushing there is restart of node.
T2: Node came back online, pod got re-started/re-scheduled where csi-node driver received NodeStageVolume request for same volume. So, with existing flow CSI driver detects existence of filesystem and fsck.
T3: While fsck is in progress there is restart of node which ends in corrupting primary super blocks of filesystem (Assumption that fsck corrupted it)
T4: Once node back online, CSI driver received NodeStageVolume request as part of the request driver checked the existence of filesystem which shown empty. So, at this point disk formatted with ext4 filesystem which ends in wiping the previous data.

**This will not fully fix the issue, but it is an attempt to avoid running fsck**

**Release note**:
```
none
```

### DIFF
--- a/pkg/azuredisk/azure_common_linux.go
+++ b/pkg/azuredisk/azure_common_linux.go
@@ -26,6 +26,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc/codes"
@@ -33,10 +34,19 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/volume"
 	mount "k8s.io/mount-utils"
+	utilexec "k8s.io/utils/exec"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
 )
 
-const sysClassBlockPath = "/sys/class/block/"
+const (
+	sysClassBlockPath = "/sys/class/block/"
+	// 'fsck' found errors and corrected them
+	fsckErrorsCorrected = 1
+	// 'fsck' found errors but exited without correcting them
+	fsckErrorsUncorrected = 4
+	// 'fsck' found operational error, e.g fresh block device
+	fsckOperationalError = 8
+)
 
 // exclude those used by azure as resource and OS root in /dev/disk/azure, /dev/disk/azure/scsi0
 // "/dev/disk/azure/scsi0" dir is populated in Standard_DC4s/DC2s on Ubuntu 18.04
@@ -119,12 +129,132 @@ func findDiskByLun(lun int, io azureutils.IOHandler, _ *mount.SafeFormatAndMount
 	return "", fmt.Errorf("failed to find disk by lun %d", lun)
 }
 
-func formatAndMount(source, target, fstype string, options []string, m *mount.SafeFormatAndMount) error {
+func formatAndMount(source, target, fstype string, options []string, m *mount.SafeFormatAndMount, formatSem chan any, formatTimeout time.Duration) error {
 	if newOptions, exists := azureutils.RemoveOptionIfExists(options, "directmount"); exists {
 		klog.V(2).Infof("formatAndMount - skip format for %s, old options: %v, new options: %v", target, options, newOptions)
 		return m.Mount(source, target, fstype, newOptions)
 	}
-	return m.FormatAndMount(source, target, fstype, options)
+
+	readOnly := false
+	for _, option := range options {
+		if option == "ro" {
+			readOnly = true
+			break
+		}
+	}
+
+	// Only single source of truth to know whether the disk is formatted or not is through blkid.
+	diskFSFormat, err := m.GetDiskFormat(source)
+	if err != nil {
+		klog.Errorf("formatAndMount - failed to get disk format for %s with error(%v)", source, err)
+		return fmt.Errorf("failed to get disk format for %s with error(%v)", source, err)
+	}
+
+	if diskFSFormat == "" {
+		// As part of auto-recovery from mount failures, we run fsck on the disk. If the disk was pulled
+		// out or a power loss occurred during a previous fsck run, the primary superblocks may have been
+		// corrupted depending on the stage at which fsck was interrupted. We run fsck here to detect and
+		// repair such issues before formatting.
+		// Note: fsck is a no-op if the disk already has an xfs signature.
+		// If filesystem was detected at this stage don't format return the error
+		isFilesystemExist, err := detectFilesystemExistence(source, m)
+		if err != nil {
+			klog.Errorf("formatAndMount - failed to check existence of filesystem on disk %s with error(%v)", source, err)
+			return fmt.Errorf("formatAndMount - failed to check existence of filesystem on disk %s with error(%v)", source, err)
+		}
+
+		if !isFilesystemExist {
+			if readOnly {
+				return fmt.Errorf("formatAndMount - disk %s is not formatted, but mount options specify read-only", source)
+			}
+
+			// If filesystem doesn't exist then we can format the disk with the given fstype.
+			// Disk is unformatted
+			args := []string{source}
+			if fstype == "ext4" || fstype == "ext3" {
+				args = []string{
+					"-m0", // Zero blocks reserved for super-user
+					source,
+				}
+			} else if fstype == "xfs" {
+				args = []string{
+					source,
+				}
+			}
+
+			klog.Infof("Disk %q appears to be unformatted, attempting to format as type: %q with options: %v", source, fstype, args)
+			output, err := mkfsWithConcurrencyLimit(m, formatSem, formatTimeout, fstype, args)
+			if err != nil {
+				klog.Errorf("formatAndMount - failed to format disk %s as type %s with options %v: %v error: %v", source, fstype, args, string(output), err)
+				return fmt.Errorf("failed to format disk %s as type %s with options %v: %v error: %v", source, fstype, args, string(output), err)
+			}
+			// Format was successful
+			diskFSFormat = fstype
+		} else {
+			// Re-read the filesystem signature to determine the existing format.
+			reReadFormat, err := m.GetDiskFormat(source)
+			if err != nil {
+				klog.Errorf("formatAndMount - failed to re-read filesystem signature on disk %s with error: %v", source, err)
+				return fmt.Errorf("failed to re-read filesystem signature on disk %s with error: %v", source, err)
+			}
+			klog.Infof("Disk %q contains filesystem signature post running fsck, skipping format; detected filesystem type: %q", source, reReadFormat)
+			diskFSFormat = reReadFormat
+		}
+	}
+
+	if !strings.EqualFold(fstype, diskFSFormat) {
+		// Verify that the disk is formatted with filesystem type we are expecting
+		klog.Errorf("formatAndMount - configured to mount disk %s as %s but current format is %s, things might break", source, fstype, diskFSFormat)
+		return fmt.Errorf("configured to mount disk %s as %s but current format is %s, things might break", source, fstype, diskFSFormat)
+	}
+
+	// Running fsck on the disk to detect and repair any filesystem issues before mounting
+	if !readOnly {
+		_, err := detectAndRepairFilesystem(source, []string{"-a"}, m)
+		if err != nil {
+			klog.Errorf("formatAndMount - failed to run fsck on disk %s with error: %v", source, err)
+			return fmt.Errorf("failed to run fsck on disk %s with error: %v", source, err)
+		}
+	}
+
+	klog.V(4).Infof("Attempting to mount disk %s in %s format at %s", source, fstype, target)
+	options = append(options, "defaults")
+	if err := m.Mount(source, target, fstype, options); err != nil {
+		klog.Errorf("formatAndMount - failed to mount disk %s at %s with options %v and error: %v", source, target, options, err)
+		return fmt.Errorf("failed to mount disk %s at %s with options %v and error: %v", source, target, options, err)
+	}
+
+	return nil
+}
+
+// mkfsWithConcurrencyLimit runs mkfs.<fstype> while honoring the max-concurrent-format
+// semaphore. Because this driver formats the disk with a direct exec call (instead of
+// mount.SafeFormatAndMount.FormatAndMount) to insert the pre-format fsck auto-recovery
+// step, it must reproduce SafeFormatAndMount's WithMaxConcurrentFormat gating here to
+// avoid unbounded concurrent mkfs operations under load. The logic mirrors the upstream
+// (unexported) SafeFormatAndMount.format method: once formatTimeout elapses the token is
+// released so another format can proceed, while the original mkfs is allowed to complete.
+func mkfsWithConcurrencyLimit(m *mount.SafeFormatAndMount, formatSem chan any, formatTimeout time.Duration, fstype string, args []string) ([]byte, error) {
+	if formatSem != nil {
+		done := make(chan struct{})
+		defer close(done)
+
+		formatSem <- struct{}{}
+
+		go func() {
+			defer func() { <-formatSem }()
+
+			timeout := time.NewTimer(formatTimeout)
+			defer timeout.Stop()
+
+			select {
+			case <-done:
+			case <-timeout.C:
+			}
+		}()
+	}
+
+	return m.Exec.Command("mkfs."+fstype, args...).CombinedOutput()
 }
 
 // finds a device mounted to "current" node
@@ -289,6 +419,76 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
 		}
 	}
 	return nil
+}
+
+// detectFilesystemExistence checks whether the given device already contains a filesystem signature.
+// 2. Detects filesystem signature using wipefs --no-act --noheadings --output TYPE <device>.
+func detectFilesystemExistence(source string, mounter *mount.SafeFormatAndMount) (bool, error) {
+	isFSExist, err := detectAndRepairFilesystem(source, []string{"-n"}, mounter)
+	if err != nil {
+		klog.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'fsck' with error(%v)", source, err)
+		return false, fmt.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'fsck' with error(%v)", source, err)
+	}
+	if isFSExist {
+		return true, nil
+	}
+
+	args := []string{"--no-act", "--output", "TYPE", "--noheadings", source}
+	fsType, err := mounter.Exec.Command("wipefs", args...).CombinedOutput()
+	if err != nil {
+		klog.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'wipefs' with error(%v)", source, err)
+		return false, fmt.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'wipefs' with error(%v)", source, err)
+	}
+	if len(fsType) > 0 {
+		return true, nil
+	}
+	return false, nil
+}
+
+// detectAndRepairFilesystem runs fsck on the given device to determine whether it already contains
+// a filesystem and to repair any recoverable inconsistencies in place based on provided options.
+//
+// It returns:
+//   - (true, nil): a filesystem is present (clean or recoverable errors corrected).
+//   - (false, nil): no filesystem signature (fresh/unformatted device).
+//   - (false, error): fsck could not be executed or returned an operational/ambiguous failure.
+//   - (true, error): fsck found errors it could not correct.
+//
+// fsckOptions control fsck's behavior, e.g. "-n" for a read-only detection check or "-y"/"-E ..."
+// to attempt repairs.
+func detectAndRepairFilesystem(source string, fsckOptions []string, mounter *mount.SafeFormatAndMount) (bool, error) {
+	klog.V(2).Infof("Checking for issues with fsck on disk: %s with options %v", source, fsckOptions)
+	args := append(append([]string(nil), fsckOptions...), source)
+	out, err := mounter.Exec.Command("fsck", args...).CombinedOutput()
+	if err != nil {
+		ee, isExitError := err.(utilexec.ExitError)
+		switch {
+		case err == utilexec.ErrExecutableNotFound:
+			klog.Errorf("'fsck' not found on system; cannot verify filesystem signature on %s, returning error.", source)
+			return false, fmt.Errorf("'fsck' not found to detect filesystem on device %s with options %v: %v", source, fsckOptions, err)
+		case isExitError && ee.ExitStatus() == fsckOperationalError:
+			if strings.Contains(strings.ToLower(string(out)), "superblock could not be read or does not describe a valid ext2/ext3/ext4") {
+				klog.Infof("Device %s may be fresh block device with no filesystem signature, fsck output: %s", source, string(out))
+				return false, nil
+			}
+			klog.Warningf("Unable to run fsck on device %s with options %v, fsck output: %s", source, fsckOptions, string(out))
+			return false, fmt.Errorf("Unable to run fsck on device %s with options %v, fsck error: %v", source, fsckOptions, err)
+		case isExitError && ee.ExitStatus() == fsckErrorsCorrected:
+			klog.Warningf("Device %s has errors which were corrected by fsck: %s", source, string(out))
+		case isExitError && ee.ExitStatus() == fsckErrorsUncorrected:
+			// Filesystem exists but fsck found errors that it could not correct
+			klog.Errorf("Device %s has errors which fsck could not correct with options %v: %s", source, fsckOptions, string(out))
+			return true, fmt.Errorf("'fsck' found errors on device %s with options %v but could not correct them (exit status %d)", source, fsckOptions, ee.ExitStatus())
+		case isExitError && ee.ExitStatus() > fsckErrorsUncorrected:
+			klog.Errorf("`fsck` error %s", string(out))
+			return false, fmt.Errorf("'fsck' failed on device %s with options %v: %v", source, fsckOptions, err)
+		default:
+			klog.Warningf("fsck on device %s failed with error %v, output: %v", source, err, string(out))
+		}
+	}
+	// In case if device is formatted with other filesystem, fsck will return 0
+	klog.Infof("fsck on device %s completed successfully with output: %s", source, string(out))
+	return true, nil
 }
 
 func (d *Driver) GetVolumeStats(_ context.Context, m *mount.SafeFormatAndMount, _, target string, hostutil hostUtil) ([]*csi.VolumeUsage, error) {
--- a/pkg/azuredisk/azure_common_linux.go
+++ b/pkg/azuredisk/azure_common_linux.go
@@ -173,11 +173,13 @@ func formatAndMount(source, target, fstype string, options []string, m *mount.Sa
 			args := []string{source}
 			if fstype == "ext4" || fstype == "ext3" {
 				args = []string{
+					"-F",  // Force flag, if we reached this point then mostly it's a fresh disk without any fs signature
 					"-m0", // Zero blocks reserved for super-user
 					source,
 				}
 			} else if fstype == "xfs" {
 				args = []string{
+					"-f",
 					source,
 				}
 			}
@@ -439,7 +441,7 @@ func detectFilesystemExistence(source string, mounter *mount.SafeFormatAndMount)
 		klog.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'wipefs' with error(%v)", source, err)
 		return false, fmt.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'wipefs' with error(%v)", source, err)
 	}
-	if len(fsType) > 0 {
+	if len(strings.TrimSpace(string(fsType))) > 0 {
 		return true, nil
 	}
 	return false, nil
--- a/pkg/azuredisk/azure_common_linux.go
+++ b/pkg/azuredisk/azure_common_linux.go
@@ -428,8 +428,12 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
 func detectFilesystemExistence(source string, mounter *mount.SafeFormatAndMount) (bool, error) {
 	isFSExist, err := detectAndRepairFilesystem(source, []string{"-n"}, mounter)
 	if err != nil {
-		klog.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'fsck' with error(%v)", source, err)
-		return false, fmt.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'fsck' with error(%v)", source, err)
+		if strings.Contains(strings.ToLower(err.Error()), "superblock could not be read or does not describe a valid ext2/ext3/ext4") {
+			klog.Infof("Device %s may be fresh block device with no filesystem signature, fsck output: %s", source, err.Error())
+		} else {
+			klog.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'fsck' with error(%v)", source, err)
+			return false, fmt.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'fsck' with error(%v)", source, err)
+		}
 	}
 	if isFSExist {
 		return true, nil
@@ -469,12 +473,8 @@ func detectAndRepairFilesystem(source string, fsckOptions []string, mounter *mou
 			klog.Errorf("'fsck' not found on system; cannot verify filesystem signature on %s, returning error.", source)
 			return false, fmt.Errorf("'fsck' not found to detect filesystem on device %s with options %v: %v", source, fsckOptions, err)
 		case isExitError && ee.ExitStatus() == fsckOperationalError:
-			if strings.Contains(strings.ToLower(string(out)), "superblock could not be read or does not describe a valid ext2/ext3/ext4") {
-				klog.Infof("Device %s may be fresh block device with no filesystem signature, fsck output: %s", source, string(out))
-				return false, nil
-			}
 			klog.Warningf("Unable to run fsck on device %s with options %v, fsck output: %s", source, fsckOptions, string(out))
-			return false, fmt.Errorf("Unable to run fsck on device %s with options %v, fsck error: %v", source, fsckOptions, err)
+			return false, fmt.Errorf("Unable to run fsck on device %s with options %v, fsck error: %v output: %s", source, fsckOptions, err, string(out))
 		case isExitError && ee.ExitStatus() == fsckErrorsCorrected:
 			klog.Warningf("Device %s has errors which were corrected by fsck: %s", source, string(out))
 		case isExitError && ee.ExitStatus() == fsckErrorsUncorrected:
--- a/pkg/azuredisk/azure_common_linux_test.go
+++ b/pkg/azuredisk/azure_common_linux_test.go
@@ -1,3 +1,6 @@
+//go:build linux
+// +build linux
+
 /*
 Copyright 2022 The Kubernetes Authors.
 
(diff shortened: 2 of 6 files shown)


landreasyan:
My main concern is that if blkid returns empty, we will still incorrectly format the disk which can result in data loss. Should that part be updated to a fallback logic like below: 

```
existing := blkid -p (TYPE + PTTYPE)      # settle + retry on exit 2
if existing != "" or PTTYPE present:      # formatted / has partitions
    mount, don't format
else:
    # exit-2: ambiguous — verify it's genuinely blank
    if wipefs -n finds any signature
       OR ext backup superblock found:
        FAIL: "device appears to contain data; refusing to format"  # do NOT wipe
    else:
        format only if this is expected to be a fresh disk
```

pkg/azuredisk/azure_common_linux.go
```diff
@@ -124,7 +124,66 @@ func formatAndMount(source, target, fstype string, options []string, m *mount.Sa
...
+			break
+		}
+	}
+
+	// Only single source of truth to know whether the disk is formatted or not is through blkid.
+	existingFormat, err := m.GetDiskFormat(source)
+	if err != nil {
+		klog.Errorf("formatAndMount - failed to get disk format for %s with error(%v)", source, err)
+		return fmt.Errorf("failed to get disk format for %s with error(%v)", source, err)
+	}
```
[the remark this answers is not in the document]
mittachaitu: Updated the description

mittachaitu:
> existing := blkid -p (TYPE + PTTYPE)      # settle + retry on exit 2
> if existing != "" or PTTYPE present:      # formatted / has partitions
>     mount, don't format
> else:
>     # exit-2: ambiguous — verify it's genuinely blank
>     if wipefs -n finds any signature
>        OR ext backup superblock found:
>         FAIL: "device appears to contain data; refusing to format"  # do NOT wipe
>     else:
>         format only if this is expected to be a fresh disk

I thought about it, since we are not executing `fsck` with the changes there shouldn't be any corruption of primary superblock and `blkid -p` will read the primary superblock information. 

btw.. even `wipefs -n` uses libblkid which again reads information from primary superblocks.

mittachaitu:
> > existing := blkid -p (TYPE + PTTYPE)      # settle + retry on exit 2
> > if existing != "" or PTTYPE present:      # formatted / has partitions
> > mount, don't format
> > else:
> > # exit-2: ambiguous — verify it's genuinely blank
> > if wipefs -n finds any signature
> > OR ext backup superblock found:
> > FAIL: "device appears to contain data; refusing to format"  # do NOT wipe
> > else:
> > format only if this is expected to be a fresh disk
> 
> I thought about it, since we are not executing `fsck` with the changes there shouldn't be any corruption of primary superblock and `blkid -p` will read the primary superblock information.
> 
> btw.. even `wipefs -n` uses libblkid which again reads information from primary superblocks.


@landreasyan I see only way to read from secondary super block is using fsck:
- Detect potential secondary superblocks of given disk (through `mkfs.ext4 -n`)
- Use `fsck -n -b <secondary_superblock_number>` to detect filesystem existance

```sh
# Format the device
mittas@mittas-devbox:~/acstor/k8s/disk$ sudo mkfs.ext4 /dev/loop0
mke2fs 1.47.0 (5-Feb-2023)
Discarding device blocks: done
Creating filesystem with 2621440 4k blocks and 655360 inodes
Filesystem UUID: 048f90b0-cb3e-4264-a6a1-cc523fdee906
Superblock backups stored on blocks:
        32768, 98304, 163840, 229376, 294912, 819200, 884736, 1605632

Allocating group tables: done
Writing inode tables: done
Creating journal (16384 blocks): done
Writing superblocks and filesystem accounting information: done

# Earsed primary superblock content
mittas@mittas-devbox:~/acstor/k8s/disk$ sudo wipefs -fa /dev/loop0
/dev/loop0: 2 bytes were erased at offset 0x00000438 (ext4): 53 ef

# Neither blkid/wipefs able to detect filesystem existance
mittas@mittas-devbox:~/acstor/k8s/disk$ sudo blkid -p /dev/loop0
mittas@mittas-devbox:~/acstor/k8s/disk$ sudo wipefs -n /dev/loop0

# Checking from secondary superblock to detect filesystem
mittas@mittas-devbox:~/acstor/k8s/disk$ sudo fsck.ext2 -n -b 32768 /dev/loop0
e2fsck 1.47.0 (5-Feb-2023)
/dev/loop0 was not cleanly unmounted, check forced.
Pass 1: Checking inodes, blocks, and sizes
Pass 2: Checking directory structure
Pass 3: Checking directory connectivity
Pass 4: Checking reference counts
Pass 5: Checking group summary information
Block bitmap differences:  +(32768--33794) +(98304--99330) +(163840--164866) +(229376--230402) +(294912--295938) +(819200--820226) +(884736--885762) +(1605632--1606658)
Fix? no

Free blocks count wrong for group #1 (31741, counted=32768).
Fix? no

Free blocks count wrong for group #3 (31741, counted=32768).
Fix? no

Free blocks count wrong for group #5 (31741, counted=32768).
Fix? no

Free blocks count wrong for group #7 (31741, counted=32768).
Fix? no

Free blocks count wrong for group #9 (31741, counted=32768).
Fix? no

Free blocks count wrong for group #25 (31741, counted=32768).
Fix? no

Free blocks count wrong for group #27 (31741, counted=32768).
Fix? no

Free blocks count wrong for group #49 (31741, counted=32768).
Fix? no

Free blocks count wrong (2554687, counted=2562903).
Fix? no

Padding at end of inode bitmap is not set. Fix? no

Inode bitmap differences: Group 1 inode bitmap does not match checksum.
IGNORED.
Group 2 inode bitmap does not match checksum.
IGNORED.
Group 3 inode bitmap does not match checksum.
IGNORED.
Group 4 inode bitmap does not match checksum.
IGNORED.
Group 5 inode bitmap does not match checksum.
IGNORED.
Group 6 inode bitmap does not match checksum.
IGNORED.
Group 7 inode bitmap does not match checksum.
IGNORED.
Group 8 inode bitmap does not match checksum.
IGNORED.
Group 9 inode bitmap does not match checksum.
IGNORED.
Group 10 inode bitmap does not match checksum.
IGNORED.
Group 11 inode bitmap does not match checksum.
IGNORED.
Group 12 inode bitmap does not match checksum.
IGNORED.
Group 13 inode bitmap does not match checksum.
IGNORED.
Group 14 inode bitmap does not match checksum.
IGNORED.
Group 15 inode bitmap does not match checksum.
IGNORED.
Group 16 inode bitmap does not match checksum.
IGNORED.
Group 17 inode bitmap does not match checksum.
IGNORED.
Group 18 inode bitmap does not match checksum.
IGNORED.
Group 19 inode bitmap does not match checksum.
IGNORED.
Group 20 inode bitmap does not match checksum.
IGNORED.
Group 21 inode bitmap does not match checksum.
IGNORED.
Group 22 inode bitmap does not match checksum.
IGNORED.
Group 23 inode bitmap does not match checksum.
IGNORED.
Group 24 inode bitmap does not match checksum.
IGNORED.
Group 25 inode bitmap does not match checksum.
IGNORED.
Group 26 inode bitmap does not match checksum.
IGNORED.
Group 27 inode bitmap does not match checksum.
IGNORED.
Group 28 inode bitmap does not match checksum.
IGNORED.
Group 29 inode bitmap does not match checksum.
IGNORED.
Group 30 inode bitmap does not match checksum.
IGNORED.
Group 31 inode bitmap does not match checksum.
IGNORED.
Group 32 inode bitmap does not match checksum.
IGNORED.
Group 33 inode bitmap does not match checksum.
IGNORED.
Group 34 inode bitmap does not match checksum.
IGNORED.
Group 35 inode bitmap does not match checksum.
IGNORED.
Group 36 inode bitmap does not match checksum.
IGNORED.
Group 37 inode bitmap does not match checksum.
IGNORED.
Group 38 inode bitmap does not match checksum.
IGNORED.
Group 39 inode bitmap does not match checksum.
IGNORED.
Group 40 inode bitmap does not match checksum.
IGNORED.
Group 41 inode bitmap does not match checksum.
IGNORED.
Group 42 inode bitmap does not match checksum.
IGNORED.
Group 43 inode bitmap does not match checksum.
IGNORED.
Group 44 inode bitmap does not match checksum.
IGNORED.
Group 45 inode bitmap does not match checksum.
IGNORED.
Group 46 inode bitmap does not match checksum.
IGNORED.
Group 47 inode bitmap does not match checksum.
IGNORED.
Group 48 inode bitmap does not match checksum.
IGNORED.
Group 49 inode bitmap does not match checksum.
IGNORED.
Group 50 inode bitmap does not match checksum.
IGNORED.
Group 51 inode bitmap does not match checksum.
IGNORED.
Group 52 inode bitmap does not match checksum.
IGNORED.
Group 53 inode bitmap does not match checksum.
IGNORED.
Group 54 inode bitmap does not match checksum.
IGNORED.
Group 55 inode bitmap does not match checksum.
IGNORED.
Group 56 inode bitmap does not match checksum.
IGNORED.
Group 57 inode bitmap does not match checksum.
IGNORED.
Group 58 inode bitmap does not match checksum.
IGNORED.
Group 59 inode bitmap does not match checksum.
IGNORED.
Group 60 inode bitmap does not match checksum.
IGNORED.
Group 61 inode bitmap does not match checksum.
IGNORED.
Group 62 inode bitmap does not match checksum.
IGNORED.
Group 63 inode bitmap does not match checksum.
IGNORED.
Group 64 inode bitmap does not match checksum.
IGNORED.
Group 65 inode bitmap does not match checksum.
IGNORED.
Group 66 inode bitmap does not match checksum.
IGNORED.
Group 67 inode bitmap does not match checksum.
IGNORED.
Group 68 inode bitmap does not match checksum.
IGNORED.
Group 69 inode bitmap does not match checksum.
IGNORED.
Group 70 inode bitmap does not match checksum.
IGNORED.
Group 71 inode bitmap does not match checksum.
IGNORED.
Group 72 inode bitmap does not match checksum.
IGNORED.
Group 73 inode bitmap does not match checksum.
IGNORED.
Group 74 inode bitmap does not match checksum.
IGNORED.
Group 75 inode bitmap does not match checksum.
IGNORED.
Group 76 inode bitmap does not match checksum.
IGNORED.
Group 77 inode bitmap does not match checksum.
IGNORED.
Group 78 inode bitmap does not match checksum.
IGNORED.
Group 79 inode bitmap does not match checksum.
IGNORED.
Block bitmap differences: Group 1 block bitmap does not match checksum.
IGNORED.
Group 2 block bitmap does not match checksum.
IGNORED.
Group 3 block bitmap does not match checksum.
IGNORED.
Group 4 block bitmap does not match checksum.
IGNORED.
Group 5 block bitmap does not match checksum.
IGNORED.
Group 6 block bitmap does not match checksum.
IGNORED.
Group 7 block bitmap does not match checksum.
IGNORED.
Group 8 block bitmap does not match checksum.
IGNORED.
Group 9 block bitmap does not match checksum.
IGNORED.
Group 10 block bitmap does not match checksum.
IGNORED.
Group 11 block bitmap does not match checksum.
IGNORED.
Group 12 block bitmap does not match checksum.
IGNORED.
Group 13 block bitmap does not match checksum.
IGNORED.
Group 14 block bitmap does not match checksum.
IGNORED.
Group 15 block bitmap does not match checksum.
IGNORED.
Group 17 block bitmap does not match checksum.
IGNORED.
Group 18 block bitmap does not match checksum.
IGNORED.
Group 19 block bitmap does not match checksum.
IGNORED.
Group 20 block bitmap does not match checksum.
IGNORED.
Group 21 block bitmap does not match checksum.
IGNORED.
Group 22 block bitmap does not match checksum.
IGNORED.
Group 23 block bitmap does not match checksum.
IGNORED.
Group 24 block bitmap does not match checksum.
IGNORED.
Group 25 block bitmap does not match checksum.
IGNORED.
Group 26 block bitmap does not match checksum.
IGNORED.
Group 27 block bitmap does not match checksum.
IGNORED.
Group 28 block bitmap does not match checksum.
IGNORED.
Group 29 block bitmap does not match checksum.
IGNORED.
Group 30 block bitmap does not match checksum.
IGNORED.
Group 31 block bitmap does not match checksum.
IGNORED.
Group 34 block bitmap does not match checksum.
IGNORED.
Group 35 block bitmap does not match checksum.
IGNORED.
Group 36 block bitmap does not match checksum.
IGNORED.
Group 37 block bitmap does not match checksum.
IGNORED.
Group 38 block bitmap does not match checksum.
IGNORED.
Group 39 block bitmap does not match checksum.
IGNORED.
Group 40 block bitmap does not match checksum.
IGNORED.
Group 41 block bitmap does not match checksum.
IGNORED.
Group 42 block bitmap does not match checksum.
IGNORED.
Group 43 block bitmap does not match checksum.
IGNORED.
Group 44 block bitmap does not match checksum.
IGNORED.
Group 45 block bitmap does not match checksum.
IGNORED.
Group 46 block bitmap does not match checksum.
IGNORED.
Group 47 block bitmap does not match checksum.
IGNORED.
Group 49 block bitmap does not match checksum.
IGNORED.
Group 50 block bitmap does not match checksum.
IGNORED.
Group 51 block bitmap does not match checksum.
IGNORED.
Group 52 block bitmap does not match checksum.
IGNORED.
Group 53 block bitmap does not match checksum.
IGNORED.
Group 54 block bitmap does not match checksum.
IGNORED.
Group 55 block bitmap does not match checksum.
IGNORED.
Group 56 block bitmap does not match checksum.
IGNORED.
Group 57 block bitmap does not match checksum.
IGNORED.
Group 58 block bitmap does not match checksum.
IGNORED.
Group 59 block bitmap does not match checksum.
IGNORED.
Group 60 block bitmap does not match checksum.
IGNORED.
Group 61 block bitmap does not match checksum.
IGNORED.
Group 62 block bitmap does not match checksum.
IGNORED.
Group 63 block bitmap does not match checksum.
IGNORED.
Group 65 block bitmap does not match checksum.
IGNORED.
Group 66 block bitmap does not match checksum.
IGNORED.
Group 67 block bitmap does not match checksum.
IGNORED.
Group 68 block bitmap does not match checksum.
IGNORED.
Group 69 block bitmap does not match checksum.
IGNORED.
Group 70 block bitmap does not match checksum.
IGNORED.
Group 71 block bitmap does not match checksum.
IGNORED.
Group 72 block bitmap does not match checksum.
IGNORED.
Group 73 block bitmap does not match checksum.
IGNORED.
Group 74 block bitmap does not match checksum.
IGNORED.
Group 75 block bitmap does not match checksum.
IGNORED.
Group 76 block bitmap does not match checksum.
IGNORED.
Group 77 block bitmap does not match checksum.
IGNORED.
Group 78 block bitmap does not match checksum.
IGNORED.

/dev/loop0: ********** WARNING: Filesystem still has errors **********

/dev/loop0: 11/655360 files (0.0% non-contiguous), 66753/2621440 blocks

# Fix filesystem by making changes

mittas@mittas-devbox:~/acstor/k8s/disk$ sudo fsck -b 32768 /dev/loop0
fsck from util-linux 2.39.3
e2fsck 1.47.0 (5-Feb-2023)
/dev/loop0 was not cleanly unmounted, check forced.
Pass 1: Checking inodes, blocks, and sizes
Pass 2: Checking directory structure
Pass 3: Checking directory connectivity
Pass 4: Checking reference counts
Pass 5: Checking group summary information
Block bitmap differences:  +(32768--33794) +(98304--99330) +(163840--164866) +(229376--230402) +(294912--295938) +(819200--820226) +(884736--885762) +(1605632--1606658)
Fix<y>? yes
Padding at end of inode bitmap is not set. Fix<y>? yes

/dev/loop0: ***** FILE SYSTEM WAS MODIFIED *****
/dev/loop0: 11/655360 files (0.0% non-contiguous), 66753/2621440 blocks

# Able to recover
mittas@mittas-devbox:~/acstor/k8s/disk$ sudo blkid -p /dev/loop0
/dev/loop0: UUID="048f90b0-cb3e-4264-a6a1-cc523fdee906" VERSION="1.0" FSBLOCKSIZE="4096" BLOCK_SIZE="4096" FSLASTBLOCK="2621440" FSSIZE="10737418240" TYPE="ext4" USAGE="filesystem"
mittas@mittas-devbox:~/acstor/k8s/disk$
```

pkg/azuredisk/azure_common_linux.go
```diff
@@ -124,7 +124,65 @@ func formatAndMount(source, target, fstype string, options []string, m *mount.Sa
...
+			klog.Errorf("formatAndMount - configured to mount disk %s as %s but current format is %s, things might break", source, fstype, existingFormat)
+			return fmt.Errorf("configured to mount disk %s as %s but current format is %s, things might break", source, fstype, existingFormat)
+		}
+	}
+
+	// Mount the disk
+	klog.V(4).Infof("Attempting to mount disk %s in %s format at %s", source, fstype, target)
+	options = append(options, "defaults")
+	if err := m.Mount(source, target, fstype, options); err != nil {
+		klog.Errorf("formatAndMount - failed to mount disk %s at %s with options %v and error: %v", source, target, options, err)
```
pawanpraka1, line 225: as part of mount failure, we should do fsck, a lot of users would not be able to mount the volumes after ungraceful shutdown.

pkg/azuredisk/azure_common_linux.go
```diff
@@ -124,7 +124,65 @@ func formatAndMount(source, target, fstype string, options []string, m *mount.Sa
...
+	}
+
+	// Only single source of truth to know whether the disk is formatted or not is through blkid.
+	existingFormat, err := m.GetDiskFormat(source)
+	if err != nil {
+		klog.Errorf("formatAndMount - failed to get disk format for %s with error(%v)", source, err)
+		return fmt.Errorf("failed to get disk format for %s with error(%v)", source, err)
+	}
+
+	if existingFormat == "" {
```
pawanpraka1: disk was in bad state and we upgraded to this verison of the driver. We would wipe the data. This is a corner case and should be rare. Maybe worth making a note.

pkg/azuredisk/azure_common_linux.go
mittachaitu: Pawan, again we may end in bad situation if there is powerless while running fsck right while making changes to primary superblock?

pkg/azuredisk/azure_common_linux.go
mittachaitu: Do you suggested reading from information from secondary superblock?

pkg/azuredisk/azure_common_linux.go
pawanpraka1: yes, we could, so what I would suggest is, run fsck when we are not able to find existingfilesystem before doing mkfs. If there was a superblock corruption, fsck would fix it.

pkg/azuredisk/azure_common_linux.go
pawanpraka1: doing fsck when existingfilesystem is not found should fix it.

pkg/azuredisk/azure_common_linux.go
mittachaitu: Updated the code

pkg/azuredisk/azure_common_linux.go
mittachaitu: updated the code, PTAL

pkg/azuredisk/azure_common_linux.go
```diff
@@ -291,6 +381,34 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
...
+	klog.V(2).Infof("Checking for issues with fsck on disk: %s with options %v", source, fsckOptions)
+	args := append(fsckOptions, source)
+	out, err := mounter.Exec.Command("fsck", args...).CombinedOutput()
+	if err != nil {
+		ee, isExitError := err.(utilexec.ExitError)
+		switch {
+		case err == utilexec.ErrExecutableNotFound:
+			klog.Warningf("'fsck' not found on system; continuing without running 'fsck'.")
+			return false, nil
+		case isExitError && ee.ExitStatus() == fsckOperationalError:
```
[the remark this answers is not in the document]
mittachaitu: Updated as true

pkg/azuredisk/azure_common_linux_test.go
```diff
@@ -17,16 +20,220 @@ limitations under the License.
...
+			fsckErr:     testingexec.FakeExitError{Status: 16},
+			wantFsExist: true,
+			wantErr:     false,
+		},
+		{
+			name:        "fsck binary not found",
+			fsckErr:     exec.ErrExecutableNotFound,
+			wantFsExist: false,
+			wantErr:     false,
+		},
```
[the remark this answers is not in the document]
mittachaitu: Updated

pkg/azuredisk/azure_common_linux.go
```diff
@@ -124,7 +133,98 @@ func formatAndMount(source, target, fstype string, options []string, m *mount.Sa
...
+			}
+
+			// If filesystem doesn't exist then we can format the disk with the given fstype.
+			// Disk is unformatted
+			args := []string{source}
+			if fstype == "ext4" || fstype == "ext3" {
+				args = []string{
+					"-m0", // Zero blocks reserved for super-user
+					source,
+				}
```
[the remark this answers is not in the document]
mittachaitu: Intentionally dropped the `force` flag

pkg/azuredisk/azure_common_linux.go
```diff
@@ -124,7 +133,98 @@ func formatAndMount(source, target, fstype string, options []string, m *mount.Sa
...
+			} else if fstype == "xfs" {
+				args = []string{
+					source,
+				}
```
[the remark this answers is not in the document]
mittachaitu: Intentionally dropped the flag

pkg/azuredisk/azure_common_linux.go
```diff
@@ -124,7 +133,98 @@ func formatAndMount(source, target, fstype string, options []string, m *mount.Sa
...
+			klog.Infof("Disk %q contains filesystem signature post running fsck, skipping format as type: %q", source, fstype)
+			diskFSFormat = reReadFormat
+		}
+	}
+
+	if fstype != diskFSFormat {
+		// Verify that the disk is formatted with filesystem type we are expecting
+		klog.Errorf("formatAndMount - configured to mount disk %s as %s but current format is %s, things might break", source, fstype, diskFSFormat)
+		return fmt.Errorf("configured to mount disk %s as %s but current format is %s, things might break", source, fstype, diskFSFormat)
+	}
```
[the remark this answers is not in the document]
mittachaitu: It is always to be double sure about the filesystem type instead of allowing it with warning

pkg/azuredisk/azure_common_linux.go
```diff
@@ -124,7 +133,98 @@ func formatAndMount(source, target, fstype string, options []string, m *mount.Sa
...
+
+	// Mount the disk
+	klog.V(4).Infof("Attempting to mount disk %s in %s format at %s", source, fstype, target)
+	options = append(options, "defaults")
+	if err := m.Mount(source, target, fstype, options); err != nil {
+		klog.Errorf("formatAndMount - failed to mount disk %s at %s with options %v and error: %v", source, target, options, err)
+		if !readOnly && (strings.Contains(err.Error(), "can't read superblock") || strings.Contains(err.Error(), "structure needs cleaning")) {
+			// Trying to auto correct the filesystem by running fsck
+			_, repairErr := detectAndRepairFilesystem(source, []string{"-E", "journal_only", "-a"}, m)
```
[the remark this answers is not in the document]
mittachaitu: running fsck on xfs formatted device is no-op

pkg/azuredisk/azure_common_linux.go
```diff
@@ -291,6 +391,45 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
...
+	if err != nil {
+		ee, isExitError := err.(utilexec.ExitError)
+		switch {
+		case err == utilexec.ErrExecutableNotFound:
+			klog.Warningf("'fsck' not found on system; continuing without running 'fsck'.")
+			// Treating it as filesystem existence check, since we cannot determine if the filesystem is valid without fsck
+			return true, nil
+		case isExitError && ee.ExitStatus() == fsckOperationalError:
+			klog.Infof("Device %s may be fresh block device with no filesystem signature, fsck output: %s", source, string(out))
+			return false, nil
```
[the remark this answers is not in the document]
mittachaitu: Since we reach here only if blkid returns empty response.. so we should be good considering it as `fresh block device`

pkg/azuredisk/azure_common_linux_test.go
```diff
@@ -17,16 +20,224 @@ limitations under the License.
...
+	"k8s.io/utils/exec"
+	testingexec "k8s.io/utils/exec/testing"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
+	testmounter "sigs.k8s.io/azuredisk-csi-driver/pkg/mounter"
 )
 
-func TestRescanAllVolumes(t *testing.T) {
-	if runtime.GOOS == "darwin" {
-		t.Skipf("skip test on GOOS=%s", runtime.GOOS)
+func TestFormatAndMountFormatsUnformattedDisk(t *testing.T) {
```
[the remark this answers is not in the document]
mittachaitu: Not required

pkg/azuredisk/azure_common_linux.go
```diff
@@ -124,7 +133,98 @@ func formatAndMount(source, target, fstype string, options []string, m *mount.Sa
...
+	}
+
+	if diskFSFormat == "" {
+		// As part of auto-recovery from mount failures, we run fsck on the disk. If the disk was pulled
+		// out or a power loss occurred during a previous fsck run, the primary superblocks may have been
+		// corrupted depending on the stage at which fsck was interrupted. We run fsck here to detect and
+		// repair such issues before formatting.
+		// Note: fsck is a no-op if the disk already has an xfs signature.
+		// If filesystem was detected at this stage don't format return the error
+		isFilesystemExist, repairErr := detectAndRepairFilesystem(source, []string{"-n"}, m)
```
NedAnd1: Can we add a test which expects `detectAndRepairFilesystem` on a filesystem wiped via `wipefs` to return `false`, `nil`?

pkg/azuredisk/azure_common_linux.go
```diff
@@ -291,6 +391,45 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
...
+	klog.V(2).Infof("Checking for issues with fsck on disk: %s with options %v", source, fsckOptions)
+	args := append(append([]string(nil), fsckOptions...), source)
+	out, err := mounter.Exec.Command("fsck", args...).CombinedOutput()
+	if err != nil {
+		ee, isExitError := err.(utilexec.ExitError)
+		switch {
+		case err == utilexec.ErrExecutableNotFound:
+			klog.Warningf("'fsck' not found on system; continuing without running 'fsck'.")
+			// Treating it as filesystem existence check, since we cannot determine if the filesystem is valid without fsck
+			return true, nil
```
pawanpraka1: unlikely case but we should fail here

mittachaitu: Since it returns filesystem existence, if the disk is not formated then we will return error in subsequent steps i.e during Mount call

pkg/azuredisk/azure_common_linux.go
```diff
@@ -291,6 +391,45 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
...
+		case isExitError && ee.ExitStatus() == fsckOperationalError:
+			klog.Infof("Device %s may be fresh block device with no filesystem signature, fsck output: %s", source, string(out))
+			return false, nil
+		case isExitError && ee.ExitStatus() == fsckErrorsCorrected:
+			klog.Infof("Device %s has errors which were corrected by fsck: %s", source, string(out))
+		case isExitError && ee.ExitStatus() == fsckErrorsUncorrected:
+			// Filesystem exists but fsck found errors that it could not correct
+			return true, fmt.Errorf("'fsck' found errors on device %s with options %v but could not correct them: %s", source, fsckOptions, string(out))
+		case isExitError && ee.ExitStatus() > fsckErrorsUncorrected:
+			klog.Infof("`fsck` error %s", string(out))
```
pawanpraka1: we should return error here.

pkg/azuredisk/azure_common_linux.go
mittachaitu: ideally returning error if `fsck` binary doesn't exist doesn't looks good, since we are returning true we will be re-reading filesystem at that time we will fail.

```sh
// Re-read the filesystem signature to determine the existing format.
			reReadFormat, err := m.GetDiskFormat(source)
			if err != nil {
				klog.Errorf("formatAndMount - failed to re-read filesystem signature on disk %s with error: %v", source, err)
				return fmt.Errorf("failed to re-read filesystem signature on disk %s with error: %v", source, err)
			}
````

Comment (pawanpraka1) [reply pkg/azuredisk/azure_common_linux.go]: function detectAndRepairFilesystem should do what it is supposed to do so that it can be reused at other places also. If the function can not detect, it should fail.

Comment (mittachaitu) [reply pkg/azuredisk/azure_common_linux.go]: Since we are shipping [fsck along](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/36a6a3f023736306329cf387d1a2ddc1414e59be/pkg/azurediskplugin/Dockerfile#L17) with the container we should be good mostly will not hit it... I will return error sake of completeness

File: pkg/azuredisk/azure_common_linux.go (hunk 14 above)
Comment (mittachaitu) [reply pkg/azuredisk/azure_common_linux.go]: Updated the code

File: pkg/azuredisk/azure_common_linux.go (hunk 13 above)
Comment (mittachaitu) [reply pkg/azuredisk/azure_common_linux.go]: Updated the code

File: pkg/azuredisk/azure_common_linux.go [hunk 15]
```diff
@@ -291,6 +391,50 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
...
+// to attempt repairs.
+func detectAndRepairFilesystem(source string, fsckOptions []string, mounter *mount.SafeFormatAndMount) (bool, error) {
+	klog.V(2).Infof("Checking for issues with fsck on disk: %s with options %v", source, fsckOptions)
+	args := append(append([]string(nil), fsckOptions...), source)
+	out, err := mounter.Exec.Command("fsck", args...).CombinedOutput()
+	if err != nil {
+		ee, isExitError := err.(utilexec.ExitError)
+		switch {
+		case err == utilexec.ErrExecutableNotFound:
+			klog.Warningf("'fsck' not found on system; continuing without running 'fsck'.")
```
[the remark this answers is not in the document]
mittachaitu: Done

pkg/azuredisk/azure_common_linux.go
```diff
@@ -291,6 +391,76 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
...
+	args := []string{"--no-act", "--output", "TYPE", source}
+	fsType, err := mounter.Exec.Command("wipefs", args...).CombinedOutput()
+	if err != nil {
+		klog.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'wipefs' with error(%v)", source, err)
+		return false, fmt.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'wipefs' with error(%v)", source, err)
+	}
+	if len(fsType) > 0 {
+		return true, nil
+	}
+	return false, nil
```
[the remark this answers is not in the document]
mittachaitu: Updated the command that we execute

pkg/azuredisk/azure_common_linux.go
```diff
@@ -291,6 +390,76 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
...
+//
+// fsckOptions control fsck's behavior, e.g. "-n" for a read-only detection check or "-y"/"-E ..."
+// to attempt repairs.
+func detectAndRepairFilesystem(source string, fsckOptions []string, mounter *mount.SafeFormatAndMount) (bool, error) {
+	klog.V(2).Infof("Checking for issues with fsck on disk: %s with options %v", source, fsckOptions)
+	args := append(append([]string(nil), fsckOptions...), source)
+	out, err := mounter.Exec.Command("fsck", args...).CombinedOutput()
+	if err != nil {
+		ee, isExitError := err.(utilexec.ExitError)
+		switch {
```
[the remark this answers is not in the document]
mittachaitu: Chances for permission denied is not possible in CSI driver since it runs in privileged mode.

pkg/azuredisk/azure_common_linux_test.go
```diff
@@ -17,16 +20,266 @@ limitations under the License.
...
 import (
-	"runtime"
+	"reflect"
 	"testing"
 
+	"k8s.io/utils/exec"
+	testingexec "k8s.io/utils/exec/testing"
 	"sigs.k8s.io/azuredisk-csi-driver/pkg/azureutils"
+	testmounter "sigs.k8s.io/azuredisk-csi-driver/pkg/mounter"
 )
```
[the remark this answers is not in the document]
mittachaitu: Permission denied is not possible since driver runs in privileged permissions.

pkg/azuredisk/azure_common_linux.go
```diff
@@ -124,7 +133,97 @@ func formatAndMount(source, target, fstype string, options []string, m *mount.Sa
...
+				}
+			} else if fstype == "xfs" {
+				args = []string{
+					source,
+				}
+			}
+
+			klog.Infof("Disk %q appears to be unformatted, attempting to format as type: %q with options: %v", source, fstype, args)
+			output, err := m.Exec.Command("mkfs."+fstype, args...).CombinedOutput()
+			if err != nil {
```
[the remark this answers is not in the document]
mittachaitu: Updated the code

pkg/azuredisk/azuredisk.go
```diff
@@ -206,6 +210,10 @@ func NewDriver(options *DriverOptions) *Driver {
 	driver.neverStopTaintRemoval = options.NeverStopTaintRemoval
 	driver.maxConcurrentFormat = options.MaxConcurrentFormat
 	driver.concurrentFormatTimeout = options.ConcurrentFormatTimeout
+	if options.MaxConcurrentFormat > 0 {
+		driver.formatSem = make(chan any, options.MaxConcurrentFormat)
+		driver.formatTimeout = time.Duration(options.ConcurrentFormatTimeout) * time.Second
+	}
```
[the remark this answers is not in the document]
mittachaitu: Updated the code

pkg/azuredisk/azure_common_linux.go
```diff
@@ -119,12 +129,132 @@ func findDiskByLun(lun int, io azureutils.IOHandler, _ *mount.SafeFormatAndMount
...
+				klog.Errorf("formatAndMount - failed to re-read filesystem signature on disk %s with error: %v", source, err)
+				return fmt.Errorf("failed to re-read filesystem signature on disk %s with error: %v", source, err)
+			}
+			klog.Infof("Disk %q contains filesystem signature post running fsck, skipping format; detected filesystem type: %q", source, reReadFormat)
+			diskFSFormat = reReadFormat
+		}
+	}
+
+	if !strings.EqualFold(fstype, diskFSFormat) {
```
pawanpraka1, line 207: nice change.

pkg/azuredisk/azure_common_linux.go
```diff
@@ -291,6 +421,76 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
...
+			klog.Warningf("Device %s has errors which were corrected by fsck: %s", source, string(out))
+		case isExitError && ee.ExitStatus() == fsckErrorsUncorrected:
+			// Filesystem exists but fsck found errors that it could not correct
+			klog.Errorf("Device %s has errors which fsck could not correct with options %v: %s", source, fsckOptions, string(out))
+			return true, fmt.Errorf("'fsck' found errors on device %s with options %v but could not correct them (exit status %d)", source, fsckOptions, ee.ExitStatus())
+		case isExitError && ee.ExitStatus() > fsckErrorsUncorrected:
+			klog.Errorf("`fsck` error %s", string(out))
+			return false, fmt.Errorf("'fsck' failed on device %s with options %v: %v", source, fsckOptions, err)
+		default:
+			klog.Warningf("fsck on device %s failed with error %v, output: %v", source, err, string(out))
```
pawanpraka1, line 488: in what cases we will come here? Do we return error?

pkg/azuredisk/azure_common_linux.go
```diff
@@ -291,6 +421,76 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
...
+		return true, nil
+	}
+
+	args := []string{"--no-act", "--output", "TYPE", "--noheadings", source}
+	fsType, err := mounter.Exec.Command("wipefs", args...).CombinedOutput()
+	if err != nil {
+		klog.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'wipefs' with error(%v)", source, err)
+		return false, fmt.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'wipefs' with error(%v)", source, err)
+	}
+	if len(fsType) > 0 {
```
pawanpraka1: use strings.TrimSpace(string(fsType)). it may contain white space.

pkg/azuredisk/azure_common_linux.go
```diff
@@ -119,12 +129,132 @@ func findDiskByLun(lun int, io azureutils.IOHandler, _ *mount.SafeFormatAndMount
...
+			if readOnly {
+				return fmt.Errorf("formatAndMount - disk %s is not formatted, but mount options specify read-only", source)
+			}
+
+			// If filesystem doesn't exist then we can format the disk with the given fstype.
+			// Disk is unformatted
+			args := []string{source}
+			if fstype == "ext4" || fstype == "ext3" {
+				args = []string{
+					"-m0", // Zero blocks reserved for super-user
```
pawanpraka1, line 177: add -F option also.

pkg/azuredisk/azure_common_linux.go
mittachaitu: When fsck returns following exit codes we will log and proceed...
```sh

       2
           System should be rebooted
      16
           Usage or syntax error

       32
           Checking canceled by user request

       128
           Shared-library error
```
> Do we return error?

No, we don't return error, but returns existence of filesystem. So, when caller re-reads information and finds no filesystem then we are returning same error to the NodeStageVolume request.

pkg/azuredisk/azure_common_linux.go
mittachaitu: If it is fresh disk, why we need to add `-F` force option?

mittachaitu: Updated

pawanpraka1 (approved):
looks good.

pkg/azuredisk/azure_common_linux.go
```diff
@@ -119,12 +129,134 @@ func findDiskByLun(lun int, io azureutils.IOHandler, _ *mount.SafeFormatAndMount
...
+	if !readOnly {
+		_, err := detectAndRepairFilesystem(source, []string{"-a"}, m)
+		if err != nil {
+			klog.Errorf("formatAndMount - failed to run fsck on disk %s with error: %v", source, err)
+			return fmt.Errorf("failed to run fsck on disk %s with error: %v", source, err)
+		}
+	}
+
+	klog.V(4).Infof("Attempting to mount disk %s in %s format at %s", source, fstype, target)
+	options = append(options, "defaults")
```
landreasyan, line 223: Can this silently turn a read-only mount into read-write?
`defaults` expands to include `rw` and mount option precedence is last-wins. Can we drop this append? Is this needed?

pkg/azuredisk/azure_common_linux.go
```diff
@@ -291,6 +423,76 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
...
+	if err != nil {
+		ee, isExitError := err.(utilexec.ExitError)
+		switch {
+		case err == utilexec.ErrExecutableNotFound:
+			klog.Errorf("'fsck' not found on system; cannot verify filesystem signature on %s, returning error.", source)
+			return false, fmt.Errorf("'fsck' not found to detect filesystem on device %s with options %v: %v", source, fsckOptions, err)
+		case isExitError && ee.ExitStatus() == fsckOperationalError:
+			if strings.Contains(strings.ToLower(string(out)), "superblock could not be read or does not describe a valid ext2/ext3/ext4") {
+				klog.Infof("Device %s may be fresh block device with no filesystem signature, fsck output: %s", source, string(out))
+				return false, nil
```
landreasyan: Should we return an error in this case as well?

mittachaitu: If we return error format will not be executed by caller... When we run fsck on unformatted disk then above error is expected and as per function name this error is signs of not having filesystem so we are nullifying the error (We can treat same as NotFound error).

landreasyan:
This change adds several decision branches (blkid → fsck -n → wipefs → format vs. recover vs. already-formatted), and the riskiest one with reformatting after detection reports the device as empty. Today we'd only be able to reconstruct which path ran from logs after the fact; however, since these operations now happen in our code and we already expose `azuredisk_csi_driver_operations_total{operation, success}` metrics, I think we can get good observability here without adding any new metric families by emitting a few new operation values. 

Copilot complied this list:
- format_and_mount.directmount (success t/f): directmount skip path
- format_and_mount.already_formatted (success true): blkid returned a type
- format_and_mount.recovered (success true): blkid empty, but fsck/wipefs found a fs (device saved from reformat)
- format_and_mount.formatted (success t/f): mkfs ran (false = mkfs failed)
- format_and_mount.format_mismatch (success false): detected fs ≠ requested fstype
- reformat_after_empty_detection (success true): blkid + fsck -n + wipefs all empty → about to mkfs (the data-loss path)
- disk_format (success t/f): mkfs invocation result
- fsck_detect.<outcome> (success t/f): -n detection fsck; outcome ∈ {clean, fresh_device, errors_corrected, errors_uncorrected, operational_error, not_found, fatal_error}
- fsck_repair.<outcome> (success t/f): -a repair fsck, same outcomes

These two would probably be the most interesting: 
- operations_total{operation="reformat_after_empty_detection"} > 0: the potential data-loss path is being exercised (worth alerting on).
- operations_total{operation="format_and_mount.recovered"} climbing: blkid is under-reporting on non-blank disks and the new detection fallback is doing real work.

/suggestion can be added later in a subsequent PR.

pkg/azuredisk/azure_common_linux.go
landreasyan: This sounds very error prone. Maybe we can omit an error that the call can explicitly ignore?

pkg/azuredisk/azure_common_linux.go
NedAnd1: I don't think we should ignore unknown errors as that would be problematic if the error was context cancellation as one example, and if we ignore exit code 2 `System should be rebooted`, it really does mean that the system i.e. kernel should be rebooted &/or volume moved to another node before the file system should be mounted,
but seeing as this is in the upstream k8s code, we should probably take this up with them while deferring a change to this in a follow-up PR.

pkg/azuredisk/azure_common_linux.go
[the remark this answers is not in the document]
ganesan23: @mittachaitu just get rid of the "This will not fully fix the issue, but it is an attempt to avoid running fsck" line. Make the description say "an attempt to avoid re-formatting an already-formatted disk". That would make it clearer

pkg/azuredisk/azure_common_linux.go
```diff
@@ -291,6 +423,76 @@ func rescanAllVolumes(io azureutils.IOHandler) error {
...
+	if err != nil {
+		klog.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'fsck' with error(%v)", source, err)
+		return false, fmt.Errorf("detectFilesystemExistence - failed to check existence of filesystem on disk %s through 'fsck' with error(%v)", source, err)
+	}
+	if isFSExist {
+		return true, nil
+	}
+
+	args := []string{"--no-act", "--output", "TYPE", "--noheadings", source}
+	fsType, err := mounter.Exec.Command("wipefs", args...).CombinedOutput()
```
ganesan23, line 443: please make sure that wipefs is available in the container image of node driver, on ubuntu/al3/acl3 on amd64/arm64 etc., otherwise we could end up with a hard error

pkg/azuredisk/azure_common_linux.go
mittachaitu: I tried to retain the mount-utils [behavior](https://github.com/kubernetes/kubernetes/blob/42850664fd99edca42ca8bbfd20ecfb197fdbdc7/staging/src/k8s.io/mount-utils/mount_linux.go#L571), so it shouldn't turn `RO` into `RW`

pkg/azuredisk/azure_common_linux.go
mittachaitu: With this PR we are expecting users to wipe the disk fully with zeroes before it is used by azure-disk-csi-driver, if users erase primary superblock and retain in backup superblocks then we need user intervention to clean the filesystem manually

mittachaitu: From CSI driver level we can't differentiate cases of corrupted disk Vs primary sb earsed disk, so expectation is azure will provide clean disk.

pkg/azuredisk/azure_common_linux.go
mittachaitu: We are installing [utils-linux](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/497bdb95f37dea77ce12de675c798991cf616f1a/pkg/azurediskplugin/Dockerfile#L17) so we should be good here

PR #3712, opened 2026-07-10
fix: increase default VMSS detach timeout from 20s to 200s

### LINKED ISSUE #3692: v1.34: aggressive vmss-detach-timeout (20s default) + force detach fallback breaks VMSS rolling upgrade (cloud-init not mounted on ~5% of reimaged nodes)
**What happened**:

A customer reported an unexpected disk detachment / VM inconsistency issue after upgrading the Azure Disk CSI Driver Helm chart from **v1.33.4 → v1.34.3**. The upgrade had to be rolled back.

Symptoms:
- During **VMSS rolling upgrades**, when a VM is rebooted / reimaged as part of the upgrade, **cloud-init (Azure custom data) is sometimes not mounted correctly** when the instance starts with the new image.
- Affected nodes fail to initialize, causing the VMSS rolling upgrade to fail.
- The issue is **random and only affects ~5% of worker nodes** — always nodes that had Azure Disk PVs attached. Control-plane nodes (no Azure Disk PVs) are never affected.
- Downgrading the driver back to v1.33.4 makes the problem go away.

Azure Support has independently correlated the failure to **force detach operations** that happened during the upgrade window:

> "The issue seems to happen each time the disks are forcefully detached. Instance 5, which worked correctly, did not experience any force detach operations."
>
> "The failure was triggered by disks being forcefully detached from the VM by the Kubernetes/Rancher orchestration layer during the upgrade window. This left the VM in an inconsistent state that is not supported for upgrade operations, leading to the failure."

**Suspected root cause**:

The behavior change strongly matches the new detach path introduced in v1.34.0, in particular [#3462](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3462) (`perf: Optimize VMSS disk detach with early timeout and polling strategy`), which introduced the `--vmss-detach-timeout-seconds` flag with a **default of 20 seconds**. Relevant code in `pkg/azuredisk/azure_controller_common.go`:

```go
vmssDetachTimeout := time.Duration(c.VMSSDetachTimeoutInSeconds) * time.Second // default 20s
earlyExitCtx, earlyCancel := context.WithTimeout(detachCtx, vmssDetachTimeout)

if err = vmset.DetachDisk(earlyExitCtx, nodeName, diskMap, false); err != nil {
    ...
    if errors.Is(err, context.DeadlineExceeded) {
        err = c.pollForDetachCompletion(detachCtx, diskName, diskURI, nodeName, vmset)
    }
    if err != nil && c.ForceDetachBackoff {
        err = vmset.DetachDisk(ctx, nodeName, diskMap, true) // force detach
    }
}
```

Combined with:
- The reduction of the external-attacher `-timeout` from `1200s` to `600s` (same PR).
- The dangling-detach changes (`waitForDiskManagedByTobeRemoved`, [#3469](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3469)) that already caused one confirmed regression in [#3524](https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/3524).

Hypothesized timeline for an affected node:

1. VMSS rolling upgrade begins → a VM enters reimage.
2. kubelet initiates pod termination → CSI `ControllerUnpublishVolume` is called for its Azure Disk PVs.
3. Because the VM is already in a transitional state, the Azure control plane responds slowly to the detach request (normal, but slow).
4. On **v1.33.4**, the driver would wait a long time (hundreds of seconds), and the graceful detach would usually succeed.
5. On **v1.34.3**, the 20s `vmssDetachTimeout` fires → driver switches to polling → then falls back to `DetachDisk(..., true)` (force detach).
6. That force detach lands on a VM that is already reimaging / booting into the new image → VM ends up in the "inconsistent state" Azure Support described → cloud-init / Azure custom data disk fails to mount on boot → node never becomes Ready → VMSS rolling upgrade fails.

This is consistent with everything the customer / Azure Support observed:
- Only PV-holding worker nodes are affected (control plane never enters this path).
- Behavior is timing-dependent, so only a fraction of nodes trip the race.
- Reverting to v1.33.4 removes the aggressive early-timeout + force detach fallback and hides the issue.

The default detach behavior in v1.34.x should be at least as safe as v1.33.x for VMSS rolling-upgrade scenarios. Specifically:
- Force detach should not be triggered against a VM that is in the middle of a reimage / upgrade operation just because the graceful detach did not confirm within 20 seconds.
- The default `--vmss-detach-timeout-seconds=20` looks too aggressive when the CSI driver is running against VMSSes that are being upgraded (where Azure control-plane responses to per-VM operations can legitimately be slow for minutes).

**How to reproduce it**:

Reliable reproduction is time-sensitive, but the pattern is:
1. AKS-like environment (customer uses Rancher on VMSS) with a workload using multiple Azure Disk PVs.
2. Upgrade the CSI driver Helm chart to v1.34.x (>= v1.34.0), keep default `--vmss-detach-timeout-seconds=20` and default `--force-detach-backoff=true`.
3. Trigger a VMSS rolling upgrade (new image / reimage) on the worker VMSS.
4. Observe that a fraction of VMs come back with cloud-init / Azure custom data mount failures, correlating in Azure Activity Log with a "detach data disk" operation flagged as force detach immediately before the reimage completes.

Two workarounds that both restore the pre-1.34 behavior (I've suggested the customer test both to confirm the diagnosis before we commit to a code-level fix):

1. Raise `--vmss-detach-timeout-seconds` to match the pre-1.34 semantics, e.g.
   ```yaml
   controller:
     VMSSDetachTimeoutInSeconds: 240
   ```
2. Disable the force detach fallback:
   ```yaml
   controller:
     extraArgs:
       - "--force-detach-backoff=false"
   ```

If either of these makes the issue disappear on v1.34.3, that confirms the detach-path change is the source and we can discuss a proper fix (e.g. a longer default for `vmssDetachTimeout`, or gating force detach when the target VM is in a known transitional state like `Reimaging` / `Upgrading` / `Deallocating`).

Related:
- [#3462](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3462) — perf: Optimize VMSS disk detach with early timeout and polling strategy (v1.34.0)
- [#3469](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3469) — Handling dangling detaches in a better way
- [#3524](https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/3524) — earlier confirmed regression in the same code path
- [#3294](https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/3294) — later fix around the detach context / active-context-for-force-detach behavior

**Environment**:

- CSI Driver version: **v1.34.3** (regressed vs **v1.33.4**)
- Kubernetes version: (Rancher-managed, AKS-like) — will update once confirmed by customer
- Cloud provider or hardware configuration: Azure VMSS worker nodepool with Azure Disk PVs
- OS: reported on customer's standard VMSS image (cloud-init / Azure custom data based)
- Kernel: N/A
- Install tools: Helm chart from this repo
- Others: force detach was confirmed by Azure Support to be present on every failing VM and absent on the ones that upgraded cleanly.

andyzhangx:
## Problem

The current `defaultVMSSDetachTimeoutInSeconds` is set to **20 seconds**, which is the initial timeout for waiting on the VMSS DetachDisk API result (`earlyExitCtx`). When the operation does not return within this window, the driver:

1. Stops waiting for the operation result and enters **GET-based polling** (`pollForDetachCompletion`) to check if the disk was removed from the VM
2. If polling also fails within the remaining `DetachOperationMinTimeout` (240s) budget, triggers a **force detach**

The 20s default is too short — most Azure disk detach operations complete in 30-120s. With only 20s, the driver almost always exits early, enters the slower polling path, and frequently triggers force detach as a routine fallback rather than a last resort.

### Observed impact in CI

In [PR #3709 Windows 2022 HostProcess E2E](https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_azuredisk-csi-driver/3709/pull-azuredisk-csi-driver-e2e-capz-windows-2022-hostprocess/2075214246420942848/build-log.txt):

**All 22 tests PASSED**, but the job was marked FAILURE due to Prow 3h timeout (`EXIT_VALUE=130`). A single test case (`reattach-disk-multiple-nodes`) took **5318 seconds (88 minutes)** instead of ~2 minutes, pushing total wall time over 3h.

### Root cause analysis

The 88-minute test was caused by repeated detach timeout → force detach cycles:

1. `pre-provisioned-readOnly` detach from `jsb4d`: earlyExit at 20s → polling ~16s → force detach (~36s total)
2. Force detach left `jsb4d` VM in `Failed` provisioning state  
3. ~85 minutes later, `reattach-disk-multiple-nodes` test needed to attach a shared disk to `jsb4d`
4. The disk was still attached to `6n89v` → driver tried detach from `6n89v` → earlyExit at 20s → polling ~42s → force detach again (~62s total)
5. Each detach cycle took ~60s due to the 20s earlyExit + polling + force detach path. Multiple retries with kubelet backoff accumulated to 88 minutes.

Note: the `vmState Failed` itself did **not** cause additional blocking — Azure recovered the VM during subsequent attach operations within ~3 minutes. The real overhead came entirely from the repeated slow detach path caused by the 20s timeout being too short.

### Timeline from the CI log

```
14:36:25 attach pre-provisioned-readOnly to jsb4d (vmState Succeeded)
14:40:47 detach pre-provisioned-readOnly from jsb4d started
14:41:07 earlyExit after 20s → polling started
14:41:50 polling timed out → force detach triggered
14:41:56 force detach completed (jsb4d → vmState Failed)

(85 min gap — no disk operations on jsb4d, other tests used different nodes)

16:06:58 attach reattach-disk to jsb4d (vmState Failed)
16:06:59 disk still on 6n89v → detach from 6n89v started
16:07:19 earlyExit after 20s → polling started
16:08:01 polling timed out → force detach from 6n89v
16:08:07 retry attach to jsb4d (still Failed)
16:10:59 attach pvc-c91d4791 to jsb4d (vmState Succeeded ← recovered)
16:10:54 reattach test finally completed — took 5318s total
```

## Fix

Increase `defaultVMSSDetachTimeoutInSeconds` from **20** to **200**.

### Why 200s?

The detach path has a two-phase timeout design:

```
DetachDisk
├─ Phase 1: earlyExitCtx (VMSSDetachTimeout)
│   Wait for VMSS DetachDisk API operation result
│   ├─ Success → done ✅
│   └─ Timeout → enter Phase 2
│
├─ Phase 2: remaining time (DetachOperationMinTimeout - VMSSDetachTimeout)
│   GET VM polling to confirm disk removal
│   ├─ Disk gone → done ✅  (critical for zone outage scenarios)
│   └─ Timeout → force detach ⚠️
│
└─ Force detach (last resort, may cause VM Failed state)
```

With `VMSSDetachTimeout = 200s` and `DetachOperationMinTimeout = 240s`:
- **Phase 1 (200s)**: gives the VMSS API sufficient time to complete. Covers the vast majority of detach operations (30-120s) plus slower cases under Azure platform throttling or zone issues (3-5 min).
- **Phase 2 (remaining ~40s)**: GET VM polling to confirm disk removal. Critical during zone outages where disks may be detached outside the CSI driver's control.
- **Force detach**: only triggered after both phases are exhausted — a true last resort.

With the old 20s timeout, the same detach that would take 30-60s via the normal VMSS API path would instead take 60s+ via the earlyExit → polling → force detach slow path. Increasing to 200s lets normal operations complete directly without entering the fallback chain.

### What this does NOT change

- `DetachOperationMinTimeoutInSeconds` (240s default) — unchanged
- The force detach fallback mechanism — still exists, just triggered much less frequently
- Any existing `--vmss-detach-timeout-in-seconds` flag overrides — still honored
- Normal detach operations — still complete as fast as they can (timeout is just an upper bound)

Fixes #3692

```release-note
fix: increase default VMSS detach timeout from 20s to 200s to reduce unnecessary force detach operations
```


andyzhangx:
Addressed Copilot review comments:

1. **CLI flag default** — updated `--vmss-detach-timeout-seconds` default from 20 to 200 in `pkg/azuredisk/azuredisk_option.go` and the reset log message in `pkg/azuredisk/azuredisk.go` (commit 32e7b7569).
2. **Versioned charts** — verified that only `charts/latest` contains `VMSSDetachTimeoutInSeconds`; no other versioned charts (v1.1.0 through v1.34.x) reference this field, so no additional updates needed.

PR #3731, opened 2026-07-24
feat: add OpenTelemetry tracing to the Azure Disk CSI driver

srranjan-sri:
**What type of PR is this?**
/kind feature

This adds tracing so we can see exactly how long each CSI operation takes and where the time actually goes, without needing to run a separate tracing collector.

/kind feature

Every CSI gRPC call (CreateVolume, ControllerPublishVolume, NodeStageVolume, etc.) becomes a timed span, so we can measure end-to-end latency per operation.

Each operation is broken into sub-spans for the expensive steps underneath it:

Azure ARM calls, e.g. CreateManagedDisk, DeleteManagedDisk, ResizeDisk, ModifyDisk, AttachDisk, DetachDisk, UpdateVM, GetDiskLun, checkDiskExists
Kubernetes API calls: GetNode, ListVolumeAttachments
Node-side work, e.g. waitForDevice, formatAndMount, resizeVolume, bindMount, unmount, rescanDevice, resizeFilesystem
This makes it obvious whether Azure, the Kubernetes API, or the mount step is the slow part.

Traces are correlatable: root spans are labeled with disk URI, disk name, node name and volume ID, and child spans link back to their parent, so all the spans of one workflow group together.

Throttling and back-pressure are now visible:

ARM throttling (HTTP 429) is recorded as a "throttled" event with the real Retry-After delay.
Client-side rate-limiter waits (QPS/Burst back-pressure) are recorded as a "kube_client_ratelimited" event with the wait time. This lives in a new helper (ratelimiter_trace.go) that talks to the OpenTelemetry API directly to avoid an import cycle.
Spans are written to the normal container logs (via klog at verbosity 2 by default, overridable with the OTEL_KLOG_SPAN_VERBOSITY env var), so no OTLP collector is required.

Tracing is off by default and gated behind --enable-otel-tracing. When disabled, a no-op tracer is used, so there is effectively zero overhead.

High-frequency health-check RPCs (Probe, GetPluginInfo/Capabilities, NodeGetCapabilities, NodeGetVolumeStats, ControllerGetCapabilities) are filtered out so they don't flood the logs.

Also adds unit tests verifying spans appear in logs, that nothing is emitted when tracing is disabled, and that the Retry-After value is parsed correctly from a throttling error.

image

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azuredisk/controllerserver.go
```diff
@@ -493,6 +499,11 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 		// operation as ProvisioningInBackground and retry; CSI idempotency on
 		// the retry returns the existing disk's URI.
 		if azureutils.IsThrottlingError(err) {
+			retryAfter := ""
```
hccheng72: We can reuse recordThrottleIfThrottled(ctx, err). I think re-check IsThrottlingError(err) is cheap, but this removes the copy-pasted parsing logic so there's a single source of truth for how a throttle event is built. It also benefits DRY/maintainability.

pkg/azureutils/azure_disk_utils.go
```diff
@@ -271,6 +272,19 @@ func GetKubeClient(kubeconfig string, qps float64, burst int) (clientset.Interfa
 		return nil, err
 	}
 
+	// Wrap the client-side rate limiter so that requests blocked by QPS/Burst
```
hccheng72: GetKubeClient is only used by the prestop hook (main.go:79) and a unit test. The kube_client_ratelimited events will essentially never fire on the real driver path. The one that actually runs the traced ListVolumeAttachments and GetNode calls — is built separately at azuredisk.go:256

pkg/azuredisk/azuredisk.go
hccheng72: nit: update it to tracerProvider based on your change. Also apply to line 445 and 448

pkg/azuredisk/trace_test.go
```diff
@@ -0,0 +1,182 @@
+/*
+Copyright 2023 The Kubernetes Authors.
```
hccheng72: nit: should we put 2026 here for the new files?

mittachaitu:
@srranjan-ms , Can you help in adding output how does this looks for volume creation/volume attachments either in PR description or comment to this?

srranjan-sri:
> @srranjan-ms , Can you help in adding output how does this looks for volume creation/volume attachments either in PR description or comment to this?

Updated the description with the logs screenshot.

pkg/azuredisk/controllerserver.go
```diff
@@ -736,15 +777,22 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
...
+		occSpan.End()
 		klog.V(2).Infof("Trying to attach volume %s to node %s", diskName, nodeName)
 
 		attachDiskInitialDelay := azureutils.GetAttachDiskInitialDelay(volumeContext)
 		if attachDiskInitialDelay > 0 {
 			klog.V(2).Infof("attachDiskInitialDelayInMs is set to %d", attachDiskInitialDelay)
 			d.diskController.AttachDetachInitialDelayInMs = attachDiskInitialDelay
 		}
-		lun, err = d.diskController.AttachDisk(ctx, diskName, diskURI, nodeName, cachingMode, disk, occupiedLuns)
+		attachSpanCtx, attachSpan := startSpan(ctx, "AttachDisk")
```
andyzhangx: what about moving this logic to `AttachDisk` func, and write like following, that would make the logic quite clear, and same for other places, e.g. if you want to record trace of function A, then record inside function A instead of outside.

```
func (c *controllerCommon) AttachDisk(...) (lun int32, err error) {
    ctx, attachSpan := startSpan(ctx, "AttachDisk",
        attribute.String(attrDiskURI, diskURI),
        attribute.String(attrNode, string(nodeName)))
    defer func() {
        recordSpanResult(attachSpan, err)
        recordThrottleIfThrottled(ctx, err)
        attachSpan.End()
    }()
    ...
}
```

andyzhangx: and use `defer func() ` , that would be more graceful

andyzhangx:
Suggestion: could we move the `startSpan("AttachDisk")` / `recordSpanResult` / `recordThrottleIfThrottled` / `attachSpan.End()` boilerplate **into `diskController.AttachDisk` itself**, instead of wrapping it at every call site?

Right now each caller does:

```go
attachSpanCtx, attachSpan := startSpan(ctx, "AttachDisk")
lun, err = d.diskController.AttachDisk(attachSpanCtx, diskName, diskURI, nodeName, cachingMode, disk, occupiedLuns)
recordSpanResult(attachSpan, err)
recordThrottleIfThrottled(attachSpanCtx, err)
attachSpan.End()
```

If we move that into `controllerCommon.AttachDisk`, callers just do:

```go
lun, err = d.diskController.AttachDisk(ctx, diskName, diskURI, nodeName, cachingMode, disk, occupiedLuns)
```

and `AttachDisk` owns its own span end-to-end, e.g.:

```go
func (c *controllerCommon) AttachDisk(ctx context.Context, diskName, diskURI string, nodeName types.NodeName,
    cachingMode armcompute.CachingTypes, disk *armcompute.Disk, occupiedLuns []int) (lun int32, err error) {
    ctx, span := startSpan(ctx, "AttachDisk",
        attribute.String(attrDiskURI, diskURI),
        attribute.String(attrNode, string(nodeName)))
    defer func() {
        recordSpanResult(span, err)
        recordThrottleIfThrottled(ctx, err)
        span.End()
    }()
    ...
}
```

Benefits:

- Trace shape stays the same and matches responsibility boundaries: `ControllerPublishVolume -> AttachDisk -> setDiskLun / vmUpdate / verifyAttach`.
- No duplicated span boilerplate on first attach vs. dangling-detach retry path (and any future caller).
- Throttle/result recording can't accidentally be forgotten on a new call site.

Same idea applies to `DetachDisk`, `UpdateVM`, and other disk-controller entry points if we want tracing style to be consistent.

srranjan-sri:
> Suggestion: could we move the `startSpan("AttachDisk")` / `recordSpanResult` / `recordThrottleIfThrottled` / `attachSpan.End()` boilerplate **into `diskController.AttachDisk` itself**, instead of wrapping it at every call site?

Addressed the suggestion. I moved the startSpan / recordSpanResult / recordThrottleIfThrottled  / span.End() boilerplate inside AttachDisk so it owns its span via named returns + defer. Callers now just call AttachDisk(ctx, ...) with no wrapping, and both the first-attach and dangling-detach retry paths are covered automatically.
also applied the same pattern to the other disk-controller entry points for consistency: DetachDisk, UpdateVM, GetDiskLun, CreateManagedDisk, DeleteManagedDisk, ResizeDisk, and ModifyDisk.
The trace shape is unchanged, and I verified the full lifecycle on a live cluster, also attached the output in description.

pkg/azuredisk/trace.go
```diff
@@ -19,41 +19,280 @@ package azuredisk
...
+	}
+
+	opts := []trace.TracerProviderOption{
+		trace.WithResource(res),
+		// Sample based on the parent's decision, falling back to always
+		// sampling for root spans, so an entire trace is kept or dropped
+		// together.
+		trace.WithSampler(trace.ParentBased(trace.AlwaysSample())),
+		// Always write spans to the container logs.
+		trace.WithBatcher(&klogSpanExporter{}),
```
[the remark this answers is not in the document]
andyzhangx: Addressed in 2cd1e4a — went with **Option A**: keep OTLP alongside the klog exporter and let env decide.

- `InitOtelTracing()` now attaches the klog span exporter and, when `OTEL_EXPORTER_OTLP_ENDPOINT` (or `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`) is set, an OTLP/gRPC exporter as well. No endpoint = OTLP is skipped silently (no connection-error spam).
- New `OTEL_TRACES_EXPORTER=klog|otlp|klog,otlp|none` env override for explicit control.
- `--enable-otel-tracing` help text updated to describe both exporters, `OTEL_KLOG_SPAN_VERBOSITY`, and `OTEL_TRACES_EXPORTER`.
- Helm chart `otelExporterEndpoint` now defaults to `""` and the template only injects `OTEL_EXPORTER_OTLP_ENDPOINT` when it is non-empty; `OTEL_SERVICE_NAME` still emitted whenever tracing is enabled.
- Unit tests for `parseTracesExporters` and `hasOTLPEndpoint` added.

Net effect: existing deployments that pointed the driver at a collector keep receiving OTLP traces after upgrade; new deployments get klog spans out of the box, and users can opt into either exporter alone.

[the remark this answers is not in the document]
andyzhangx: this is a blocking issue since it's changing original behavior

PR #3756, opened 2026-08-10
fix: optionally skip reading the config from the API server

### LINKED ISSUE #3755: The driver reads configuration from a Secret when AZURE_CREDENTIAL_FILE is set
**What happened**:
When AZURE_CREDENTIAL_FILE is set and its content has everything the driver needs, the driver still tries to read its credentials from the Kubernetes API server Secret.

The driver uses the env. var and does not talk to the API server at all, at least at the driver startup.

(This issue is the same as it used to be for azure-file csi driver: https://github.com/kubernetes-sigs/azurefile-csi-driver/issues/2939)

mpatlasov:
**What type of PR is this?**
/kind bug

Skip reading the driver configuration from the API server when -cloud-config-secret-name or -cloud-config-secret-namespace is explicitly set to an empty string.

The unit test change is technically not necessary - no unit test actually tries to get the Secret from provided namespace / name, it only make the unit test future proof once they start doing so.

Fixes #3755

This fixes the aforementioned bug in a backward compatible way. If anyone set AZURE_CREDENTIAL_FILE, the driver by default still reads the config from the Kubernetes API server Secret. Only when someone explicitly sets -cloud-config-secret-name to "", the driver uses AZURE_CREDENTIAL_FILE env var without talking to the API server.
I think this makes the Deployments in this repository working as they are today.

(The text above and fix implementation was borrowed from https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/2940)

**Requirements**:
- [x] includes documentation
- [ ] adds unit tests

**Release note**:
```
The driver no longer tries to read its config from the API server Secret if `-cloud-config-secret-name` or `-cloud-config-secret-namespace` are empty strings. `AZURE_CREDENTIAL_FILE` must be set and point to a valid config file then.
```


mpatlasov:
/assign @andyzhangx 
Hi Andy, we have many customers suffering from [this issue](https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/3755), the fix is the same as [PR#2940](https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/2940) in azure-file. Please take a look.

andyzhangx (approved):
/retest
/lgtm

PR #3761, opened 2026-08-12
[hold] fix: Sync before unmounting as part of NodeUnstageVolume request

mittachaitu:
**What type of PR is this?**
/kind bug

This PR does the following changes in NodeUnstageVolume:
1. Perform Sync on global mount before performing Unmount
2. Wait for filesystem shutdown before responding to NodeUnstageVolume request.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx:
Thanks for tackling this — the direction is right (Kubelet's single `NodeUnstageVolume` → controller detach means we really do need to make sure dirty pages hit the disk and the FS is fully torn down before we return). A few things I'd like to see addressed before this merges.

### 🔴 Blocker: sync failure now makes `NodeUnstageVolume` un-recoverable

```go
if err := syncFilesystemAtMountPoint(stagingTargetPath); err != nil {
    klog.Errorf("...")
    return nil, status.Errorf(codes.Internal, "Failed to sync filesystem at mount point %q: %v", ...)
}
```

Old behavior: `CleanupMountPoint` always ran, so an unhealthy volume could still be unstaged and Azure could still detach it.

New behavior: any failure from `syncFilesystemAtMountPoint` (EIO on a bad disk, a weird permission state on the mount point, transient errors from `os.Open`, etc.) causes `NodeUnstageVolume` to fail with `Internal`. Kubelet will keep retrying with the same error, the volume stays in `NodeStaged`, and every Pod using it is stuck `Terminating` forever. `ControllerUnpublish`/detach never gets a chance to run.

`sync` is best-effort by nature — the `waitFSShutdown` path below already treats its failures as `klog.Warning` for exactly this reason. `syncFilesystemAtMountPoint` should do the same:

```go
if err := syncFilesystemAtMountPoint(stagingTargetPath); err != nil {
    klog.Errorf("NodeUnstageVolume - sync failed at %q, proceeding with unmount: %v", stagingTargetPath, err)
    // fall through: better to unmount a partially-synced FS than to hang forever
}
```

Somewhat related: consider opening with `O_PATH|O_DIRECTORY` instead of `os.Open` for the sync fd — `O_PATH` skips permission checks and is the canonical “just give me an fd on this dentry” pattern. Removes one class of avoidable `EACCES` failures.

### 🟡 Comment on the sync step is misleading

```go
// unmount will be a no-op if the mount is referenced elsewhere, so we need to sync
// the filesystem at the mount point before unmounting it.
```

`umount(2)` never becomes a no-op because of bind mounts — it unbinds *this* mount instance. What's actually happening is that other references delay superblock teardown, so dirty pages could still be around when we return to Kubelet. Suggest rewording along the lines of “explicitly sync so we don't rely on FS shutdown for durability; other bind references may keep the superblock alive past unmount”.

### 🟡 `waitFSShutdown` diagnostics depend on `hostPID`

```go
// Iterate through all processes in /proc to find mount references for the given device, hostPID should set to true
entries, err := os.ReadDir("/proc")
```

If the DaemonSet isn't running under `hostPID: true` (older charts, hardened deployments), we only see the CSI plugin's own `/proc/1`, and the diagnostic log line becomes actively misleading (“no other references found”). Two options:

1. Assert `hostPID: true` in the Helm chart and document it as a requirement for this diagnostic to work.
2. Detect it at runtime (e.g., check whether `/proc/1/comm` is `systemd`/`kubelet`, or count entries) and log a clear “diagnostic skipped: not running in host PID namespace” warning instead of returning empty results.

### 🟡 5s timeout is hardcoded

```go
timeout := 5 * time.Second
```

For premium SSDs with hundreds of MB of dirty cache under load, 5s is thin. It only produces warnings today, but noisy warnings from a real production path are expensive to debug. Consider making it a driver-level config with a 5s default, or reusing an existing driver timeout knob.

### 🟢 XFS path relies on `/sys/fs/xfs/<dev>` disappearance

`journalEntryGlob` correctly returns empty for XFS (no jbd2, XFS has an internal log), and the `/sys/fs/xfs/<dev>` entry disappearing on `put_super` is a reasonable “journal drained” proxy today. Worth a comment noting the dependency in case a kernel refactor moves the sysfs entry lifecycle around.

### 🟢 Test coverage

- `TestSyncFilesystemAtMountPoint` covers happy path + missing dir but not the sync-failure path — which is exactly the branch we need to be confident about after the fix above.
- `waitFSShutdown`, `findDeviceMountReferences`, and `unescapeMountInfo` have no direct tests. `unescapeMountInfo` is trivially table-testable and worth adding. For `findDeviceMountReferences`, consider parameterizing the `/proc` root so a fake proc tree can drive the whole scan under test.

### 🟢 Minor

- In `nodeserver.go`, calling `d.mounter.GetDiskFormat(devicePath)` *after* `CleanupMountPoint` is fine (it uses `blkid` on the block device, not the mount), but if the controller side raced the detach in the meantime, `GetDiskFormat` will error and we skip `waitFSShutdown` — which is actually the correct fallback, just noting for future readers that this ordering is intentional.

Once the sync-error path is downgraded to a warning, I'm happy with the overall design.

pkg/azuredisk/azure_common_linux.go
```diff
@@ -566,3 +567,214 @@ func (d *Driver) GetVolumeStats(_ context.Context, m *mount.SafeFormatAndMount,
 		},
 	}, nil
 }
+
+// Performs sync operation on the filesystem mounted at the given mount point
+func syncFilesystemAtMountPoint(mountPoint string) error {
+	pathExists, pathErr := mount.PathExists(mountPoint)
+	shouldSync, err := shouldSyncFilesystem(pathExists, pathErr)
```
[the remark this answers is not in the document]
mittachaitu: Good catch!!

pkg/azuredisk/nodeserver.go
```diff
@@ -215,12 +215,36 @@ func (d *Driver) NodeUnstageVolume(_ context.Context, req *csi.NodeUnstageVolume
...
 	}
 	klog.V(2).Infof("NodeUnstageVolume: unmount %s successfully", stagingTargetPath)
 
+	if devicePath != "" {
+		fsType, err := d.mounter.GetDiskFormat(devicePath)
+		if err != nil {
+			klog.Warningf("NodeUnstageVolume - failed to get filesystem type for device %q: %v", devicePath, err)
+		} else {
+			klog.V(2).Infof("NodeUnstageVolume: waiting for filesystem to shutdown staging path %s device %s", stagingTargetPath, devicePath)
+			waitFSShutdown(devicePath, fsType)
```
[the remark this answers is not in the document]
mittachaitu: Intentionally added warning log instead of returning error because in Kubernetes cluster other pods might be holding reference if they start/re-start post csi driver perform NodeStageVolume request. So returning error will never allow detach stage to succeed.

mittachaitu:
> Thanks for tackling this — the direction is right (Kubelet's single `NodeUnstageVolume` → controller detach means we really do need to make sure dirty pages hit the disk and the FS is fully torn down before we return). A few things I'd like to see addressed before this merges.
> 
> ### 🔴 Blocker: sync failure now makes `NodeUnstageVolume` un-recoverable
> ```go
> if err := syncFilesystemAtMountPoint(stagingTargetPath); err != nil {
>     klog.Errorf("...")
>     return nil, status.Errorf(codes.Internal, "Failed to sync filesystem at mount point %q: %v", ...)
> }
> ```
> 
> Old behavior: `CleanupMountPoint` always ran, so an unhealthy volume could still be unstaged and Azure could still detach it.
> 
> New behavior: any failure from `syncFilesystemAtMountPoint` (EIO on a bad disk, a weird permission state on the mount point, transient errors from `os.Open`, etc.) causes `NodeUnstageVolume` to fail with `Internal`. Kubelet will keep retrying with the same error, the volume stays in `NodeStaged`, and every Pod using it is stuck `Terminating` forever. `ControllerUnpublish`/detach never gets a chance to run.
> 
> `sync` is best-effort by nature — the `waitFSShutdown` path below already treats its failures as `klog.Warning` for exactly this reason. `syncFilesystemAtMountPoint` should do the same:
> 
> ```go
> if err := syncFilesystemAtMountPoint(stagingTargetPath); err != nil {
>     klog.Errorf("NodeUnstageVolume - sync failed at %q, proceeding with unmount: %v", stagingTargetPath, err)
>     // fall through: better to unmount a partially-synced FS than to hang forever
> }
> ```
> 
> Somewhat related: consider opening with `O_PATH|O_DIRECTORY` instead of `os.Open` for the sync fd — `O_PATH` skips permission checks and is the canonical “just give me an fd on this dentry” pattern. Removes one class of avoidable `EACCES` failures.
> 
> ### 🟡 Comment on the sync step is misleading
> ```go
> // unmount will be a no-op if the mount is referenced elsewhere, so we need to sync
> // the filesystem at the mount point before unmounting it.
> ```
> 
> `umount(2)` never becomes a no-op because of bind mounts — it unbinds _this_ mount instance. What's actually happening is that other references delay superblock teardown, so dirty pages could still be around when we return to Kubelet. Suggest rewording along the lines of “explicitly sync so we don't rely on FS shutdown for durability; other bind references may keep the superblock alive past unmount”.
> 
> ### 🟡 `waitFSShutdown` diagnostics depend on `hostPID`
> ```go
> // Iterate through all processes in /proc to find mount references for the given device, hostPID should set to true
> entries, err := os.ReadDir("/proc")
> ```
> 
> If the DaemonSet isn't running under `hostPID: true` (older charts, hardened deployments), we only see the CSI plugin's own `/proc/1`, and the diagnostic log line becomes actively misleading (“no other references found”). Two options:
> 
> 1. Assert `hostPID: true` in the Helm chart and document it as a requirement for this diagnostic to work.
> 2. Detect it at runtime (e.g., check whether `/proc/1/comm` is `systemd`/`kubelet`, or count entries) and log a clear “diagnostic skipped: not running in host PID namespace” warning instead of returning empty results.
> 
> ### 🟡 5s timeout is hardcoded
> ```go
> timeout := 5 * time.Second
> ```
> 
> For premium SSDs with hundreds of MB of dirty cache under load, 5s is thin. It only produces warnings today, but noisy warnings from a real production path are expensive to debug. Consider making it a driver-level config with a 5s default, or reusing an existing driver timeout knob.
> 
> ### 🟢 XFS path relies on `/sys/fs/xfs/<dev>` disappearance
> `journalEntryGlob` correctly returns empty for XFS (no jbd2, XFS has an internal log), and the `/sys/fs/xfs/<dev>` entry disappearing on `put_super` is a reasonable “journal drained” proxy today. Worth a comment noting the dependency in case a kernel refactor moves the sysfs entry lifecycle around.
> 
> ### 🟢 Test coverage
> * `TestSyncFilesystemAtMountPoint` covers happy path + missing dir but not the sync-failure path — which is exactly the branch we need to be confident about after the fix above.
> * `waitFSShutdown`, `findDeviceMountReferences`, and `unescapeMountInfo` have no direct tests. `unescapeMountInfo` is trivially table-testable and worth adding. For `findDeviceMountReferences`, consider parameterizing the `/proc` root so a fake proc tree can drive the whole scan under test.
> 
> ### 🟢 Minor
> * In `nodeserver.go`, calling `d.mounter.GetDiskFormat(devicePath)` _after_ `CleanupMountPoint` is fine (it uses `blkid` on the block device, not the mount), but if the controller side raced the detach in the meantime, `GetDiskFormat` will error and we skip `waitFSShutdown` — which is actually the correct fallback, just noting for future readers that this ordering is intentional.
> 
> Once the sync-error path is downgraded to a warning, I'm happy with the overall design.

Thanks @andyzhangx for valuable suggestions, I made sync-error path as logging instead of failing. On the suggestion of opening mount directory with `O_PATH|O_DIRECTORY` is not good for syncfs(fd) because it uses fdget(fd) and return `EBADF`. If open(..., O_PATH) creates an `FMODE_PATH` descriptor, so  `fdget()` treats it as empty and syncfs() returns `EBADF`

pkg/azuredisk/nodeserver.go
```diff
@@ -215,12 +215,36 @@ func (d *Driver) NodeUnstageVolume(_ context.Context, req *csi.NodeUnstageVolume
...
 
+	// Read the device path from the mount point before unmounting it
+	devicePath, err := getDevicePathWithMountPath(stagingTargetPath, d.mounter)
+	if err != nil {
+		klog.Warningf("NodeUnstageVolume - failed to get device path for mount point %q: %v", stagingTargetPath, err)
+	}
+
+	// To ensure data integrity, we should perform sync explicitly before unmounting the volume.
+	// unmount will be a no-op if the mount is referenced elsewhere, so we need to sync the filesystem at the mount point before unmounting it.
+	if err := syncFilesystemAtMountPoint(stagingTargetPath); err != nil {
```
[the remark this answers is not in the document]
mittachaitu: Updated

andyzhangx (commented):
/ok-to-test

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node.yaml
```diff
@@ -35,6 +35,8 @@ spec:
 {{ toYaml . | indent 8 }}
 {{- end }}
     spec:
+      # Required for mount-reference diagnostics across pod mount namespaces.
+      hostPID: true
```
pawanpraka1: this could be security concern, need to check with security team. Better to disable is by default.

pkg/azuredisk/azure_common_linux.go
```diff
@@ -566,3 +574,309 @@ func (d *Driver) GetVolumeStats(_ context.Context, m *mount.SafeFormatAndMount,
...
+	}
+	if pathErr != nil {
+		return fmt.Errorf("error checking mount point path %q before sync: %w", mountPoint, pathErr)
+	}
+	if !pathExists {
+		klog.Warningf("Skipping filesystem sync because mount point does not exist: %q", mountPoint)
+		return nil
+	}
+
+	isMountPoint, mountErr := mounter.IsMountPoint(mountPoint)
```
pawanpraka1: move this inside shouldSyncFilesystem.

pkg/azuredisk/azure_common_linux.go
```diff
@@ -566,3 +574,309 @@ func (d *Driver) GetVolumeStats(_ context.Context, m *mount.SafeFormatAndMount,
...
+	}
+
+	mountDir, err := openMountPoint(mountPoint, unix.O_RDONLY|unix.O_DIRECTORY|unix.O_CLOEXEC, 0)
+	if err != nil {
+		return err
+	}
+	defer unix.Close(mountDir)
+
+	klog.V(2).Infof("Syncing filesystem at mount point %q", mountPoint)
+	if err := syncfs(mountDir); err != nil {
```
pawanpraka1: syncfs is does not guarranty data persistence, you should use fsync.

pkg/azuredisk/nodeserver.go
```diff
@@ -215,12 +215,29 @@ func (d *Driver) NodeUnstageVolume(_ context.Context, req *csi.NodeUnstageVolume
...
+	devicePath, err := getDevicePathWithMountPath(stagingTargetPath, d.mounter)
+	if err != nil {
+		klog.Warningf("NodeUnstageVolume - failed to get device path for mount point %q: %v", stagingTargetPath, err)
+	}
+
+	// Explicitly sync so we don't rely on FS shutdown for durability; other mount references may keep the superblock alive past unmount
+	if err := syncFilesystemAtMountPoint(stagingTargetPath, d.mounter); err != nil {
+		klog.Errorf("NodeUnstageVolume - sync failed at %q, proceeding with unmount: %v", stagingTargetPath, err)
+		// fall through: better to unmount a partially-synced FS than to hang forever
+	}
```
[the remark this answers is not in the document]
mittachaitu: To retain the earlier behavior when underlying disk is in bad state we are logging and continuing with dettach process.

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node.yaml
mittachaitu: We need to set `hostPID` to `true` to understand who are holding reference to mount point

mittachaitu: Will ask security team for their input

andyzhangx:
unmount on linux would already do the sync_filesystem operation, what's the benefit of this PR? and also this PR adds `hostPID: true` which opens a security hole in this CSI driver.

mittachaitu:
> unmount on linux would already do the sync_filesystem operation, what's the benefit of this PR?

There is need for performing unmount due to Several security / monitoring / debug pods bind-mount the host root path (`/`) with mountPropagation: None (kernel rprivate). Because `/` contains the CSI staging tree `/var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/…,` these agents privately mount the very subtree of the PVs are staged under.

**How This Can Lead to Filesystem Inconsistencies**:
-  A private (rprivate) bind of `/` captures a private copy of whatever is mounted at bind time. Arming condition — either: (a) the agent starts after a application pod on that node (PV already staged when the agent binds /), or (b) the agent is restarted — while a PV is mounted. Either way the agent's namespace holds a private reference to the PV mount.
-  **The unmount SUCCEEDS** - There is no EBUSY. [`inferred, source-verified: linux fs/namespace.c, fs/pnode.c, fs/super.c`] do_umount's busy test is scoped per struct mount, not per super_block. The non-detach path returns `-EBUSY` only if `propagate_mount_busy(mnt, 2)` is true — i.e. `!list_empty(&mnt->mnt_mounts)` (something stacked on this mount) or `do_refcount_check(mnt, 2)` (an fd/cwd held through this mount pushes mnt_count past the 2 expected refs). A private clone in another container is a separate **struct mount** (`clone_mnt`, own parent, own `mnt_count`): not in this mount's `mnt_mounts`, and its open fds bump the _clone's_ count, not this one. **`Neither condition trips → propagate_mount_busy returns 0 → umount_tree runs → umount(2) returns 0`**.
- **Filesystem is NOT torn down**. `clone_mnt` did `atomic_inc(&sb->s_active)`, and ->`put_super` fires only when the last `s_active` ref drops. The surviving clone still holds one, so `ext4_put_super / jbd2_journal_destroy / EXT4_VALID_FS` stamping **never run**. The ext4 sb + journal stay live and rw behind the hidden clone. (A plain `MNT_DETACH` lazy umount behaves the same — never EBUSY, sb stays alive if a clone remains).
- **Silent false-success → detach of a live filesystem**. The orchestrator sees a clean unmount, believes the volume is released, and proceeds to detach the device while ext4 is still mounted rw with an active journal in the other namespace → torn journal → dirty ext4 filesystem.

So to avoid inconsistencies we are performing explicit sync instead of depending on `linux sync_filesystem` as part of unmount.

pkg/azuredisk/azure_common_linux.go
mittachaitu: Updated code to use syncfs

andyzhangx:
> > unmount on linux would already do the sync_filesystem operation, what's the benefit of this PR?
> 
> There is need for performing unmount due to Several security / monitoring / debug pods bind-mount the host root path (`/`) with mountPropagation: None (kernel rprivate). Because `/` contains the CSI staging tree `/var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/…,` these agents privately mount the very subtree of the PVs are staged under.
> 
> **How This Can Lead to Filesystem Inconsistencies**:
> 
> * A private (rprivate) bind of `/` captures a private copy of whatever is mounted at bind time. Arming condition — either: (a) the agent starts after a application pod on that node (PV already staged when the agent binds /), or (b) the agent is restarted — while a PV is mounted. Either way the agent's namespace holds a private reference to the PV mount.
> * **The unmount SUCCEEDS** - There is no EBUSY. [`inferred, source-verified: linux fs/namespace.c, fs/pnode.c, fs/super.c`] do_umount's busy test is scoped per struct mount, not per super_block. The non-detach path returns `-EBUSY` only if `propagate_mount_busy(mnt, 2)` is true — i.e. `!list_empty(&mnt->mnt_mounts)` (something stacked on this mount) or `do_refcount_check(mnt, 2)` (an fd/cwd held through this mount pushes mnt_count past the 2 expected refs). A private clone in another container is a separate **struct mount** (`clone_mnt`, own parent, own `mnt_count`): not in this mount's `mnt_mounts`, and its open fds bump the _clone's_ count, not this one. **`Neither condition trips → propagate_mount_busy returns 0 → umount_tree runs → umount(2) returns 0`**.
> * **Filesystem is NOT torn down**. `clone_mnt` did `atomic_inc(&sb->s_active)`, and ->`put_super` fires only when the last `s_active` ref drops. The surviving clone still holds one, so `ext4_put_super / jbd2_journal_destroy / EXT4_VALID_FS` stamping **never run**. The ext4 sb + journal stay live and rw behind the hidden clone. (A plain `MNT_DETACH` lazy umount behaves the same — never EBUSY, sb stays alive if a clone remains).
> * **Silent false-success → detach of a live filesystem**. The orchestrator sees a clean unmount, believes the volume is released, and proceeds to detach the device while ext4 is still mounted rw with an active journal in the other namespace → torn journal → dirty ext4 filesystem.
> 
> So to avoid inconsistencies we are performing explicit sync instead of depending on `linux sync_filesystem` as part of unmount.

@mittachaitu but the bind-mount only happens in `NodePublishVolume`

andyzhangx:
Trying to think through whether this PR actually protects against data loss in the **azuredisk** topology specifically — I'm having a hard time constructing a scenario where it does.

**azuredisk stage/publish layout:**
- `NodeStageVolume`: real fs mount of `/dev/sdX` at `globalmount`
- `NodePublishVolume`: `mount --bind globalmount → pod targetPath`
- `NodeUnpublishVolume`: `umount targetPath` (tears down the bind)
- `NodeUnstageVolume`: `umount globalmount` (tears down the real fs mount)

At the point `NodeUnstageVolume` runs, the bind clone is already gone, so `s_active` on the ext4/xfs superblock is 1. `umount(2)` on the real mount is synchronous and takes the path:

```
umount → deactivate_super → generic_shutdown_super → sync_filesystem
```

`sync_filesystem` blocks until every dirty page is written back before `umount(2)` returns. Both ext4 and xfs implement `->put_super` correctly here — there's no path that skips the sync. So calling `sync` (or `syncfs`) right before `umount` on `globalmount` is doing the same work the kernel is about to do unconditionally a few instructions later.

The `propagate_mount_busy` / `clone_mnt` / `s_active > 0` argument in the description would apply to a driver that lets `NodeUnstage` run while a bind clone is still alive somewhere. In the default azuredisk deployment I don't see that: no `hostPID`, `Bidirectional` propagation on the driver container, no subPath-specific clones held by the CSI driver itself (those are kubelet-side). If `s_active` is 1 by the time `umount(globalmount)` runs, the kernel path above always fires.

I tried to enumerate cases where this PR would save data on azuredisk:

| scenario | does the PR help? |
|---|---|
| app doesn't fsync, kubelet does normal unpublish + unstage | no — kernel syncs in `umount` |
| app crashes / OOMKilled | no — kubelet still runs normal unpublish + unstage |
| node graceful reboot | no — systemd `umount` runs the same kernel path |
| node hard power loss | no — CSI code doesn't get to run |
| Azure force-detach of the disk | no — detach already happened |
| journal corruption | no — that's an fsck problem |

And detach ordering also protects us: kubelet only reports `NodeUnstageVolume` success upward after the syscall returns, and `ControllerUnpublishVolume` (the actual ARM detach) only fires after that. So the disk is quiescent-on-medium before the device disappears.

Concerns on the cost side of shipping this anyway:

1. **`sync(2)` is process-wide across the whole node.** One volume teardown will flush every dirty page from every other pod on the node. That's a real p99 regression on `NodeUnstageVolume` and can stall unrelated workloads on busy nodes. If we do end up needing something, `syncfs(fd)` scoped to the mountpoint is the right primitive.
2. **`hostPID: true`** is a nontrivial security concession for the daemonset. What in the sync/wait path actually needs to see host PIDs?
3. **New failure modes on the unmount path** (sync EIO, timeout on `waitForShutdown`) that we don't have today, in exchange for behavior the kernel already guarantees.

If there's a concrete repro I'm missing — an azuredisk deployment + workload where you can observe data on the disk lagging behind writes acked to the app after `NodeUnstageVolume` returns success — I'd love to see it, because that would change my read here. Otherwise this looks like an argument imported from an in-place bind-mount driver topology (where `NodePublish` *is* the fs mount and unmount race is possible) that doesn't map to azuredisk's two-hop stage/publish model.

/cc @andyzhangx

mittachaitu:
Hi @andyzhangx,
    The Azure Disk CSI driver correctly cleans up the volume within its own mount namespace and the host namespace. However, monitoring or debugging pods that mount the host root filesystem with  `mountPropagation: None`  ( `rprivate` ) can inherit private references to Azure Disk mounts if they start after `NodeStageVolume` completes.
    During `NodeUnstageVolume` , the unmount propagates between the CSI driver and host because `/var/lib/kubelet` uses bidirectional propagation, but it does not propagate into those private namespaces. Their remaining references can keep the filesystem’s superblock active (`s_active > 1`), preventing final filesystem shutdown. Normal writeback and explicit sync operations can still flush dirty data, but the driver cannot rely on the final-unmount flush until all references are released.


**Note**: 
1. _We couldn't see these mount references in other pods post `NodeUnstageVolume` request because we are performing `rm <staging_path>` post unmounting. Deleting the staging_path only makes that reference unreachable by path name; It doesn't release mount. The filesystem shutdown only when all mount references, including private namespace copies are removed or their namespace terminate._
2. Information from `/proc/*/mountinfo` also can't prove a lingering mount because linux calls `__detach_mounts()`, which lazily unmounts all mounts attached to that dentry across namespaces.
```sh
vfs_rmdir()
  -> detach_mounts()
     -> __detach_mounts()
        -> umount_tree(..., UMOUNT_CONNECTED)
```
The lazy unmount will results in delay in flushing dirty pages which may leads to filesystem check errors.

**Scenario**: I have the following proof to say that **`s_active > 0`**  even after performing NodeUnsgateVolume request.
- Create application pod which mount the disk through azuredisk-csi driver(/any other csi-driver which mounts block device).
- Once application pod comes into running state deploy another pod which mounts `/` root of hostfilesystem (let's call it as observer pod).
```yaml
  containers:
    - name: ubuntu
      image: ubuntu:24.04
      command:
        - sleep
        - infinity
      volumeMounts:
        - name: host-root
          mountPath: /host
  volumes:
    - name: host-root
      hostPath:
        path: /
        type: Directory
```
- Exec inside the observer pod and perform following things to hold the reference
```sh
# Store the global mount in TARGET variable
TARGET=/host/var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount
# Create some-file
touch $TARGET/some-file

# hold access
exec 9<"$TARGET/some-file"
cd "$TARGET"
```
- At this point from different terminal delete the application pod (I intentionally errored out ControllerUnpublishVolume by making changes in the code because I want disk to be attached to the node post unmount).  azuredisk-csi driver performs successful `NodeUnstageVolume` call (Can look from below logs).
- Go back to terminal holding reference and execute following command, since the node driver removes the path stat on $TARGET will return error.
```sh
root@pod2-none:/host/var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount# stat $TARGET
stat: cannot statx '/host/var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount': No such file or directory

## Even after unmount you could still access the content
globalmount# cat <&9
root@pod2-none:/host/var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount# ls /proc/$$/cwd
c.txt  lost+found  some-file
root@pod2-none:/host/var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount# stat -Lc 'device=%d inode=%i' /proc/$$/fd/9
device=66310 inode=13
```
- So above proves unmount may not flush the dirty pages if mount reference is held privately by different pods mounting entire host with `None` mount propagation.







**Logs of CSI node driver**:

```sh
I0819 13:02:08.473625       1 utils.go:105] GRPC call: /csi.v1.Node/NodeUnstageVolume
I0819 13:02:08.473640       1 utils.go:106] GRPC request: {"staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount","volume_id":"/subscriptions/d64ddb0c-7399-4529-a2b6-037b33265372/resourceGroups/MC_mittas-sa_mittas-containerd_centralindia/providers/Microsoft.Compute/disks/pvc-18f2e1ed-99f1-449e-be85-605d7a3ff343"}
I0819 13:02:08.475860       1 azure_common_linux.go:613] Syncing filesystem at mount point "/var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount"
I0819 13:02:08.476222       1 azure_common_linux.go:617] Finished syncing filesystem at mount point "/var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount"
I0819 13:02:08.476233       1 nodeserver.go:231] NodeUnstageVolume: unmounting /var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount
I0819 13:02:08.476241       1 mount_helper_common.go:93] unmounting "/var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount" (corruptedMount: false, mounterCanSkipMountPointChecks: true)
I0819 13:02:08.476248       1 mount_linux.go:402] Unmounting /var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount
I0819 13:02:08.477468       1 mount_helper_common.go:150] Deleting path "/var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount"
I0819 13:02:08.477937       1 nodeserver.go:235] NodeUnstageVolume: unmount /var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount successfully
I0819 13:02:08.477946       1 mount_linux.go:670] Attempting to determine if disk "/dev/nvme0n3" is formatted using blkid with args: ([-p -s TYPE -s PTTYPE -o export /dev/nvme0n3])
I0819 13:02:08.479510       1 mount_linux.go:673] Output: "DEVNAME=/dev/nvme0n3\nTYPE=ext4\n"
I0819 13:02:08.479523       1 azure_common_linux.go:638] NodeUnstageVolume: waiting for filesystem to shutdown staging path /var/lib/kubelet/plugins/kubernetes.io/csi/disk.csi.azure.com/ca1038634c9497fa8dbe9522c58234528e9aecd35288e3d4b328637dcdfcc514/globalmount device /dev/nvme0n3
W0819 13:02:13.480389       1 azure_common_linux.go:663] filesystem sysfs entry "/sys/fs/ext4/nvme0n3" was not removed: timed out waiting for "/sys/fs/ext4/nvme0n3" to be removed
W0819 13:02:13.480463       1 azure_common_linux.go:679] journal entry "/proc/fs/jbd2/nvme0n3-8" was not removed: timed out waiting for "/proc/fs/jbd2/nvme0n3-8" to be removed
W0819 13:02:13.480475       1 azure_common_linux.go:691] Filesystem shutdown checks failed for device "/dev/nvme0n3": filesystem sysfs entry "/sys/fs/ext4/nvme0n3" was not removed: timed out waiting for "/sys/fs/ext4/nvme0n3" to be removed; journal entry "/proc/fs/jbd2/nvme0n3-8" was not removed: timed out waiting for "/proc/fs/jbd2/nvme0n3-8" to be removed
I0819 13:02:13.480973       1 metrics.go:165] "Observed Request Latency" logger="logLatency" latency_seconds=5.007247189 request="azuredisk_csi_driver_node_unstage_volume" success="true" resource_group="mc_mittas-sa_mittas-containerd_centralindia" source="disk.csi.azure.com" volumeid="/subscriptions/d64ddb0c-7399-4529-a2b6-037b33265372/resourceGroups/MC_mittas-sa_mittas-containerd_centralindia/providers/Microsoft.Compute/disks/pvc-18f2e1ed-99f1-449e-be85-605d7a3ff343"
I0819 13:02:13.480987       1 utils.go:112] GRPC response: {}
```

mittachaitu:
> sync(2) is process-wide across the whole node. One volume teardown will flush every dirty page from every other pod on the node. That's a real p99 regression on NodeUnstageVolume and can stall unrelated workloads on busy nodes. If we do end up needing something, syncfs(fd) scoped to the mountpoint is the right primitive.

I have updated the code to perform `fsync` on opened `staging_path` fd instead of sync, so it should flush only particular volume data which is undergoing NodeUnstageVolume request, so this issue shouldn't be major issue.

> hostPID: true is a nontrivial security concession for the daemonset. What in the sync/wait path actually needs to see host PIDs?

I have added this change only to capture what are the other process holding reference but while testing I observed that this will not be no longer needed. As soon as disk csi driver perform `rm <staging_path>`  it removes entries under `/proc/<pid>/mountinfo`

> New failure modes on the unmount path (sync EIO, timeout on waitForShutdown) that we don't have today, in exchange for behavior the kernel already guarantees.

I believe with previous comment should be able to answer this question.

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-node.yaml
mittachaitu: I don't find need of having `hostPID`, unconfigured in last commit

pkg/azuredisk/nodeserver.go
```diff
@@ -215,12 +215,29 @@ func (d *Driver) NodeUnstageVolume(_ context.Context, req *csi.NodeUnstageVolume
...
 	defer d.volumeLocks.Release(volumeID)
 
+	// Read the device path from the mount point before unmounting it
+	devicePath, err := getDevicePathWithMountPath(stagingTargetPath, d.mounter)
+	if err != nil {
+		klog.Warningf("NodeUnstageVolume - failed to get device path for mount point %q: %v", stagingTargetPath, err)
+	}
+
+	// Explicitly sync so we don't rely on FS shutdown for durability; other mount references may keep the superblock alive past unmount
+	if err := syncFilesystemAtMountPoint(stagingTargetPath, d.mounter); err != nil {
```
pawanpraka1: we don't need to do it for raw block volume.

pkg/azurediskplugin/Dockerfile
```diff
@@ -12,9 +12,26 @@
...
 # limitations under the License.
 
+FROM registry.k8s.io/build-image/debian-base:bookworm-v1.0.8 AS e2fsprogs-builder
+
+ARG E2FSPROGS_VERSION=1.47.4
+ARG E2FSPROGS_SHA256=fd5bf388cbdbe006a3d3b318d983b2948382440acc85a87f1e7d108653e8db0b
+
+RUN clean-install build-essential ca-certificates curl xz-utils
+WORKDIR /tmp/e2fsprogs
+RUN curl -fsSLo e2fsprogs.tar.xz \
```
pawanpraka1: why we need to do this? What is the issue with existing way of installing.

pkg/azuredisk/azure_common_linux.go
```diff
@@ -566,3 +574,151 @@ func (d *Driver) GetVolumeStats(_ context.Context, m *mount.SafeFormatAndMount,
...
+
+	shouldSync, err := shouldSyncFilesystem(mountPoint, mounter)
+	if err != nil {
+		return fmt.Errorf("error checking whether %q is a mount point before sync: %w", mountPoint, err)
+	}
+	if !shouldSync {
+		return nil
+	}
+
+	mountDir, err := openMountPoint(mountPoint, unix.O_RDONLY|unix.O_DIRECTORY|unix.O_CLOEXEC, 0)
```
pawanpraka1, line 601: this will fail for raw block.

pkg/azurediskplugin/Dockerfile
mittachaitu: There is bug in existing version of e2fsck, I made this change for local testing will revert it

pkg/azuredisk/nodeserver.go
mittachaitu: This is taken care by implementation of `syncFilesystemAtMountPoint`, i.e before opening FD on staging path it checks for `shouldSyncFilesystem` which checks for staging path mount point and only perform fsync so we should be good.

pkg/azuredisk/azure_common_linux.go
mittachaitu: We are doing open only if `mountPoint` is mounted.. for rawblock volume it will get skip

mittachaitu:
**Note**: We will hold this PR since it can cause delays in pod movement and leads to application down time.

mittachaitu:
**Currently PR is hold because of following reasons**:
1. **Scope** — The CSI driver's responsibility is limited to orchestrating the volume lifecycle; it makes no guarantee of data consistency prior to unmount. If an application requires crash consistency, it must handle that itself.
2. **Latency / availability** — Issuing a `sync` before unmounting delays the disk detach and move, directly extending application downtime.
3. **Kernel already handles it** — As long as the cluster admin ensures no other pod in the cluster mounts the root directory with `mountPropagation` set to `None/default`, the unmount performed by the CSI driver is sufficient on its own: the kernel flushes the page cache automatically when the last mount of a filesystem is unmounted.

PR #3767, opened 2026-08-15
feat: Initial changes to support QAD workflow

nearora-msft:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azuredisk/controllerserver.go
```diff
@@ -569,6 +773,11 @@ func (d *Driver) ControllerModifyVolume(ctx context.Context, req *csi.Controller
 	}
 
 	diskURI := volumeID
+
+	if _, isQAD, err := d.isUsingQADPath(ctx, diskURI); err == nil && isQAD {
```
hccheng72, line 777: should we handle the case when it returns error here? Or if it has error, we take it as non-qad case? Same question for ControllerExpandVolume.

pkg/azuredisk/nodeserver.go
```diff
@@ -221,6 +431,61 @@ func (d *Driver) NodeUnstageVolume(_ context.Context, req *csi.NodeUnstageVolume
 	}
 	klog.V(2).Infof("NodeUnstageVolume: unmount %s successfully", stagingTargetPath)
 
+	if pv, isUsingQAD, err := d.isUsingQADPath(ctx, volumeID); isUsingQAD && err == nil {
```
hccheng72, line 434: Q: what will happen if err != nil but it's using QAD, and we don't call wireserver to detach?

pkg/azuredisk/nodeserver.go
```diff
@@ -221,6 +431,61 @@ func (d *Driver) NodeUnstageVolume(_ context.Context, req *csi.NodeUnstageVolume
...
+			}); err != nil {
+				klog.Errorf("NodeUnstageVolume: Error occurred while waiting for disk: %s to be detached from node: %s, error: %v", volumeID, d.NodeID, err)
+			}
+
+			if !detached {
+				return nil, status.Errorf(codes.DeadlineExceeded, "NodeUnstageVolume: Timed out waiting for disk: %s to be detached from node: %s", volumeID, d.NodeID)
+			}
+
+		} else {
+			return nil, status.Errorf(codes.Internal, "NodeUnStageVolume: detach for volume %s returned status %s (message: %q, error: %+v)", volumeID, statusResp.Status, statusResp.StatusMessage, statusResp.Error)
```
hccheng72, line 486: Q: Will it need to take some time to update the status after CleanupMountPoint()? So we probably see this error due to the delay.

pkg/azuredisk/nodeserver.go
```diff
@@ -51,6 +64,134 @@ const (
...
+const (
+	SeverityHintRetriable SeverityHint = "RETRIABLE"
+	SeverityHintFatal     SeverityHint = "FATAL"
+)
+
+type AttachmentStatus string
+
+const (
+	AttachmentStatusAttached  AttachmentStatus = "DISK_STATUS_ATTACHED"
+	AttachmentStatusDetached  AttachmentStatus = "DISK_STATUS_DETACHED"
```
hccheng72, line 186: Is this used? Based on the code in NodeUnstageVolume(), it looks like we won't find the disk from `detachResponse` if it's detached.

pkg/azuredisk/nodeserver.go
```diff
@@ -221,6 +431,61 @@ func (d *Driver) NodeUnstageVolume(_ context.Context, req *csi.NodeUnstageVolume
...
+			AttachSequence:  attachSequenceVal,
+		}, "DETACH")
+		if err != nil {
+			klog.Errorf("NodeUnStageVolume: failed to make POST call to wireserver for volume %s: %v", volumeID, err)
+			return nil, err
+		}
+
+		lowercaseVolumeID := strings.ToLower(volumeID)
+		if statusResp, ok := detachResponse[lowercaseVolumeID]; !ok {
+			klog.Infof("The volume with id %s is already detached", volumeID)
```
hccheng72, line 457: Based on the code in flushQADDiskBatch(), this line is never reached, right? 
```
diskStatus, ok := batchResponse[lowercaseDiskURI]
		if !ok {
			item.resultCh <- qadDiskBatchResult{err: status.Errorf(codes.Internal, "The response from wireserver doesn't contain volume %s", item.request.DiskURI)}
			continue
		}
```

pkg/azuredisk/controllerserver.go
```diff
@@ -516,6 +547,163 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
...
+	var result struct {
+		Properties *struct {
+			BlobURL         string `json:"blobUrl"`
+			ClaimIdentifier string `json:"claimIdentifier"`
+		} `json:"properties"`
+	}
+	if err := json.Unmarshal(body, &result); err != nil {
+		return "", "", fmt.Errorf("failed to parse claimResource response: %w", err)
+	}
+	return result.Properties.BlobURL, result.Properties.ClaimIdentifier, nil
```
hccheng72, line 649: need to check if result.Properties is nil before return

pkg/azuredisk/azuredisk.go
```diff
@@ -764,7 +817,39 @@ func GetNodeInfoFromNodeLister(nodeName string, nodeLister cache.GenericLister)
...
+	klog.Infof("Looking for PV with handle %s", diskURI)
+
+	// Use cached PV lister if available
+	if d.pvLister != nil {
+		pvs, err := d.pvLister.List(labels.Everything())
+		if err != nil {
+			return nil, fmt.Errorf("failed to list PersistentVolumes from cache: %v", err)
+		}
+		for _, pv := range pvs {
+			if pv.Spec.CSI != nil && pv.Spec.CSI.VolumeHandle == diskURI {
```
hccheng72, line 830: should we use `EqualFold` for diskURI and add the filter `pv.Spec.CSI.Driver == d.Name` to be consistent with the way we do in hasVolumeAttachmentForDiskOnNode()?

pkg/azuredisk/azuredisk.go
```diff
@@ -421,6 +452,13 @@ func NewDriver(options *DriverOptions) *Driver {
 		csi.NodeServiceCapability_RPC_SINGLE_NODE_MULTI_WRITER,
 	})
 
+	if kubeClient != nil {
```
hccheng72, line 455: We can gate it when the driver-level flag is ready.

pkg/azuredisk/controllerserver.go
nearora-msft: If it returns an error, we should just return an error saying that "Failed to determine whether QAD path is being used or not".

pkg/azuredisk/nodeserver.go
nearora-msft: We should return that error.

pkg/azuredisk/controllerserver.go
nearora-msft: Also, for preview we should use a different flag because if QAD is supported on a VM, only QAD can be used. 
This flag will come from AKS and should be passed to the CSI driver.
We should use an or condition here, something like, if (flag || d.isUsingQADPath) --> then use the QAD path.
This flag will always be true for preview.
Post preview, when both QAD and non-QAD attach is supported on a VM we will get rid of the flag and only use the isUsingQADPath.

pkg/azuredisk/controllerserver.go
```diff
@@ -538,14 +726,30 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
 	}
 	defer d.volumeLocks.Release(volumeID)
 
+	// If the PV is a QAD PV, unclaim the disk before deletion
+	pv, err := d.getPVFromDiskURI(ctx, diskURI)
+	if err != nil {
+		klog.Warningf("failed to get PV from disk URI %s: %v", diskURI, err)
+	} else if pv.Annotations != nil {
```
hccheng72, line 733: need to check pv != nil first

pkg/azuredisk/controllerserver.go
```diff
@@ -538,14 +726,30 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
...
+		if _, exists := pv.Annotations[azureconstants.AttachSequenceAnnotation]; exists {
+			klog.V(2).Infof("PV %s has QAD enabled, unclaiming disk %s before deletion", pv.Name, diskURI)
+			ownerResource, ownerErr := d.getAKSClusterResourceID(ctx)
+			if ownerErr != nil {
+				klog.Warningf("failed to determine QAD owner resource for disk %s (proceeding with deletion): %v", diskURI, ownerErr)
+			} else if unclaimErr := d.unclaimDiskResource(ctx, diskURI, ownerResource); unclaimErr != nil {
+				klog.Warningf("failed to unclaim QAD disk %s (proceeding with deletion): %v", diskURI, unclaimErr)
```
hccheng72, line 740: If failing to unclaim a QAD disk, is it able to be deleted? If not, should we return error here rather than warn-and-continue?

pkg/azuredisk/nodeserver.go
nearora-msft: True, it doesn't seem to be getting used. Feel free to remove if not needed.

pkg/azuredisk/nodeserver.go
nearora-msft: Thanks for catching that. Will update.

pkg/azuredisk/nodeserver.go
```diff
@@ -740,3 +1012,438 @@ func collectMountOptions(fsType string, mntFlags []string) []string {
...
+
+	// Update the annotation
+	if pv.Annotations == nil {
+		pv.Annotations = make(map[string]string)
+	}
+	currentAttachSequence := 0
+	if val, ok := pv.Annotations[azureconstants.AttachSequenceAnnotation]; ok {
+		currentAttachSequence, _ = strconv.Atoi(val)
+	}
+	updatedCounter := currentAttachSequence + 1
```
hccheng72, line 1052: need to update the variable name accordingly

pkg/azuredisk/nodeserver.go
```diff
@@ -740,3 +1012,438 @@ func collectMountOptions(fsType string, mntFlags []string) []string {
...
+	pv.Annotations[azureconstants.AttachSequenceAnnotation] = fmt.Sprintf("%d", updatedCounter)
```
hccheng72, line 1053: the pv is a pointer, need `pv := pv.DeepCopy()` before mutating?

pkg/azuredisk/nodeserver.go
```diff
@@ -740,3 +1012,438 @@ func collectMountOptions(fsType string, mntFlags []string) []string {
...
+
+	// Update the PV in Kubernetes
+	_, err := kubeClient.CoreV1().PersistentVolumes().Update(context.TODO(), pv, metav1.UpdateOptions{})
```
hccheng72, line 1056: plus retry.RetryOnConflict around the Update?

pkg/azuredisk/azuredisk.go
```diff
@@ -764,7 +817,39 @@ func GetNodeInfoFromNodeLister(nodeName string, nodeLister cache.GenericLister)
...
+			return nil, fmt.Errorf("failed to list PersistentVolumes from cache: %v", err)
+		}
+		for _, pv := range pvs {
+			if pv.Spec.CSI != nil && pv.Spec.CSI.VolumeHandle == diskURI {
+				klog.Infof("Found PV %s with handle %s (from cache)", pv.Name, diskURI)
+				return pv, nil
+			}
+		}
+		return nil, fmt.Errorf("cannot find PV with diskURI(%s)", diskURI)
+	}
```
landreasyan, line 836: which scenarios is the `d.pvLister` not available?

pkg/azuredisk/nodeserver.go
```diff
@@ -97,9 +238,78 @@ func (d *Driver) NodeStageVolume(_ context.Context, req *csi.NodeStageVolumeRequ
...
+
+		if err != nil {
+			klog.Errorf("NodeStageVolume: failed to make POST call to wireserver for volume %s: %v", volumeID, err)
+			return nil, err
+		}
+
+		lowercaseDiskURI := strings.ToLower(volumeID)
+		statusResp, ok := attachResponse[lowercaseDiskURI]
+		if !ok {
+			return nil, status.Errorf(codes.Internal, "The response from wireserver doesn't contain volume %s", volumeID)
```
landreasyan, line 274: do we need to decrement on failure?

pkg/azuredisk/nodeserver.go
landreasyan: this method is returning way too many things and the naming doesn't match it's purpose. We need a get PV method that will lookup the PV and we will pass the PV to an `isQAD` method which will lookup the annotation.

pkg/azuredisk/nodeserver.go
```diff
@@ -97,9 +238,78 @@ func (d *Driver) NodeStageVolume(_ context.Context, req *csi.NodeStageVolumeRequ
...
-	if !ok {
-		return nil, status.Error(codes.InvalidArgument, "lun not provided")
+	var lun string
+	// Check if this volume is using the QAD path.
+	// If yes, increment the attach-sequence and make an HTTP request to the QAD wireserver endpoint.
+	pv, isUsingQAD, err := d.isUsingQADPath(ctx, volumeID)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "NodeStageVolume: failed to determine if volume %s is using QAD path: %v", volumeID, err)
+	}
+	if isUsingQAD {
```
landreasyan, line 248: Can we move the QAD code into its own method?

pkg/azuredisk/nodeserver.go
landreasyan, line 1056: what would happen if we increment the counter but fail to issue the attach call? Would the next attach with its own increment fail?

pkg/azuredisk/nodeserver.go
```diff
@@ -740,3 +1012,438 @@ func collectMountOptions(fsType string, mntFlags []string) []string {
...
+	_, err := kubeClient.CoreV1().PersistentVolumes().Update(context.TODO(), pv, metav1.UpdateOptions{})
+	if err != nil {
+		return 0, fmt.Errorf("failed to update PersistentVolume %s: %v", pv.Name, err)
+	}
+
+	klog.V(2).Infof("Successfully incremented attach sequence annotation for PV %s to %s", pv.Name, pv.Annotations[azureconstants.AttachSequenceAnnotation])
+	return updatedCounter, nil
+}
+
+func (d *Driver) enqueueQADDiskOperation(ctx context.Context, diskRequest DiskOperationRequest, operationType string) (WireserverDiskStatusResponse, error) {
```
landreasyan, line 1065: Can we attach and detach disks in the same batch?

pkg/azuredisk/nodeserver.go
nearora-msft: The PR changes contain the initial changes mainly to test out that e2e flow works. The changes definitely need refactoring and please feel free to do that.

pkg/azuredisk/controllerserver.go
nearora-msft: We should return an error here if unclaim fails because a claimed disk cannot be deleted, it has to be unclaimed first.

pkg/azuredisk/nodeserver.go
nearora-msft: No, this counter is monotonously increasing.

pkg/azuredisk/nodeserver.go
nearora-msft: No, the attach with the greatest counter wins.

pkg/azuredisk/nodeserver.go
nearora-msft: Yes!
We could improve the batching strategy to have a single queue for attach and detach operations.

PR #3804, opened 2026-09-03
[QAD] fix issues found from tests

hccheng72:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


pkg/azuredisk/controllerserver.go
```diff
@@ -1065,12 +1070,10 @@ func (d *Driver) ControllerUnpublishVolume(ctx context.Context, req *csi.Control
 	pv, err := d.getPVFromDiskURI(ctx, diskURI)
 	if err != nil {
```
landreasyan, line 1071: should `pv != nil` be here?

pkg/azuredisk/controllerserver.go
landreasyan, line 750: should move this prior to PV lookup.

pkg/azuredisk/controllerserver.go
```diff
@@ -729,15 +737,15 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
 	// If the PV is a QAD PV, unclaim the disk before deletion
 	pv, err := d.getPVFromDiskURI(ctx, diskURI)
 	if err != nil {
-		klog.Warningf("failed to get PV from disk URI %s: %v", diskURI, err)
-	} else if pv.Annotations != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get PV from disk URI %s: %v", diskURI, err)
+	} else if pv != nil && pv.Annotations != nil {
 		if _, exists := pv.Annotations[azureconstants.AttachSequenceAnnotation]; exists {
 			klog.V(2).Infof("PV %s has QAD enabled, unclaiming disk %s before deletion", pv.Name, diskURI)
 			ownerResource, ownerErr := d.getAKSClusterResourceID(ctx)
```
landreasyan, line 741: `ownerErr` -> `err`?

pkg/azuredisk/controllerserver.go
```diff
@@ -1927,162 +1934,162 @@ func inlineVolumeSpecMatchesDisk(driverName, diskURI string, va *storagev1.Volum
...
-	}
-
-	cleaned := url.Values{}
-	cleaned.Set("comp", "disksession")
-	cleaned.Set("timeout", "30")
-
-	parsed.RawQuery = cleaned.Encode()
-	return parsed.String(), nil
-}
+// // getDiskAccessSAS calls the Azure REST API beginGetAccess to obtain an access SAS URL for the given disk.
```
landreasyan: if not needed anymore, can we delete this?

pkg/azuredisk/nodeserver.go
```diff
@@ -1394,7 +1417,7 @@ func mapHTTPStatusToCode(httpStatus int) codes.Code {
 	}
 }
 
-func getAttachedDisks(ctx context.Context, client http.Client) (WireserverDiskStatusResponse, error) {
+func getDiskStatuses(ctx context.Context, client http.Client) (WireserverDiskStatusResponse, error) {
```
landreasyan: maybe we can use `getDiskStates`?

pkg/azuredisk/nodeserver.go
```diff
@@ -1013,53 +1021,63 @@ func collectMountOptions(fsType string, mntFlags []string) []string {
 	return options
 }
 
-func (d *Driver) isUsingQADPath(ctx context.Context, diskURI string) (*v1.PersistentVolume, bool, error) {
-
-	klog.Infof("Checking if diskURI %s is using QAD path", diskURI)
-	pv, err := d.getPVFromDiskURI(ctx, diskURI)
-	if err != nil {
-		return nil, false, err
+func (d *Driver) isQAD(pv *v1.PersistentVolume) bool {
```
landreasyan, line 1030: Great change!

pkg/azuredisk/controllerserver.go
hccheng72: I feel it’s easier to read the code when we use distinct error names, since it makes it clear which operation each error came from, e.g., ownerErr and unclaimErr. WDYT?

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
```diff
@@ -81,7 +81,7 @@ spec:
 {{- end }}
             - "--csi-address=$(ADDRESS)"
             - "--v=2"
-            - "--timeout=30s"
+            - "--timeout=120s"
```
landreasyan: this change needs to be guarded with a flag since it is QAD specific. You can leave this out from this PR and I'll add it when we introduce the flag.

pkg/azuredisk/controllerserver.go
```diff
@@ -1063,14 +1065,12 @@ func (d *Driver) ControllerUnpublishVolume(ctx context.Context, req *csi.Control
...
 		klog.Errorf("failed to get PV from disk URI %s: %v", diskURI, err)
-	} else {
-		if pv.Annotations != nil {
-			if _, exists := pv.Annotations[azureconstants.AttachSequenceAnnotation]; exists {
-				klog.V(2).Infof("PV %s has QAD enabled, skipping detach for disk %s", pv.Name, diskURI)
-				return &csi.ControllerUnpublishVolumeResponse{}, nil
-			}
+	} else if pv.Annotations != nil {
+		if _, exists := pv.Annotations[azureconstants.AttachSequenceAnnotation]; exists {
+			klog.V(2).Infof("PV %s has QAD enabled, skipping detach for disk %s", pv.Name, diskURI)
```
landreasyan: if we got to this point, disk is already detached from this node, right?

pkg/azuredisk/nodeserver.go
```diff
@@ -1013,53 +1027,63 @@ func collectMountOptions(fsType string, mntFlags []string) []string {
...
+		currentAttachSequence := 0
+		if val, ok := pv.Annotations[azureconstants.AttachSequenceAnnotation]; ok {
+			currentAttachSequence, _ = strconv.Atoi(val)
+		}
+		updatedSequence = currentAttachSequence + 1
+		pv.Annotations[azureconstants.AttachSequenceAnnotation] = strconv.Itoa(updatedSequence)
 
-	// Update the PV in Kubernetes
-	_, err := kubeClient.CoreV1().PersistentVolumes().Update(context.TODO(), pv, metav1.UpdateOptions{})
+		_, err := kubeClient.CoreV1().PersistentVolumes().Update(context.TODO(), pv, metav1.UpdateOptions{})
```
landreasyan: We should get the context from the caller and pass it in to the update call.

pkg/azuredisk/nodeserver.go
```diff
@@ -453,20 +466,20 @@ func (d *Driver) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstageVolu
...
-		if statusResp, ok := detachResponse[lowercaseVolumeID]; !ok {
+		if statusResp, ok := detachResponse[lowercaseVolumeID]; !ok || statusResp.Status == AttachmentStatusDetached {
 			klog.Infof("The volume with id %s is already detached", volumeID)
 			klog.Infof("NodeUnStageVolume: Latency observed for detach operation of disk %s is %v", volumeID, time.Since(detachTimer).Milliseconds())
 		} else if statusResp.Status == AttachmentStatusDetaching {
 			detached := false
-			if err = kwait.PollUntilContextTimeout(ctx, 500*time.Millisecond, 5*time.Second, true, func(context.Context) (bool, error) {
-				getDisksResponse, err := getAttachedDisks(ctx, *d.httpClient)
+			if err = kwait.PollUntilContextTimeout(ctx, 500*time.Millisecond, 15*time.Second, true, func(context.Context) (bool, error) {
+				getDisksResponse, err := getDiskStates(ctx, *d.httpClient)
```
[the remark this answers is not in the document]
landreasyan: good recommendation.

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
hccheng72: yes, we can reconsider the appropriate timeout once claim requests would automatically be distributed across storage stamps (DiskRP to confirm).

pkg/azuredisk/controllerserver.go
hccheng72: not necessarily, the NodeUnstageVolume could fail. I updated the log a bit.

pkg/azuredisk/nodeserver.go
```diff
@@ -431,10 +439,15 @@ func (d *Driver) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstageVolu
 	}
 	klog.V(2).Infof("NodeUnstageVolume: unmount %s successfully", stagingTargetPath)
 
-	if pv, isUsingQAD, err := d.isUsingQADPath(ctx, volumeID); isUsingQAD && err == nil {
+	pv, err := d.getPVFromDiskURI(ctx, volumeID)
+	if err != nil || pv == nil {
+		return nil, status.Errorf(codes.Internal, "NodeUnstageVolume: failed to get PV from diskURI %s: %v", volumeID, err)
```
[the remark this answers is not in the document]
landreasyan: return success when err is NotFound

pkg/azuredisk/controllerserver.go
```diff
@@ -728,16 +733,16 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
 
 	// If the PV is a QAD PV, unclaim the disk before deletion
 	pv, err := d.getPVFromDiskURI(ctx, diskURI)
-	if err != nil {
-		klog.Warningf("failed to get PV from disk URI %s: %v", diskURI, err)
+	if err != nil || pv == nil {
+		return nil, status.Errorf(codes.Internal, "failed to get PV from disk URI %s: %v", diskURI, err)
```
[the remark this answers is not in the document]
landreasyan: if err is NotFound skip the unclaim/qad related path but still proceed to calling disk delete. If we can distinguish from the delete error that unclaim is needed, call unclaim and retry delete.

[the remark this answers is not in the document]
hccheng72: line 930 is in ControllerPublishVolume, that PV lookup occurs only after the request explicitly indicates QAD is enabled. Publishing QAD requires updating the PV with QAD annotations, so a missing PV must fail.

pkg/azuredisk/controllerserver.go
```diff
@@ -1435,7 +1436,11 @@ func (d *Driver) ControllerExpandVolume(ctx context.Context, req *csi.Controller
 
 	diskURI := req.GetVolumeId()
 
-	if _, isQAD, err := d.isUsingQADPath(ctx, diskURI); err == nil && isQAD {
+	pv, err := d.getPVFromDiskURI(ctx, diskURI)
+	if err != nil || pv == nil {
+		return nil, status.Errorf(codes.Internal, "failed to get PV from diskURI %s: %v", diskURI, err)
```
landreasyan: Is there a reason this is not working for QAD? 

Since we are returning unimplemented, I would recommend not returning an error when PV lookup fails, as this alters existing behavior. Only do the QAD check when PV lookup is successful.

pkg/azuredisk/controllerserver.go
```diff
@@ -728,16 +734,17 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
 
 	// If the PV is a QAD PV, unclaim the disk before deletion
 	pv, err := d.getPVFromDiskURI(ctx, diskURI)
-	if err != nil {
-		klog.Warningf("failed to get PV from disk URI %s: %v", diskURI, err)
-	} else if pv.Annotations != nil {
+	if err != nil && !errors.Is(err, errPVNotFound) {
+		return nil, status.Errorf(codes.Internal, "failed to get PV from disk URI %s: %v", diskURI, err)
+	}
+	if pv != nil && pv.Annotations != nil {
```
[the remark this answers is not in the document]
landreasyan: there is a fallback to unclaim if delete fails.

pkg/azuredisk/controllerserver.go
```diff
@@ -774,7 +796,12 @@ func (d *Driver) ControllerModifyVolume(ctx context.Context, req *csi.Controller
 
 	diskURI := volumeID
 
-	if _, isQAD, err := d.isUsingQADPath(ctx, diskURI); err == nil && isQAD {
+	pv, err := d.getPVFromDiskURI(ctx, diskURI)
+	if err != nil || pv == nil {
+		return nil, status.Errorf(codes.Internal, "failed to get PV from diskURI %s: %v", diskURI, err)
```
landreasyan: To keep the old behavior - only log a warning when PV is not found. When pv != nil &&  d.isQAD(pv) return the Unimplemented error.

landreasyan: There is no reason to add a pv lookup dependency in this call if we only return an Unimplemented after the lookup.

pkg/azuredisk/controllerserver.go
```diff
@@ -917,47 +944,16 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
...
-
-		// blobURL := volumeContext[azureconstants.BlobURLAnnotation]
-		if blobURL == "" {
-			return nil, status.Errorf(codes.Internal, "QAD blob URL missing for disk %s", diskName)
-		}
-		// claimIdentifier := volumeContext[azureconstants.ClaimIdentifierAnnotation]
-		if claimIdentifier == "" {
-			return nil, status.Errorf(codes.Internal, "QAD claim identifier missing for disk %s", diskName)
-		}
+		blobURL := volumeContext[azureconstants.BlobURLAnnotation]
```
landreasyan, line 947: how is this populated for static volumes?

pkg/azuredisk/controllerserver.go
```diff
@@ -1053,24 +1049,44 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
 	return &csi.ControllerPublishVolumeResponse{PublishContext: publishContext}, nil
 }
 
+func ensureQADPVAnnotations(ctx context.Context, kubeClient clientset.Interface, pvName, blobURL, claimIdentifier string) error {
```
landreasyan, line 1052: check for non-empty parameters = pvName, blobUR, claimIdentifier

pkg/azuredisk/controllerserver.go
landreasyan: Turns out resize affects the blobUri and isn't supported for claimed disks.

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
landreasyan: Can we remove this change from this PR?

pkg/azuredisk/nodeserver.go
```diff
@@ -275,14 +279,18 @@ func (d *Driver) NodeStageVolume(ctx context.Context, req *csi.NodeStageVolumeRe
 		}
 		if statusResp.Status == AttachmentStatusAttaching {
 			// Wait for the disk to be attached
-			if err = kwait.PollUntilContextTimeout(ctx, 250*time.Millisecond, 5*time.Second, true, func(context.Context) (bool, error) {
-				getDisksResponse, err := getAttachedDisks(ctx, *d.httpClient)
+			if err = kwait.PollUntilContextTimeout(ctx, 250*time.Millisecond, 15*time.Second, true, func(context.Context) (bool, error) {
+				getDisksResponse, err := getDiskStates(ctx, *d.httpClient)
```
landreasyan: this call should use the context from the func(context.Context) instead of the main ctx.

pkg/azuredisk/controllerserver.go
```diff
@@ -726,31 +732,47 @@ func (d *Driver) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest)
...
+	if pv != nil && pv.Annotations != nil {
 		if _, exists := pv.Annotations[azureconstants.AttachSequenceAnnotation]; exists {
 			klog.V(2).Infof("PV %s has QAD enabled, unclaiming disk %s before deletion", pv.Name, diskURI)
 			ownerResource, ownerErr := d.getAKSClusterResourceID(ctx)
 			if ownerErr != nil {
-				klog.Warningf("failed to determine QAD owner resource for disk %s (proceeding with deletion): %v", diskURI, ownerErr)
+				return nil, status.Errorf(codes.Internal, "failed to determine QAD owner resource for disk %s: %v", diskURI, ownerErr)
 			} else if unclaimErr := d.unclaimDiskResource(ctx, diskURI, ownerResource); unclaimErr != nil {
-				klog.Warningf("failed to unclaim QAD disk %s (proceeding with deletion): %v", diskURI, unclaimErr)
+				return nil, status.Errorf(codes.Internal, "failed to unclaim QAD disk %s: %v", diskURI, unclaimErr)
```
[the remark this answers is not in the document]
hccheng72: unclaim should be idempotent

pkg/azuredisk/controllerserver.go
hccheng72: will check static volumes have those annotations in ensureQADPVAnnotations()

charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml
hccheng72: removed

pkg/azuredisk/azure_common_windows.go
```diff
@@ -87,6 +87,13 @@ func preparePublishPath(path string, m *mount.SafeFormatAndMount) error {
 
 func CleanupMountPoint(path string, m *mount.SafeFormatAndMount, unmountVolume bool) error {
 	if proxy, ok := m.Interface.(mounter.CSIProxyMounter); ok {
+		exists, err := proxy.ExistsPath(path)
```
landreasyan: what is this change for?

hccheng72: It makes Windows NodeUnstageVolume idempotent when the staging path is already absent. This also lets the QAD detached-response tests reach the QAD logic on Windows.

hccheng72: reverted it and skipped the unit test on Windows since QAD is not supported on Windows.

PR #3809, opened 2026-09-09
[QAD] Feature gate qad

landreasyan:
**What type of PR is this?**

Introduces `NodeDrivenAttachDetach` as an `Alpha`, disabled-by-default driver feature gate for the node-driven/QAD attachment architecture.

The feature separates:

- Cluster capability: NodeDrivenAttachDetach feature gate
- Per-volume selection: `attachMode: NodeDriven`
- Durable runtime state: PV attachment-mode annotations

**Behavior**
- New volumes default to ControllerDriven.
- `attachMode: NodeDriven` requires NodeDrivenAttachDetach=true.
- Dynamic volumes are validated during CreateVolume.
- Static volumes are validated during initial adoption in ControllerPublishVolume.
- Once adopted, the volume’s attachment mode is persisted on the PV.
- Existing node-driven volumes continue using the node-driven path even if the feature gate is later disabled.
- Node attach/detach behavior is selected from persisted volume state rather than the current gate value.

**Configuration**
Enable the feature through the standard driver feature-gate flag:

```--feature-gates=NodeDrivenAttachDetach=true```

Then opt individual StorageClasses into the architecture:
```
parameters:
  attachMode: NodeDriven
```

The previous qadEnabled StorageClass parameter is not exposed as part of this Alpha API.
**Release note**:
```
none
```

### CODE REVIEW

File: deploy/example/storageclass-azuredisk-csi-node-driven.yaml [hunk 1]
```diff
@@ -0,0 +1,16 @@
...
+# pools must support QAD. Do not make this the default StorageClass during preview.
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: managed-csi-node-driven
+provisioner: disk.csi.azure.com
+parameters:
+  skuName: Premium_LRS
+  attachMode: NodeDriven
+  networkAccessPolicy: AllowAll
```
andyzhangx, line 12: are these two settings required in `NodeDriven` ?

```
  networkAccessPolicy: AllowAll
  publicNetworkAccess: Enabled
```

deploy/example/storageclass-azuredisk-csi-node-driven.yaml
```diff
@@ -0,0 +1,16 @@
...
+# --feature-gates=NodeDrivenAttachDetach=true, and the Azure platform and node
+# pools must support QAD. Do not make this the default StorageClass during preview.
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: managed-csi-node-driven
+provisioner: disk.csi.azure.com
+parameters:
+  skuName: Premium_LRS
+  attachMode: NodeDriven
```
andyzhangx, line 11: `ControllerDriven` and `NodeDriven` would likely be confusing for end users. They describe internal implementation details, not user-facing value. From a user perspective, the important question is why they would enable this feature at all; whether the attach/detach flow is controller-driven or node-driven is not especially meaningful. Why not keep the UX simple and expose it as something like `enableQAD: "true"` instead?

charts/latest/azuredisk-csi-driver/values.yaml
```diff
@@ -205,6 +205,9 @@ feature:
 
 driver:
   name: disk.csi.azure.com
+  # Azure Disk CSI driver feature gates. Alpha features are disabled by default.
+  featureGates:
+    NodeDrivenAttachDetach: false
```
andyzhangx, line 210: there should be CRDs and RBAC rules which is related to this feature gate, if it's not enabled, those CRDs and RBAC rules should not be installed.

deploy/example/storageclass-azuredisk-csi-node-driven.yaml
landreasyan: I agree the names are implementation-oriented, but QAD is more than an optional optimization in the OSS driver, it selects a different attachment architecture with different infrastructure and identity prerequisites. Unlike managed AKS, where the platform configures those prerequisites, OSS operators must explicitly prepare the cluster for QAD. A Boolean like enableQAD makes it look like a simple feature toggle and does not clearly represent the mutually exclusive operating modes or leave room for Beta, when both modes may be supported.

pkg/azuredisk/controllerserver.go
```diff
@@ -908,14 +914,22 @@ func (d *Driver) ControllerPublishVolume(ctx context.Context, req *csi.Controlle
...
-	// FAD will not be supported.
-	qadEnabledValue, _ := azureutils.ParseDiskParametersForKey(volumeContext, azureconstants.QADEnabledField)
-	if strings.EqualFold(qadEnabledValue, consts.TrueValue) {
+	// The feature gate permits adoption of the Alpha architecture, while the
+	// persisted volume attachment mode selects QAD for this disk.
+	attachMode, err := getAttachMode(volumeContext)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "failed to determine attachment mode: %v", err)
+	}
+	if attachMode == consts.AttachModeNodeDriven {
```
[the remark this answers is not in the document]
landreasyan: backwards compatibility is not relevant here

PR #3810, opened 2026-09-10
fix: clear snapshot continuation token after the last page

Hanabi9248:
**What type of PR is this?**

/kind bug

ListSnapshots returns a numeric next token even after exhausting the snapshot list. A client following that token gets Aborted instead of completing pagination. An empty list similarly returns "0", which the next request rejects.

Leave the token empty when the scan reaches the end. Keep the existing offset for intermediate pages. Add regression coverage that follows the returned tokens for empty lists, multiple pages and source-volume filters, and update the controller test's final-page expectation.

No linked issue.

**Requirements**:
- [x] uses conventional commit messages
- [ ] includes documentation
- [x] adds unit tests

Tested on Windows:
- `go test ./pkg/azureutils -skip '^TestOSIOHandler_ReadLink$' -count=1`
- `go test ./pkg/azuredisk -run '^TestListSnapshots$' -count=1`
- `go vet ./pkg/azureutils`

The four pagination regressions fail before the fix and pass afterward. The excluded ReadLink test requires symlink privileges unavailable here. Broader `go vet ./pkg/azuredisk` reports existing copylocks warnings in azure_common_windows.go; both limitations also reproduce on the unmodified base. Azure cluster tests and the full make verify target were not run.

**Release note**:
```
Fixed ListSnapshots returning an unusable next token after the last page.
```


andyzhangx:
@Hanabi9248 thanks for the contribution, can you sign the cla?

andyzhangx:
> @Hanabi9248 thanks for the contribution, can you sign the cla?

@Hanabi9248 can you sign the cla? thx

Hanabi9248:
Thanks for following up. EasyCLA now passes for both commits, and the `cncf-cla: yes` label is present. Could you enable the Prow tests when convenient? The PR is still marked `needs-ok-to-test`.

PR #3811, opened 2026-09-10
chore: wait for filesystem shutdown after NodeUnstageVolume unmount

mittachaitu:
**What type of PR is this?**

/kind bug

This PR adds a Linux-only filesystem shutdown check after unmounting a volume during NodeUnstageVolume. The driver waits for the filesystem entry in sysfs to disappear and, for filesystems, for the corresponding JBD2 journal entry in jbd2 to be removed. The wait timeout is configurable through `--filesystem-shutdown-timeout` and defaults to 5 seconds.

If an entry still exists after the timeout, the driver emits warning logs and filesystem-type-labeled metrics. These checks are informational and do not cause NodeUnstageVolume to fail.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azuredisk/azure_common_linux.go
```diff
@@ -566,3 +575,121 @@ func (d *Driver) GetVolumeStats(_ context.Context, m *mount.SafeFormatAndMount,
...
+	klog.Warningf("Filesystem shutdown checks failed for device %q: %s", devicePath, strings.Join(errorMessages, "; "))
+	return errorMessages
+}
+
+func journalEntryGlob(fsType string, deviceName string) string {
+	return journalEntryGlobAt(procFSJbd2Root, fsType, deviceName)
+}
+
+func journalEntryGlobAt(jbd2Root string, fsType string, deviceName string) string {
+	if fsType != "ext4" {
```
NedAnd1: suggest also checking that fsType is not `ext3`

mittachaitu: Updated the code to have ext3 check

charts/latest/azuredisk-csi-driver/values.yaml
```diff
@@ -153,6 +153,7 @@ node:
   maxUnavailable: 1
   logLevel: 5
   metricsPort: 29605
+  filesystemShutdownTimeout: 5
```
andyzhangx, line 156: `filesystemShutdownTimeout` should be under `linux` section since it's only for linux, and if it's not a common parameter user needs to adjust, you may not set param in helm chart, just leave it as a driver parameter

andyzhangx (commented):
from this PR, it actually `waitFSShutdown` instead of just add inform logs as in PR title, and it's `/kind bug` PR? would you add more info in the PR description and change the PR title?

charts/latest/azuredisk-csi-driver/values.yaml
mittachaitu: moved under `linux` section

mittachaitu:
> from this PR, it actually `waitFSShutdown` instead of just add inform logs as in PR title, and it's `/kind bug` PR? would you add more info in the PR description and change the PR title?

Updated the PR title and description, PTAL

PR #3830, opened 2026-09-21
fix: skip fsck on newly formatted volumes

NedAnd1:
**What type of PR is this?**

/kind bug

Undoes a behavior change in this driver that started calling fsck on newly formatted volumes.
Increases behavior parity with prior versions & other CSI drivers, and reduces the vector for errors in the mount path.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/azuredisk/azure_common_linux.go
```diff
@@ -201,21 +201,21 @@ func formatAndMount(source, target, fstype string, options []string, m *mount.Sa
...
 			diskFSFormat = reReadFormat
-		}
-	}
 
-	if !strings.EqualFold(fstype, diskFSFormat) {
-		// Verify that the disk is formatted with filesystem type we are expecting
-		klog.Errorf("formatAndMount - configured to mount disk %s as %s but current format is %s, things might break", source, fstype, diskFSFormat)
-		return fmt.Errorf("configured to mount disk %s as %s but current format is %s, things might break", source, fstype, diskFSFormat)
-	}
+			if !strings.EqualFold(fstype, diskFSFormat) {
```
ganesan23: Should this check and the detectAndRepairFilesystem check be outside of the 

if diskFSFormat == "" {}

?. 

We could go with something like

newlyFormatted := false
if diskFSFormat == "" {
    if !isFilesystemExist {
        // mkfs
        diskFSFormat = fstype
        newlyFormatted = true
    } else {
        diskFSFormat = reReadFormat
    }
}

// Always validate the actual/requested filesystem types.
if !strings.EqualFold(fstype, diskFSFormat) {
    return fmt.Errorf(...)
}

// Repair existing read-write filesystems, but not a filesystem just created.
if !readOnly && !newlyFormatted {
    detectAndRepairFilesystem(...)
}

ganesan23 (changes requested):

pkg/azuredisk/azure_common_linux.go
NedAnd1: Good callout! Resolved.

pkg/azuredisk/azure_common_linux.go
```diff
@@ -210,8 +212,8 @@ func formatAndMount(source, target, fstype string, options []string, m *mount.Sa
 		return fmt.Errorf("configured to mount disk %s as %s but current format is %s, things might break", source, fstype, diskFSFormat)
 	}
 
-	// Running fsck on the disk to detect and repair any filesystem issues before mounting
-	if !readOnly {
+	// Run fsck on the disk to fix repairable issues, only do this for already formatted volumes requested as read-write.
+	if !newlyFormatted && !readOnly {
```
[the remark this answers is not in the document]
NedAnd1: Resolved.

ganesan23 (approved):

Voyagerroc-Lab (commented):
Reviewed the newlyFormatted state transition in formatAndMount, the preservation of the existing formatted-volume fsck path, and the command-count test updates. Skipping repair fsck only for a volume formatted in this invocation avoids touching a filesystem before its first mount while retaining repair behavior for previously formatted read-write volumes and read-only safety. I found no blocking issue.

mittachaitu (approved):
LGTM!!

PR #3831, opened 2026-09-21
[QAD] fix several issues on the feature branch

hccheng72:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


deploy/example/storageclass-azuredisk-csi-node-driven.yaml
```diff
@@ -11,4 +11,4 @@ parameters:
   attachMode: NodeDriven
 reclaimPolicy: Delete
 volumeBindingMode: WaitForFirstConsumer
-allowVolumeExpansion: true
+allowVolumeExpansion: false
```
landreasyan: do we need this change?

pkg/azuredisk/azuredisk.go
```diff
@@ -848,27 +866,41 @@ func pvDiskURIIndexFunc(obj interface{}) ([]string, error) {
...
 		}
 		for _, obj := range objs {
-			pv, ok := obj.(*v1.PersistentVolume)
-			if !ok {
+			candidate, ok := obj.(*v1.PersistentVolume)
+			if !ok || candidate.Spec.CSI == nil || candidate.Spec.CSI.Driver != d.Name ||
+				!strings.EqualFold(candidate.Spec.CSI.VolumeHandle, diskURI) {
 				continue
 			}
+			if d.kubeClient == nil {
```
landreasyan: why is this fallback needed?

pkg/azuredisk/azuredisk.go
```diff
@@ -848,27 +866,41 @@ func pvDiskURIIndexFunc(obj interface{}) ([]string, error) {
...
+				return nil, fmt.Errorf("%w: Kubernetes client is not initialized", errPVMetadataUnavailable)
+			}
+			pv, err := d.kubeClient.CoreV1().PersistentVolumes().Get(ctx, candidate.Name, metav1.GetOptions{})
```
landreasyan: No need to lookup the candidate if it was already in the cache. There is already a fallback to lookup pv-s on line 904

landreasyan: actually, is this lookup to avoid stale cache?

landreasyan: if so, lets only make this call for when we need a current attach sequence number, i.e. in node stage and unstage, otherwise, the pv from cache should work just fine.

deploy/example/storageclass-azuredisk-csi-node-driven.yaml
landreasyan: I see why :D

pkg/azuredisk/azuredisk.go
```diff
@@ -483,6 +475,32 @@ func NewDriver(options *DriverOptions) *Driver {
 	return &driver
 }
 
+func (d *Driver) initializePVInformer(kubeClient clientset.Interface) {
+	if kubeClient == nil || !d.nodeDrivenAttachDetachEnabled {
+		return
```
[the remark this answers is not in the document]
hccheng72: all getPVFromDiskURI is call under the feature gate.

pkg/azuredisk/azuredisk.go
```diff
@@ -461,19 +463,7 @@ func NewDriver(options *DriverOptions) *Driver {
...
-		pvInformer := driver.informerFactory.Core().V1().PersistentVolumes()
-		if err := pvInformer.Informer().AddIndexers(cache.Indexers{pvDiskURIIndex: pvDiskURIIndexFunc}); err != nil {
-			klog.Warningf("failed to add diskURI indexer to PV informer, will fall back to API list: %v", err)
-		} else {
-			driver.pvIndexer = pvInformer.Informer().GetIndexer()
-		}
-		driver.pvLister = pvInformer.Lister()
-		driver.pvListerSynced = pvInformer.Informer().HasSynced
-	}
+	driver.initializePVInformer(kubeClient)
```
landreasyan, line 466: can we check for ` !d.nodeDrivenAttachDetachEnabled ` before call the method, so the intention is more clear?

pkg/azuredisk/azuredisk.go
```diff
@@ -534,8 +550,8 @@ func (d *Driver) Run(ctx context.Context) error {
 	// Start informer factory if initialized
 	if d.informerFactory != nil {
 		d.informerFactory.Start(ctx.Done())
-		if !cache.WaitForCacheSync(ctx.Done(), d.pvListerSynced) {
-			klog.Errorf("failed to sync PV informer cache")
+		if !d.waitForPVInformerCacheSync(ctx, informerCacheSyncTimeout) {
```
landreasyan, line 553: should this be called only when `d.nodeDrivenAttachDetachEnabled = true`?

pkg/azuredisk/azuredisk.go
hccheng72: we check it in the first line of this function. Is it okay?
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/d88ec59703aed793357383be0c7057058e12aa3d/pkg/azuredisk/azuredisk.go#L479

pkg/azuredisk/azuredisk.go
hccheng72: It already is, indirectly. initializePVInformer returns unless d.nodeDrivenAttachDetachEnabled is true, so d.informerFactory remains nil otherwise.

pkg/azuredisk/controllerserver.go
```diff
@@ -450,6 +456,19 @@ func (d *Driver) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest)
 	}, nil
 }
 
+func (d *Driver) unclaimAfterFailedQADSetup(ctx context.Context, diskURI, ownerResourceID string, setupErr error) error {
+	cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), qadCreateCleanupTimeout)
+	defer cancel()
+
+	if err := d.unclaimDiskResource(cleanupCtx, diskURI, ownerResourceID); err != nil {
```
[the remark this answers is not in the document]
hccheng72: copilot:
for an ambiguous claim result, leaving it claimed is safer than unclaiming it.
It only ensures the normal Aborted retry recovers the claim. If provisioning is abandoned before retry, the claim may remain orphaned.

These guarantees cannot both be achieved with the current API response:
- Unclaim after ambiguity: may destroy a valid preexisting claim.
- Preserve after ambiguity: may leave a new claim orphaned.

A complete fix requires DiskRP to expose whether this request newly established the claim, or support an idempotency/operation token. Until then, preserving the claim avoids the more destructive data-path regression. Definite failures before the request is sent do not need cleanup; definite newly-established claims could be cleaned up only if we gain reliable provenance.

